PR #3, opened 2023-08-28
Relay supported fee assets

cam-schultz:
Adds Teleporter configuration option for supported fee assets. The relayer will only deliver a message if it is incentivized with a supported fee asset, or if the configured list of fee assets is empty.

In addition to that check, the main change here is adding an additional eth client in the Teleporter message manager, so that we can call getFeeInfo on the Teleporter contract. The message manager constructor takes the source subnet configuration to do so.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -42,6 +42,7 @@ type SourceSubnet struct {
 	APINodeHost       string                           `mapstructure:"api-node-host" json:"api-node-host"`
 	APINodePort       uint32                           `mapstructure:"api-node-port" json:"api-node-port"`
 	EncryptConnection bool                             `mapstructure:"encrypt-connection" json:"encrypt-connection"`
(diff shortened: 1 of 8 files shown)


minghinmatthewlam:
@cam-schultz do we want to close this draft as wellf or now, and revisit as it's picked back up?

cam-schultz:
> @cam-schultz do we want to close this draft as wellf or now, and revisit as it's picked back up?

This is still a candidate change, so let's leave it open.

minghinmatthewlam (commented):
High level LGTM, should be similar to https://github.com/ava-labs/awm-relayer/pull/59 in terms of allowing specific fields, otherwise allow all. Will re-review after it's updated with latest changes, some of the abi packing/unpacking might look different with Teleporter pkg.

Issue #14, opened 2023-08-28
Explore catch up options for relayer downtime


minghinmatthewlam:
**Context and scope**
Currently, the relayer requires a live connection to receive message requests to relay. If the relayer is offline, it is unable to relay historical messages that it may have missed.

**Discussion and alternatives**
Some ideas and considerations:

- Explore options in the ethClient interface for fetching historical logs from a certain block height
- Perform a bootstrap from a particular block height or fetch specific message IDs on startup
  - However, we need to think about how to do this in a way that doesn't repeat these startup actions after they have been successfully performed. This might require adding some state to the relayer.

Issue #15, opened 2023-08-28
Set up end-to-end testing framework


minghinmatthewlam:
**Context and scope**
Relayer repo right now has some unit tests, but no e2e tests. We should add testing for e2e coverage, from relayer start up and reading its config, and making sure it can process mock cross chain message requests.

PR #20, opened 2023-08-30
Process missed blocks on startup

cam-schultz:
## Why this should be merged
Enables the relayer to process warp messages sent while the relayer was offline. Fixes https://github.com/ava-labs/awm-relayer/issues/14

## How this works
Adds a `ProcessFromHeight` method to `Subscriber`. The EVM implementation calls `eth_getLogs` to retrieve Warp logs since the last processed block. Adds JSON storage to persist the last processed block.

## How this was tested
CI, Teleporter integration tests, added unit tests for JSON storage, manual testing. Will circle back and add end-to-end tests once https://github.com/ava-labs/awm-relayer/issues/15 is complete.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -219,6 +219,9 @@ func (s *SourceSubnet) Validate() error {
 	if _, err := url.ParseRequestURI(s.GetNodeWSEndpoint()); err != nil {
 		return fmt.Errorf("invalid relayer subscribe URL in source subnet configuration: %v", err)
 	}
+	if _, err := url.ParseRequestURI(s.GetNodeRPCEndpoint()); err != nil {
+		return fmt.Errorf("invalid relayer RPC URL in source subnet configuration: %v", err)
+	}
 
 	// Validate the VM specific settings
 	switch ParseVM(s.VM) {
--- a/config/config.go
+++ b/config/config.go
@@ -42,6 +42,7 @@ type SourceSubnet struct {
 	APINodeHost       string                           `mapstructure:"api-node-host" json:"api-node-host"`
 	APINodePort       uint32                           `mapstructure:"api-node-port" json:"api-node-port"`
 	EncryptConnection bool                             `mapstructure:"encrypt-connection" json:"encrypt-connection"`
+	RPCEndpoint       string                           `mapstructure:"rpc-endpoint" json:"rpc-endpoint"`
 	WSEndpoint        string                           `mapstructure:"ws-endpoint" json:"ws-endpoint"`
 	MessageContracts  map[string]MessageProtocolConfig `mapstructure:"message-contracts" json:"message-contracts"`
 }
@@ -283,6 +284,23 @@ func constructURL(protocol string, host string, port uint32, encrypt bool) strin
 // Otherwise, constructs the endpoint from the APINodeHost, APINodePort, and EncryptConnection fields,
 // following the /ext/bc/{chainID}/rpc format.
 func (s *DestinationSubnet) GetNodeRPCEndpoint() string {
+	if s.RPCEndpoint != "" {
+		return s.RPCEndpoint
+	}
+	baseUrl := constructURL("http", s.APINodeHost, s.APINodePort, s.EncryptConnection)
+	chainID := s.ChainID
+	subnetID, _ := ids.FromString(s.SubnetID) // already validated in Validate()
+	if subnetID == constants.PrimaryNetworkID {
+		chainID = cChainIdentifierString
+	}
+	return fmt.Sprintf("%s/ext/bc/%s/rpc", baseUrl, chainID)
+}
+
+// Constructs an RPC endpoint for the subnet.
+// If the RPCEndpoint field is set in the configuration, returns that directly.
+// Otherwise, constructs the endpoint from the APINodeHost, APINodePort, and EncryptConnection fields,
+// following the /ext/bc/{chainID}/rpc format.
+func (s *SourceSubnet) GetNodeRPCEndpoint() string {
 	if s.RPCEndpoint != "" {
 		return s.RPCEndpoint
 	}
--- a/config/config.go
+++ b/config/config.go
@@ -299,7 +299,9 @@ func (s *DestinationSubnet) GetNodeRPCEndpoint() string {
 	if subnetID == constants.PrimaryNetworkID {
 		chainID = cChainIdentifierString
 	}
-	return fmt.Sprintf("%s/ext/bc/%s/rpc", baseUrl, chainID)
+	// Save this result for future use
+	s.RPCEndpoint = fmt.Sprintf("%s/ext/bc/%s/rpc", baseUrl, chainID)
+	return s.RPCEndpoint
 }
 
 // Constructs an RPC endpoint for the subnet.
@@ -316,7 +318,9 @@ func (s *SourceSubnet) GetNodeRPCEndpoint() string {
 	if subnetID == constants.PrimaryNetworkID {
 		chainID = cChainIdentifierString
 	}
-	return fmt.Sprintf("%s/ext/bc/%s/rpc", baseUrl, chainID)
+	// Save this result for future use
+	s.RPCEndpoint = fmt.Sprintf("%s/ext/bc/%s/rpc", baseUrl, chainID)
+	return s.RPCEndpoint
 }
 
 // Constructs a WS endpoint for the subnet.
@@ -333,7 +337,9 @@ func (s *SourceSubnet) GetNodeWSEndpoint() string {
 	if subnetID == constants.PrimaryNetworkID {
 		chainID = cChainIdentifierString
 	}
-	return fmt.Sprintf("%s/ext/bc/%s/ws", baseUrl, chainID)
+	// Save this result for future use
+	s.WSEndpoint = fmt.Sprintf("%s/ext/bc/%s/ws", baseUrl, chainID)
+	return s.WSEndpoint
 }
 
 // Get the private key and derive the wallet address from a relayer's configured private key for a given destination subnet.
--- a/config/config.go
+++ b/config/config.go
@@ -274,7 +274,13 @@ func (s *DestinationSubnet) Validate() error {
 	return nil
 }
 
-func constructURL(protocol string, host string, port uint32, encrypt bool) string {
+func constructURL(protocol string, host string, port uint32, encrypt bool, chainIDStr string, subnetIDStr string) string {
+	var protocolPathMap = map[string]string{
+		"http": "rpc",
+		"ws":   "ws",
+	}
+	path := protocolPathMap[protocol]
+
 	if encrypt {
 		protocol = protocol + "s"
 	}
@@ -282,7 +288,11 @@ func constructURL(protocol string, host string, port uint32, encrypt bool) strin
 	if port != 0 {
 		portStr = fmt.Sprintf(":%d", port)
 	}
-	return fmt.Sprintf("%s://%s%s", protocol, host, portStr)
+	subnetID, _ := ids.FromString(subnetIDStr) // already validated in Validate()
+	if subnetID == constants.PrimaryNetworkID {
+		chainIDStr = cChainIdentifierString
+	}
+	return fmt.Sprintf("%s://%s%s/ext/bc/%s/%s", protocol, host, portStr, chainIDStr, path)
 }
 
 // Constructs an RPC endpoint for the subnet.
@@ -293,14 +303,16 @@ func (s *DestinationSubnet) GetNodeRPCEndpoint() string {
 	if s.RPCEndpoint != "" {
 		return s.RPCEndpoint
 	}
-	baseUrl := constructURL("http", s.APINodeHost, s.APINodePort, s.EncryptConnection)
-	chainID := s.ChainID
-	subnetID, _ := ids.FromString(s.SubnetID) // already validated in Validate()
-	if subnetID == constants.PrimaryNetworkID {
-		chainID = cChainIdentifierString
-	}
+
 	// Save this result for future use
-	s.RPCEndpoint = fmt.Sprintf("%s/ext/bc/%s/rpc", baseUrl, chainID)
+	s.RPCEndpoint = constructURL(
+		"http",
+		s.APINodeHost,
+		s.APINodePort,
+		s.EncryptConnection,
+		s.ChainID,
+		s.SubnetID,
+	)
 	return s.RPCEndpoint
 }
 
@@ -312,14 +324,16 @@ func (s *SourceSubnet) GetNodeRPCEndpoint() string {
 	if s.RPCEndpoint != "" {
 		return s.RPCEndpoint
 	}
-	baseUrl := constructURL("http", s.APINodeHost, s.APINodePort, s.EncryptConnection)
-	chainID := s.ChainID
-	subnetID, _ := ids.FromString(s.SubnetID) // already validated in Validate()
-	if subnetID == constants.PrimaryNetworkID {
-		chainID = cChainIdentifierString
-	}
+
 	// Save this result for future use
-	s.RPCEndpoint = fmt.Sprintf("%s/ext/bc/%s/rpc", baseUrl, chainID)
+	s.RPCEndpoint = constructURL(
+		"http",
+		s.APINodeHost,
+		s.APINodePort,
+		s.EncryptConnection,
+		s.ChainID,
+		s.SubnetID,
+	)
 	return s.RPCEndpoint
 }
 
@@ -331,14 +345,16 @@ func (s *SourceSubnet) GetNodeWSEndpoint() string {
 	if s.WSEndpoint != "" {
 		return s.WSEndpoint
 	}
-	baseUrl := constructURL("ws", s.APINodeHost, s.APINodePort, s.EncryptConnection)
-	chainID := s.ChainID
-	subnetID, _ := ids.FromString(s.SubnetID) // already validated in Validate()
-	if subnetID == constants.PrimaryNetworkID {
-		chainID = cChainIdentifierString
-	}
+
 	// Save this result for future use
-	s.WSEndpoint = fmt.Sprintf("%s/ext/bc/%s/ws", baseUrl, chainID)
+	s.WSEndpoint = constructURL(
+		"ws",
+		s.APINodeHost,
+		s.APINodePort,
+		s.EncryptConnection,
+		s.ChainID,
+		s.SubnetID,
+	)
 	return s.WSEndpoint
 }
 
--- a/config/config.go
+++ b/config/config.go
@@ -63,6 +63,7 @@ type Config struct {
 	NetworkID          uint32              `mapstructure:"network-id" json:"network-id"`
 	PChainAPIURL       string              `mapstructure:"p-chain-api-url" json:"p-chain-api-url"`
 	EncryptConnection  bool                `mapstructure:"encrypt-connection" json:"encrypt-connection"`
+	StorageLocation    string              `mapstructure:"storage-location" json:"storage-location"`
 	SourceSubnets      []SourceSubnet      `mapstructure:"source-subnets" json:"source-subnets"`
 	DestinationSubnets []DestinationSubnet `mapstructure:"destination-subnets" json:"destination-subnets"`
 }
@@ -72,6 +73,7 @@ func SetDefaultConfigValues(v *viper.Viper) {
 	v.SetDefault(NetworkIDKey, constants.MainnetID)
 	v.SetDefault(PChainAPIURLKey, "https://api.avax.network")
 	v.SetDefault(EncryptConnectionKey, true)
+	v.SetDefault(StorageLocationKey, "./awm-relayer-storage")
 }
 
 // BuildConfig constructs the relayer config using Viper.
--- a/config/config.go
+++ b/config/config.go
@@ -73,7 +73,7 @@ func SetDefaultConfigValues(v *viper.Viper) {
 	v.SetDefault(NetworkIDKey, constants.MainnetID)
 	v.SetDefault(PChainAPIURLKey, "https://api.avax.network")
 	v.SetDefault(EncryptConnectionKey, true)
-	v.SetDefault(StorageLocationKey, "./awm-relayer-storage")
+	v.SetDefault(StorageLocationKey, "./.awm-relayer-storage")
 }
 
 // BuildConfig constructs the relayer config using Viper.
@@ -102,6 +102,7 @@ func BuildConfig(v *viper.Viper) (Config, bool, error) {
 	cfg.NetworkID = v.GetUint32(NetworkIDKey)
 	cfg.PChainAPIURL = v.GetString(PChainAPIURLKey)
 	cfg.EncryptConnection = v.GetBool(EncryptConnectionKey)
+	cfg.StorageLocation = v.GetString(StorageLocationKey)
 	if err := v.UnmarshalKey(DestinationSubnetsKey, &cfg.DestinationSubnets); err != nil {
 		return Config{}, false, fmt.Errorf("failed to unmarshal destination subnets: %v", err)
 	}
--- a/database/database.go
+++ b/database/database.go
@@ -3,7 +3,7 @@ package database
 import "github.com/ava-labs/avalanchego/ids"
 
 const (
-	LatestBlockHeightKey = "latestBlockHeight"
+	LatestProcessedBlockKey = "latestProcessedBlock"
 )
 
 // RelayerDatabase is a key-value store for relayer state, with each chainID maintaining its own state
--- a/database/database.go
+++ b/database/database.go
@@ -13,8 +13,9 @@ const (
 )
 
 var (
-	ErrKeyNotFound   = errors.New("key not found")
-	ErrChainNotFound = errors.New("no database for chain")
+	ErrKeyNotFound              = errors.New("key not found")
+	ErrChainNotFound            = errors.New("no database for chain")
+	ErrDatabaseMisconfiguration = errors.New("database misconfiguration")
 )
 
 // RelayerDatabase is a key-value store for relayer state, with each chainID maintaining its own state
--- a/database/database.go
+++ b/database/database.go
@@ -3,10 +3,18 @@
 
 package database
 
-import "github.com/ava-labs/avalanchego/ids"
+import (
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/pkg/errors"
+)
 
 const (
 	LatestSeenBlockKey = "latestSeenBlock"
+)
+
+var (
+	ErrKeyNotFound   = errors.New("key not found")
+	ErrChainNotFound = errors.New("no database for chain")
 )
 
 // RelayerDatabase is a key-value store for relayer state, with each chainID maintaining its own state
--- a/database/database.go
+++ b/database/database.go
@@ -2,6 +2,10 @@ package database
 
 import "github.com/ava-labs/avalanchego/ids"
 
+const (
+	LatestBlockHeightKey = "latestBlockHeight"
+)
+
 type RelayerDatabase interface {
 	Get(chainID ids.ID, key []byte) ([]byte, error)
 	Put(chainID ids.ID, key []byte, value []byte) error
--- a/database/database.go
+++ b/database/database.go
@@ -1,0 +1,8 @@
+package database
+
+import "github.com/ava-labs/avalanchego/ids"
+
+type RelayerDatabase interface {
+	Get(chainID ids.ID, key []byte) ([]byte, error)
+	Put(chainID ids.ID, key []byte, value []byte) error
+}
--- a/database/database.go
+++ b/database/database.go
@@ -1,3 +1,6 @@
+// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
 package database
 
 import "github.com/ava-labs/avalanchego/ids"
--- a/database/database.go
+++ b/database/database.go
@@ -9,7 +9,7 @@ import (
 )
 
 const (
-	LatestSeenBlockKey = "latestSeenBlock"
+	LatestProcessedBlockKey = "latestProcessedBlock"
 )
 
 var (
--- a/database/database.go
+++ b/database/database.go
@@ -6,6 +6,7 @@ const (
 	LatestBlockHeightKey = "latestBlockHeight"
 )
 
+// RelayerDatabase is a key-value store for relayer state, with each chainID maintaining its own state
 type RelayerDatabase interface {
 	Get(chainID ids.ID, key []byte) ([]byte, error)
 	Put(chainID ids.ID, key []byte, value []byte) error
--- a/database/database.go
+++ b/database/database.go
@@ -3,7 +3,7 @@ package database
 import "github.com/ava-labs/avalanchego/ids"
 
 const (
-	LatestProcessedBlockKey = "latestProcessedBlock"
+	LatestSeenBlockKey = "latestSeenBlock"
 )
 
 // RelayerDatabase is a key-value store for relayer state, with each chainID maintaining its own state
--- a/database/json_file_storage.go
+++ b/database/json_file_storage.go
@@ -25,16 +25,16 @@ type JsonFileStorage struct {
 	// we predefined the network when creating the storage
 	// also we don't allow adding new network after the storage is created
 	// so we don't need to lock the map
-	mutexes map[string]*sync.RWMutex
+	mutexes map[ids.ID]*sync.RWMutex
 	logger  logging.Logger
 }
 
 // NewJSONFileStorage creates a new JsonFileStorage instance
 // with predefined networks. e.g "ethereum", "avalanche"
-func NewJSONFileStorage(logger logging.Logger, dir string, networks []string) (*JsonFileStorage, error) {
+func NewJSONFileStorage(logger logging.Logger, dir string, networks []ids.ID) (*JsonFileStorage, error) {
 	storage := &JsonFileStorage{
 		dir:     filepath.Clean(dir),
-		mutexes: make(map[string]*sync.RWMutex),
+		mutexes: make(map[ids.ID]*sync.RWMutex),
 		logger:  logger,
 	}
 
@@ -64,7 +64,7 @@ func NewJSONFileStorage(logger logging.Logger, dir string, networks []string) (*
 // Get the latest chain state from the json database, and retrieve the value from the key
 func (s *JsonFileStorage) Get(chainID ids.ID, key []byte) ([]byte, error) {
 	var currentState chainState
-	err := s.read(chainID.String(), &currentState)
+	err := s.read(chainID, &currentState)
 	if err != nil {
 		return nil, err
 	}
@@ -76,7 +76,7 @@ func (s *JsonFileStorage) Get(chainID ids.ID, key []byte) ([]byte, error) {
 func (s *JsonFileStorage) Put(chainID ids.ID, key []byte, value []byte) error {
 	currentState := make(chainState)
 	var currentStateData []byte
-	err := s.read(chainID.String(), &currentStateData)
+	err := s.read(chainID, &currentStateData)
 	if err == nil {
 		err = json.Unmarshal(currentStateData, &currentState)
 		if err != nil {
@@ -86,11 +86,11 @@ func (s *JsonFileStorage) Put(chainID ids.ID, key []byte, value []byte) error {
 
 	currentState[string(key)] = string(value)
 
-	return s.write(chainID.String(), currentState)
+	return s.write(chainID, currentState)
 }
 
 // write value into the file
-func (s *JsonFileStorage) write(network string, v interface{}) error {
+func (s *JsonFileStorage) write(network ids.ID, v interface{}) error {
 	mutex, ok := s.mutexes[network]
 	if !ok {
 		return errors.New("network does not exist")
@@ -99,7 +99,7 @@ func (s *JsonFileStorage) write(network string, v interface{}) error {
 	mutex.Lock()
 	defer mutex.Unlock()
 
-	fnlPath := filepath.Join(s.dir, network+".json")
+	fnlPath := filepath.Join(s.dir, network.String()+".json")
 	tmpPath := fnlPath + ".tmp"
 
 	b, err := json.MarshalIndent(v, "", "\t")
@@ -124,7 +124,7 @@ func (s *JsonFileStorage) write(network string, v interface{}) error {
 }
 
 // Read from disk and unmarshal into v
-func (s *JsonFileStorage) read(network string, v interface{}) error {
+func (s *JsonFileStorage) read(network ids.ID, v interface{}) error {
 	mutex, ok := s.mutexes[network]
 	if !ok {
 		return errors.New("network does not exist")
@@ -133,7 +133,7 @@ func (s *JsonFileStorage) read(network string, v interface{}) error {
 	mutex.RLock()
 	defer mutex.RUnlock()
 
-	path := filepath.Join(s.dir, network+".json")
+	path := filepath.Join(s.dir, network.String()+".json")
 	_, err := os.Stat(path)
 	if err != nil {
 		return err
--- a/database/json_file_storage.go
+++ b/database/json_file_storage.go
@@ -12,7 +12,7 @@ import (
 	"go.uber.org/zap"
 )
 
-var _ RelayerDatabase = &JsonFileStorage{}
+var _ RelayerDatabase = &JSONFileStorage{}
 
 var (
 	fileDoesNotExistErr = errors.New("JSON database file does not exist")
@@ -20,21 +20,21 @@ var (
 
 type chainState map[string]string
 
-// JsonFileStorage implements RelayerDatabase
-type JsonFileStorage struct {
+// JSONFileStorage implements RelayerDatabase
+type JSONFileStorage struct {
 	// the directory where the json files are stored
 	dir string
 
 	// Each network has its own mutex
-	// The chainIDs used to index the JsonFileStorage are created at initialization
+	// The chainIDs used to index the JSONFileStorage are created at initialization
 	// and are not modified afterwards, so we don't need to lock the map itself.
 	mutexes map[ids.ID]*sync.RWMutex
 	logger  logging.Logger
 }
 
-// NewJSONFileStorage creates a new JsonFileStorage instance
-func NewJSONFileStorage(logger logging.Logger, dir string, networks []ids.ID) (*JsonFileStorage, error) {
-	storage := &JsonFileStorage{
+// NewJSONFileStorage creates a new JSONFileStorage instance
+func NewJSONFileStorage(logger logging.Logger, dir string, networks []ids.ID) (*JSONFileStorage, error) {
+	storage := &JSONFileStorage{
 		dir:     filepath.Clean(dir),
 		mutexes: make(map[ids.ID]*sync.RWMutex),
 		logger:  logger,
@@ -64,7 +64,7 @@ func NewJSONFileStorage(logger logging.Logger, dir string, networks []ids.ID) (*
 }
 
 // Get the latest chain state from the json database, and retrieve the value from the key
-func (s *JsonFileStorage) Get(chainID ids.ID, key []byte) ([]byte, error) {
+func (s *JSONFileStorage) Get(chainID ids.ID, key []byte) ([]byte, error) {
 	currentState := make(chainState)
 	fileExists, err := s.read(chainID, &currentState)
 	if err != nil {
@@ -82,7 +82,7 @@ func (s *JsonFileStorage) Get(chainID ids.ID, key []byte) ([]byte, error) {
 }
 
 // Put the value into the json database. Read the current chain state and overwrite the key, if it exists
-func (s *JsonFileStorage) Put(chainID ids.ID, key []byte, value []byte) error {
+func (s *JSONFileStorage) Put(chainID ids.ID, key []byte, value []byte) error {
 	currentState := make(chainState)
 	_, err := s.read(chainID, &currentState)
 	if err != nil {
@@ -99,7 +99,7 @@ func (s *JsonFileStorage) Put(chainID ids.ID, key []byte, value []byte) error {
 }
 
 // write value into the file
-func (s *JsonFileStorage) write(network ids.ID, v interface{}) error {
+func (s *JSONFileStorage) write(network ids.ID, v interface{}) error {
 	mutex, ok := s.mutexes[network]
 	if !ok {
 		return errors.New("network does not exist")
@@ -135,7 +135,7 @@ func (s *JsonFileStorage) write(network ids.ID, v interface{}) error {
 // Read from disk and unmarshal into v
 // Returns a bool indicating whether the file exists, and an error.
 // If an error is returned, the bool should be ignored.
-func (s *JsonFileStorage) read(network ids.ID, v interface{}) (bool, error) {
+func (s *JSONFileStorage) read(network ids.ID, v interface{}) (bool, error) {
 	mutex, ok := s.mutexes[network]
 	if !ok {
 		return false, errors.New("network does not exist")
--- a/database/json_file_storage.go
+++ b/database/json_file_storage.go
@@ -46,8 +46,8 @@ func NewJSONFileStorage(logger logging.Logger, dir string, networks []ids.ID) (*
 
 	_, err := os.Stat(dir)
 	if err == nil {
-		// dir already exist
-		// return the existing storage
+		// Directory already exists.
+		// Return the existing storage.
 		return storage, nil
 	}
 
@@ -64,7 +64,7 @@ func NewJSONFileStorage(logger logging.Logger, dir string, networks []ids.ID) (*
 	return storage, nil
 }
 
-// Get the latest chain state from the json database, and retrieve the value from the key
+// Get the latest chain state from the JSON database, and retrieve the value from the key
 func (s *JSONFileStorage) Get(chainID ids.ID, key []byte) ([]byte, error) {
 	mutex, ok := s.mutexes[chainID]
 	if !ok {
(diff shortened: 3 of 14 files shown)


vms/evm/subscriber.go
```diff
@@ -84,7 +117,87 @@ func (s *subscriber) forwardLogs() {
 			continue
 		}
 		s.log <- *messageInfo
+
+		// Update the database with the latest block height
+		// TODO: This should also be done in a separate goroutine, rather than waiting for warp messages to be processed
```
michaelkaplan13, line 122: Just to clarify here, do you mean "rather than waiting for the write to disk to finish before processing future messages"?

I definitely agree with the TODOs here though, we should address them prior to merging I think. Another edge case to consider is if there are multiple messages in the same block. We should be careful to only update the "latest block height" once we are sure that all messages prior to (and including?) that block have been processed.

relayer/relayer.go
```diff
@@ -105,9 +107,19 @@ func NewRelayer(
 		logger:                   logger,
 	}
 
-	// Start the message router. We must do this before Subscribing for the first time, otherwise we may miss an incoming message
+	// Start the message router. We must do this before Subscribing or Initializing for the first time, otherwise we may miss an incoming message
 	go r.RouteToMessageChannel()
 
+	// Initialize the subscriber. This will poll the node for any logs that match the filter query from the stored block height,
```
michaelkaplan13, line 113: I think "query" is more accurate rather than "poll" in this case.

relayer/relayer.go
```diff
@@ -105,9 +107,19 @@ func NewRelayer(
...
+	// and process the contained warp messages. If initialization fails, continue with normal relayer operation, but log the error.
+	err = sub.Initialize()
```
michaelkaplan13, line 115: I think there is a possibility to still not process messages from all blocks here because the subscription is opened until after we finish `Initialize`, so there could theoretically be new blocks produced/accepted in between those two events. 

I'm not sure how complicated the implementation would be, but I think the desired behavior would be:
1. Open the subscription to be notified about logs in new blocks.
2. Get the current block height on chain.
3. Process messages from any blocks from the "latest processed" height up to the current block height.
4. Start processing any messages received over the subscription.

vms/evm/subscriber.go
```diff
@@ -84,7 +117,87 @@ func (s *subscriber) forwardLogs() {
...
+}
+
+func (s *subscriber) Initialize() error {
+	ethClient, err := ethclient.Dial(s.nodeRPCURL)
+	if err != nil {
+		return err
+	}
+
+	// Get the latest processed block height from the database.
+	heightData, err := s.db.Get(s.chainID, []byte(database.LatestBlockHeightKey))
```
michaelkaplan13, line 138: Should we handle the "file doesn't exist" error case here for the first initialization of the relayer?

We should clearly document the behavior in this case, but I think it likely makes the most sense to treat the case where the file doesn't exist as the first run of the relayer, and not relay any messages prior to it (since that would involve checking each message from the start of the chain).

vms/evm/subscriber.go
cam-schultz: Not quite, I more meant we should update the latest processed height on disk periodically, rather than only doing so when we process a warp message. I was thinking we'd want to avoid processing large gaps of blocks _without_ warp messages on startup. Though thinking about it some more, I don't think that's even necessary, since the `eth_getLogs` call on startup won't return any logs for that large gap anyway.

To illustrate, say there's a warp message at block 1000, and the next warp message is at block 2000, but the relayer is offline and unable to process that second message. On restart, whether or not the stored block heigh is 1000 or 1999, the `eth_getLogs` query will return the same result.

cam-schultz: Good point on updating the latest block height only after all messages have been processed. I'll add that in as part of this PR.

relayer/relayer.go
cam-schultz: Yup, I think that should be fine, as long as we grab the latest processed block from the db before opening the subscription. Message order wouldn't be guaranteed, but that's not part of the relayer spec anyway.

vms/evm/subscriber.go
cam-schultz: If `Initialize` fails for any reason, we simply continue with normal operation, opening subscriptions, relaying messages, etc. So if the file doesn't exist, we don't back-process any warp messages, and only process new incoming ones. OTOH, if the file _does_ exist and has `latestProcessedBlock: 1`, then yes, we'll deliver every warp message from genesis (though I'm sure `eth_getLogs` has some limits built in).

database/json_file_storage.go
```diff
@@ -0,0 +1,172 @@
...
+		storage.logger.Error("failed to create dir",
+			zap.Error(err))
+		return nil, err
+	}
+
+	return storage, nil
+}
+
+// Get the latest chain state from the json database, and retrieve the value from the key
+func (s *JsonFileStorage) Get(chainID ids.ID, key []byte) ([]byte, error) {
```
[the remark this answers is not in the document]
cam-schultz: Yes, that's correct.

vms/evm/subscriber.go
cam-schultz: So in the current iteration of this PR, we actually store the latest _seen_ block, not the latest fully processed block (I will rename the variables and keys to be more clear). What this means is that if the relayer falls over while processing a large number of warp messages from a particular block, we'll re-process all of those messages on restart. That's obviously not the most efficient use of gas, as already delivered messages would be re-broadcast, so I created this ticket to address that: https://github.com/ava-labs/awm-relayer/issues/22. IMO we can save that for a separate PR, since it doesn't impact correctness here.

relayer/relayer.go
cam-schultz: Updated now.

relayer/relayer.go
cam-schultz: Updated.

database/json_file_storage.go
```diff
@@ -0,0 +1,172 @@
...
+var _ RelayerDatabase = &JsonFileStorage{}
+
+var (
+	fileDoesNotExistErr = errors.New("JSON database file does not exist")
+)
+
+type chainState map[string]string
+
+// JsonFileStorage implements RelayerDatabase
+type JsonFileStorage struct {
```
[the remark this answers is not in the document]
cam-schultz: Updated.

database/json_file_storage.go
```diff
@@ -0,0 +1,172 @@
...
+		// dir already exist
+		// return the existing storage
+		return storage, nil
+	}
+
+	// 0755: The owner can read, write, execute.
+	// Everyone else can read and execute but not modify the file.
+	err = os.MkdirAll(dir, 0755)
+	if err != nil {
+		storage.logger.Error("failed to create dir",
```
minghinmatthewlam, line 58: nit: can you use full `directory` for the message

database/json_file_storage.go
```diff
@@ -0,0 +1,172 @@
...
+	return storage, nil
+}
+
+// Get the latest chain state from the json database, and retrieve the value from the key
+func (s *JSONFileStorage) Get(chainID ids.ID, key []byte) ([]byte, error) {
+	currentState := make(chainState)
+	fileExists, err := s.read(chainID, &currentState)
+	if err != nil {
+		s.logger.Error(
+			"failed to read file",
```
minghinmatthewlam, line 72: can you include chainID

minghinmatthewlam: same for put message

database/json_file_storage.go
```diff
@@ -0,0 +1,172 @@
...
+	currentState[string(key)] = string(value)
+
+	return s.write(chainID, currentState)
+}
+
+// write value into the file
+func (s *JSONFileStorage) write(network ids.ID, v interface{}) error {
+	mutex, ok := s.mutexes[network]
+	if !ok {
+		return errors.New("network does not exist")
```
minghinmatthewlam, line 105: can you include which network it is in error

database/json_file_storage.go
```diff
@@ -0,0 +1,172 @@
...
+	fileExists, err := s.read(chainID, &currentState)
+	if err != nil {
+		s.logger.Error(
+			"failed to read file",
+			zap.Error(err),
+		)
+		return nil, err
+	}
+	if !fileExists {
+		return nil, fileDoesNotExistErr
```
minghinmatthewlam, line 78: save for this error message think we should pass in `chainID`

database/json_file_storage.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	return []byte(currentState[string(key)]), nil
```
minghinmatthewlam, line 81: should we check that the key exists before returning?

database/json_file_storage.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	mutex.RLock()
+	defer mutex.RUnlock()
+
+	path := filepath.Join(s.dir, network.String()+".json")
+
+	// If the file does not exist, return false, but do not return an error as this
+	// is an expected case
+	if _, err := os.Stat(path); errors.Is(err, os.ErrNotExist) {
```
minghinmatthewlam, line 151: would `os.ErrNotExist` catch? In the comments and it mentions returning a `*PathError`

minghinmatthewlam: and how should we handle if an error is returned but it's not `os.ErrNoExist`? Right now we continue to read file

vms/evm/subscriber.go
```diff
@@ -84,7 +117,73 @@ func (s *subscriber) forwardLogs() {
...
+	// Queue each of the logs to be processed
+	s.logger.Info(
+		"Processing logs on initialization",
+		zap.String("fromBlockHeight", height.String()),
+		zap.String("toBlockHeight", strconv.Itoa(int(latestBlock))),
+	)
+	for _, log := range logs {
+		messageInfo, err := NewWarpLogInfo(log)
+		if err != nil {
+			s.logger.Info(
```
minghinmatthewlam, line 171: I think this log severity should be higher

minghinmatthewlam: same for in `forwardLog`

vms/evm/subscriber.go
```diff
@@ -118,23 +217,13 @@ func (s *subscriber) Subscribe() error {
 func (s *subscriber) dialAndSubscribe() error {
 	// Dial the configured destination chain endpoint
 	// This needs to be a websocket
-	ethClient, err := ethclient.Dial(s.nodeURL)
+	ethClient, err := ethclient.Dial(s.nodeWSURL)
```
minghinmatthewlam, line 220: - Why do we use websocket for the dial and subscribe, which does `SubscribeFilterLogs`, vs rpc url which does `FilterLogs`? 
- Is the comment meant to be source chain endpoint?

relayer/relayer.go
```diff
@@ -116,6 +123,30 @@ func NewRelayer(
 		)
 		return nil, nil, err
 	}
+
+	// Get the latest processed block height from the database.
```
minghinmatthewlam, line 127: can you move this to a helper similar to `processPreviousBlocks`?

relayer/relayer.go
minghinmatthewlam: RIght now I think we're running `subscribe` and `processFromHeight` in two separate go routines concurrently, so if new blocks came before `processFromHeight` gets back the latest block height, we'd double process that block right?

database/json_file_storage.go
cam-schultz: `os.Stat` returns a `*PathError` that wraps a `os.ErrNotExist` if the file is not found. See here for an example: https://go.dev/play/p/qFwPDKFqSZS

vms/evm/subscriber.go
cam-schultz: `FilterLogs` calls the RPC method `eth_getLogs` which grabs all of the historical logs that match a particular query, while `SubscribeFilterLogs` calls the eth_subscribe method to establish a websocket connection to transmit real time logs as they are finalized on chain.

But yes, that comment should read `source chain`

relayer/relayer.go
cam-schultz: That's correct. I think it would be somewhat tricky to cover all the edge cases here, especially since the `eth_subscribe` and `eth_ blockNumber` calls cannot be done atomically. The alternative I'm currently in favor of is implementing https://github.com/ava-labs/awm-relayer/issues/22 to guard against double processing, rather than doing so by synchronizing the various goroutines.

vms/evm/subscriber.go
cam-schultz: Good call. I also added errors logs to `NewWarpLogInfo`. Even though we're processing logs from an external source, this rises to the level of an error because those logs are pre-filtered in the `eth_getLogs` query such that we should always be able to create a `WarpLogInfo` from them.

database/json_file_storage.go
```diff
@@ -0,0 +1,175 @@
...
+			"failed to read file",
+			zap.String("chainID", chainID.String()),
+			zap.Error(err),
+		)
+		return err
+	}
+
+	currentState[string(key)] = string(value)
+
+	return s.write(chainID, currentState)
```
bernard-avalabs, line 101: Not sure if this is a problem in a use case, but we are not performing an atomic update in between the read and write since we are releasing the lock in between. Concurrent Puts can result in lost updates where one or more writes are lost.

If this can be a problem, then we should grab the lock before the read and release it after the write. We can do that by doing one of the following: move the locking out of read and write (and perform them in the functions that call them), use recursive locks (I don't think they are supported by default in golang), or add a flag to read/write to tell them that a lock is not required for this call.

database/json_file_storage.go
```diff
@@ -0,0 +1,175 @@
...
+	// write marshaled data to the temp file
+	// if write failed, the original file is not affected
+	// 0644 Only the owner can read and write.
+	// Everyone else can only read. No one can execute the file.
+	if err := os.WriteFile(tmpPath, b, 0644); err != nil {
+		return errors.Wrap(err, "failed to write file")
+	}
+
+	// move final file into place
+	if err := os.Rename(tmpPath, fnlPath); err != nil {
```
bernard-avalabs, line 131: Maybe for a future PR: If we want to keep a few old versions of the file, we can implement "log rotation" by renaming the existing file to a backup name (same name but with an added suffix) before overwriting it. Probably not necessary for our use case.

database/json_file_storage.go
```diff
@@ -0,0 +1,175 @@
...
+		return errors.Wrap(err, "failed to rename file")
+	}
+
```
bernard-avalabs, line 134: nit: We may want to consider to performing a sync on the file if we want to be sure it is on disk (and not just in the writeback buffer of the file cache). Not sure if it is necessary in our use case.

bernard-avalabs (commented):
Generally looks good. There may be a concurrency issue (lost updates) by having the locks in read/write. It might be better for them to be in Put and Get instead (I don't think other functions directly use read/write).

database/json_file_storage.go
cam-schultz: I wasn't aware of this until you pointed it out, but Go `File`s do not perform buffered writes, so as soon as `File.Write` is called (as is the case internally in all of the `os` function calls here), the changes are written to disk. See here for a discussion on this: https://stackoverflow.com/a/10890222

database/json_file_storage.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	if !fileExists {
+		return nil, errors.Errorf("file does not exist. chainID: %s", chainID)
+	}
+
+	if _, ok := currentState[string(key)]; !ok {
+		return nil, errors.Errorf("key does not exist. chainID: %s key: %s", chainID, key)
+	}
+
+	return []byte(currentState[string(key)]), nil
```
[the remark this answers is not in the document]
cam-schultz: Modified the logic to only require a sinlge access

config/config.go
```diff
@@ -295,6 +302,23 @@ func (s *DestinationSubnet) GetNodeRPCEndpoint() string {
...
+	if s.RPCEndpoint != "" {
+		return s.RPCEndpoint
+	}
+	baseUrl := constructURL("http", s.APINodeHost, s.APINodePort, s.EncryptConnection)
+	chainID := s.ChainID
+	subnetID, _ := ids.FromString(s.SubnetID) // already validated in Validate()
+	if subnetID == constants.PrimaryNetworkID {
+		chainID = cChainIdentifierString
+	}
+	return fmt.Sprintf("%s/ext/bc/%s/rpc", baseUrl, chainID)
```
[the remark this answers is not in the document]
cam-schultz: Good call. Modified these functions to save the constructed url to the config object.

database/json_file_storage.go
cam-schultz: In our use case, only under extreme circumstances would there be concurrent writes to the same JSON file in such a way that could cause issues. In order for that to happen, we'd have to have logs be [processed](https://github.com/ava-labs/awm-relayer/blob/16af742920ee732323bf3c1430f86d44538a7951/vms/evm/subscriber.go#L123) in such a way that subsequent blocks would overlap in the DB write. Even if that were to occur, the worst that would happen would be a log being processed multiple times, which wouldn't be an issue for correctness. https://github.com/ava-labs/awm-relayer/pull/23 would guard against this as well.

That all being said, having the writes be atomic is a good idea, that eliminates this class of bugs altogether, so I've moved the locking behavior into `Put`/`Get` as you suggested.

vms/evm/subscriber.go
minghinmatthewlam: right now when we do `db.Get` it looks to me if a file does not exist, we still return an error, and when the `relayer.NewRelayer` sees the errror it'll return and not continue processing. 

I think we should return a specific file does not exist error from `db.Get`, and then in `NewRelayer` if we get that specific error back we return without setting `latestSeenBlock` and going through `processFromHeight`

minghinmatthewlam: something like 
```
	// Get the latest processed block height from the database.
	latestSeenBlockData, err := r.db.Get(r.sourceChainID, []byte(database.LatestSeenBlockKey))
	if err != nil {
                if err == FileDoesNotExist {
                      r.logger.Info("First run of relayer for chain ID %s, db file does not exist")
                      return &r, sub, nil 
                }
		r.logger.Warn("failed to get latest block from database", zap.Error(err))
		return nil, nil, err
	}
	latestSeenBlock, success := new(big.Int).SetString(string(latestSeenBlockData), 10)
	if !success {
		r.logger.Error("failed to convert latest block to big.Int", zap.Error(err))
		return nil, nil, err
	}

	// Back-process all warp messages from the latest seen block to the latest block
	// This will query the node for any logs that match the filter query from the stored block height,
	// and process the contained warp messages. If initialization fails, continue with normal relayer operation, but log the error.
	err = sub.ProcessFromHeight(latestSeenBlock)
	if err != nil {
		logger.Warn(
			"Encountered an error when processing historical blocks. Continuing to normal relaying operation.",
			zap.Error(err),
		)
	}

	return &r, sub, nil
```

minghinmatthewlam (commented):
left comment regarding how we handle initial case where the file does not exist, right now IIUC the relayer creation will stop and return error

vms/evm/subscriber.go
cam-schultz: Nice catch. Fixed this now. In the case in which the file doesn't exist, we return the relayer instance, but skip the `sub.ProcessFromHeight` call.

vms/evm/subscriber.go
```diff
@@ -116,25 +228,15 @@ func (s *subscriber) Subscribe() error {
...
-		Topics: [][]common.Hash{
-			{warp.WarpABI.Events["SendWarpMessage"].ID},
-			{},
-			{},
-		},
-		Addresses: []common.Address{
-			warp.ContractAddress,
-		},
-	}
+	filterQuery := warpFilterQuery
```
geoff-vball, line 238: Is having this line functionally different than passing `warpFilterQuery` directly into `ethClient.SubscribeFilterLogs`?

geoff-vball (commented):
Just a couple nits, overall looks good to me!

database/json_file_storage.go
```diff
@@ -0,0 +1,178 @@
...
+			zap.String("chainID", chainID.String()),
+			zap.Error(err),
+		)
+		return nil, err
+	}
+	if !fileExists {
+		return nil, ErrChainNotFound
+	}
+
+	if val, ok := currentState[string(key)]; ok {
```
minghinmatthewlam, line 89: nit: can we make the last return the happy case, and flip this check to `!ok`?

database/json_file_storage.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	return nil, ErrKeyNotFound
+}
+
+// Put the value into the json database. Read the current chain state and overwrite the key, if it exists
+func (s *JSONFileStorage) Put(chainID ids.ID, key []byte, value []byte) error {
+	mutex, ok := s.mutexes[chainID]
+	if !ok {
+		return errors.Errorf("network does not exist. chainID: %s", chainID.String())
```
minghinmatthewlam, line 100: nit: can we add this error in `database.go` since it's used for Get and Put

vms/evm/subscriber.go
cam-schultz: Nope, I think that was leftover from an earlier iteration. Cleaned up now.

cam-schultz:
I discovered an edge case that could cause messages to be missed in the case that the chain DB is not already initialized. The order of events is as follows:
1. There does not exist a chainID.json for a source chain
2. On startup, we do not process any historical logs, and spin up the subscriber to listen for new messages
3. If the relayer goes offline at this point, chainID.json still does not exist, so we may miss messages in the interim.

The [f4d2c83](https://github.com/ava-labs/awm-relayer/pull/20/commits/f4d2c8304af093365f44f66f3221fab3cedd7950) is to always save the latest block height in chainID.json (creating it if it does not exist) on startup, and only process historical logs if we successfully fetch the latest seen block height from the database.

geoff-vball:
@cam-schultz If we're only processing historical logs when we already have a `chainID.json`, that means that any given message can only be sent by a relayer that was already online (at least once) before the message was sent? Is there a corner case where if any teleporter message is sent before any relayers are online, it will never get sent?

cam-schultz:
> @cam-schultz If we're only processing historical logs when we already have a `chainID.json`, that means that any given message can only be sent by a relayer that was already online (at least once) before the message was sent? Is there a corner case where if any teleporter message is sent before any relayers are online, it will never get sent?

That's the intended behavior. The only way to get around that such that any message would be guaranteed to be attempted to be relayed, regardless of relayer status at the time of sending, would for relayers to relayer every message since genesis if the `chainID.json` does not exist. This is overkill, and in any case, a relayer can always set it's `latestSeenBlock` to just before a the message in order to guarantee that message is relayed on startup.

relayer/relayer.go
```diff
@@ -116,6 +124,51 @@ func NewRelayer(
...
+	var (
+		latestSeenBlockData []byte
+		latestSeenBlock     *big.Int // initialized to nil
+	)
+	latestSeenBlockData, err = r.db.Get(r.sourceChainID, []byte(database.LatestSeenBlockKey))
```
michaelkaplan13, line 133: Is this the latest seen block height, or the latest processed block height? We seem to use the two interchangeably in comments/variable names, but I think it makes a big difference in terms of guaranteeing we don't miss any messages because we may not have processed all messages in blocks that we have "seen".

vms/evm/subscriber.go
```diff
@@ -33,33 +36,76 @@ var (
...
 // subscriber implements Subscriber
 type subscriber struct {
-	nodeURL string
-	log     chan vmtypes.WarpLogInfo
-	evmLog  <-chan types.Log
-	sub     interfaces.Subscription
+	nodeWSURL  string
+	nodeRPCURL string
+	chainID    ids.ID
+	log        chan vmtypes.WarpLogInfo
```
michaelkaplan13, line 56: Not new in this PR, but would it be more clear if we named this `logChan`?

vms/evm/subscriber.go
```diff
@@ -75,16 +121,89 @@ func NewWarpLogInfo(log types.Log) (*vmtypes.WarpLogInfo, error) {
...
-			s.logger.Info(
+			s.logger.Error(
 				"Invalid log. Continuing.",
 				zap.Error(err),
 			)
 			continue
 		}
 		s.log <- *messageInfo
+
+		// Update the database with the latest seen block height
```
michaelkaplan13, line 134: Is it correct that we may update the seen block height here prior to all the logs in the block being processed?

vms/evm/subscriber.go
michaelkaplan13: > though I'm sure eth_getLogs has some limits built in

That's a good call out. We should check this to make sure we handle any page limits properly.

vms/evm/subscriber.go
michaelkaplan13: Yeah all of that reasoning sounds good to me. I just want to confirm that we don't update the latest seen block until after all messages in that block have been processed. Otherwise, if we update the latest seen block prior to processing all messages in that block, we may end up missing messages in that block if the relayer hits a failure and restarts.

relayer/relayer.go
cam-schultz: This is the latest **seen** block height, and should be across the entire application. I am not seeing any references to latest _processed_ block height, but please let me know if I missed any.

As far as behavior goes, we want to operate on the latest seen block height so that if the relayer exits mid-block, we can resume from that latest seen block, potentially re-processing warp messages, but not missing any. We store the latest seen block height before processing any of the warp messages contained therein.

vms/evm/subscriber.go
cam-schultz: Yes. This is desired as described here: https://github.com/ava-labs/awm-relayer/pull/20#discussion_r1316422605

vms/evm/subscriber.go
cam-schultz: `eth_getLogs` returns logs from the provided block height _inclusive_. So say we see block 1000, and only process some of the warp messages contained therein. On restart, we'll reprocess block 1000, potentially re processing warp messages, but definitely not missing any.

relayer/relayer.go
```diff
@@ -125,29 +125,42 @@ func NewRelayer(
...
+	// The following cases are treated as successful:
+	// 1) The database contains the latest seen block data for the chain
+	//    - In this case, we parse the block height and process warp logs from that height to the current block
+	// 2) The database has been configured for the chain, but does not contain the latest seen block data
+	//    - In this case, we save the current block height in the database, but do not process any historical warp logs
+	if err == nil {
+		// If the database contains the latest seen block data, then  back-process all warp messages from the
+		// latest seen block to the latest block
+		// This will query the node for any logs that match the filter query from the stored block height,
+		r.logger.Info("latest processed block", zap.String("block", string(latestSeenBlockData)))
```
minghinmatthewlam, line 144: should this log be latest seen block instead

config/config.go
```diff
@@ -292,7 +299,28 @@ func (s *DestinationSubnet) GetNodeRPCEndpoint() string {
...
+	// Save this result for future use
+	s.RPCEndpoint = fmt.Sprintf("%s/ext/bc/%s/rpc", baseUrl, chainID)
+	return s.RPCEndpoint
+}
+
+// Constructs an RPC endpoint for the subnet.
+// If the RPCEndpoint field is set in the configuration, returns that directly.
+// Otherwise, constructs the endpoint from the APINodeHost, APINodePort, and EncryptConnection fields,
+// following the /ext/bc/{chainID}/rpc format.
+func (s *SourceSubnet) GetNodeRPCEndpoint() string {
```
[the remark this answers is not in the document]
cam-schultz: Rewrote the helper function `constructURL` to be more reusable

vms/evm/subscriber.go
cam-schultz: Good idea. Changed.

relayer/relayer.go
cam-schultz: Yup, updated this comment and some others now.

relayer/relayer.go
```diff
@@ -125,29 +125,42 @@ func NewRelayer(
...
+		var success bool
+		latestSeenBlock, success = new(big.Int).SetString(string(latestSeenBlockData), 10)
+		if !success {
+			r.logger.Error("failed to convert latest block to big.Int", zap.Error(err))
+			return nil, nil, err
+		}
+	} else if errors.Is(err, database.ErrChainNotFound) || errors.Is(err, database.ErrKeyNotFound) {
+		// Otherwise, latestSeenBlock is nil, so the call to ProcessFromHeight will simply update the database with the
```
minghinmatthewlam, line 152: I think this if/elseif/else combination with processing from height even when height is nil is confusing. I'd recommend separating a helper for `updateLatestSeenBlock()` or similar, and only calling that for when error is `ErrChainNotFound` or `ErrKeyNotFound`, instead of handling a nil case in `ProcessFromHeight`. Then we can also rearrange the statements to something like:
```
if err != nil:
  if err is errChainNotFound or ErrKeyNotFound
    log latest seen block not found in db, etc
    get block number
    update db with latest seen block
  else
    log warning and return error
parse latest seen block
process block height
return
```

vms/evm/subscriber.go
cam-schultz: It seems like this limit is imposed by the deployment or client, and not by the method itself. Most services limit the number of blocks to 500-2000. I'd recommend limiting it to 200 as a conservative starting point. We can reevaluate this figure or add batched calls to accomodate larger ranges in the future, but IMO this is not an immediate concern.

database/json_file_storage.go
```diff
@@ -0,0 +1,186 @@
...
+		return err
+	}
+
+	currentState[string(key)] = string(value)
+
+	return s.write(chainID, currentState)
+}
+
+// Write the value to the file. The caller is responsible for ensuring proper synchronization
+func (s *JSONFileStorage) write(network ids.ID, v interface{}) error {
```
minghinmatthewlam, line 131: nit: for consistency, is this always a `chainID`? Think network id has other meanings

minghinmatthewlam: same for `read`

database/json_file_storage.go
```diff
@@ -0,0 +1,186 @@
...
+			ErrDatabaseMisconfiguration,
+			fmt.Sprintf("database not configured for chain %s", chainID.String()),
+		)
+	}
+
+	mutex.Lock()
+	defer mutex.Unlock()
+
+	currentState := make(chainState)
+	_, err := s.read(chainID, &currentState)
```
minghinmatthewlam, line 115: can you add a comment either at top of function or near here about if no file previously existed for this `chainID` we won't error and will make a new one

minghinmatthewlam (commented):
Most of the logic LGTM, left some nits and comment about separating `ProcessFromHeight` and updating the db with a helper so we don't need to separately handle `height == nil` case.

relayer/relayer.go
cam-schultz: I added a method to the `Subscriber` interface called `UpdateLatestSeenBlock`. Lmk if that makes things less confusing.

vms/evm/subscriber.go
```diff
@@ -161,58 +162,78 @@ func (s *subscriber) ProcessFromHeight(height *big.Int) error {
...
-			s.logger.Warn(
-				fmt.Sprintf("Requested to process too many blocks. Processing only the most recent %s blocks", MaxBlocksToProcess),
-				zap.String("requestedBlockHeight", height.String()),
-				zap.String("latestBlockHeight", toBlock.String()),
-			)
-			height = big.NewInt(0).Add(toBlock, big.NewInt(-MaxBlocksToProcess))
-		}
+	// Cap the number of blocks to process to MaxBlocksToProcess
+	toBlock := big.NewInt(0).SetUint64(latestBlock)
+	if height.Cmp(big.NewInt(0).Add(toBlock, big.NewInt(-MaxBlocksToProcess))) < 0 {
```
minghinmatthewlam, line 167: sorry missed this in previous reviews, but I'm thinking whether we should error out if there are too many blocks to process. Or, maybe we should let the max amount be configured by the user, in case they do want to just process all of the previous blocks.

Also, what happens if the subscriber sees new logs before the relayer creation finishes? I'm worried that there would be nil reference since the relayer hasnt been set yet and we try to call `relayer.RelayMessage()`

vms/evm/subscriber.go
```diff
@@ -161,58 +162,78 @@ func (s *subscriber) ProcessFromHeight(height *big.Int) error {
...
-		if height.Cmp(big.NewInt(0).Add(toBlock, big.NewInt(-MaxBlocksToProcess))) < 0 {
-			s.logger.Warn(
-				fmt.Sprintf("Requested to process too many blocks. Processing only the most recent %s blocks", MaxBlocksToProcess),
-				zap.String("requestedBlockHeight", height.String()),
-				zap.String("latestBlockHeight", toBlock.String()),
-			)
-			height = big.NewInt(0).Add(toBlock, big.NewInt(-MaxBlocksToProcess))
-		}
+	// Cap the number of blocks to process to MaxBlocksToProcess
+	toBlock := big.NewInt(0).SetUint64(latestBlock)
```
minghinmatthewlam, line 166: nit: can we use `new(big.Int)` instead of `big.NewInt(0)`

vms/evm/subscriber.go
```diff
@@ -161,58 +162,78 @@ func (s *subscriber) ProcessFromHeight(height *big.Int) error {
...
+	if height.Cmp(big.NewInt(0).Add(toBlock, big.NewInt(-MaxBlocksToProcess))) < 0 {
+		s.logger.Warn(
+			fmt.Sprintf("Requested to process too many blocks. Processing only the most recent %s blocks", MaxBlocksToProcess),
+			zap.String("requestedBlockHeight", height.String()),
+			zap.String("latestBlockHeight", toBlock.String()),
+		)
+		height = big.NewInt(0).Add(toBlock, big.NewInt(-MaxBlocksToProcess))
```
minghinmatthewlam, line 173: nit: can we use `Sub(x,y)` instead of adding a negative

vms/evm/subscriber.go
cam-schultz: I'm going to leave the max blocks behavior as is for now, as I think we should step back and discuss how to handle the case in which the relayer is severely behind. We should the tradeoff between reasonable look back limits and interaction among concurrently running relayer instances. I also think the scope of this PR is starting to balloon, and at the moment we're not very unlikely to bump against these limits anyway.

cam-schultz: > Also, what happens if the subscriber sees new logs before the relayer creation finishes? I'm worried that there would be nil reference since the relayer hasnt been set yet and we try to call relayer.RelayMessage()

This won't be an issue since we don't call `RelayMessage` until after `NewRelayer` returns anyway. If a new message is seen before relayer creation finishes, it'll just be queued in the subscriber channel until it is read [here].(https://github.com/ava-labs/awm-relayer/blob/ba3f2fb6c794b05c33c9a46b83966fa03faa12ec/main/main.go#L233)

vms/evm/subscriber.go
cam-schultz: I'd prefer to use the package's constructor rather than a call to `new` directly.

minghinmatthewlam: sgtm no strong pref here. Though is that the standard we should use throughout repos? I'm assuming you prefer the constructor incase it includes some other initialization that `new(big.Int)` wouldn't?

vms/evm/subscriber.go
minghinmatthewlam: sounds good, ye we can make that separate case, can you create an issue to keep track?

Ahh ye covers the case I was worried about.

minghinmatthewlam (dismissed):
LGTM, just the comment about creating issue for max blocks we're willing to look back and process.

vms/evm/subscriber.go
cam-schultz: Yes, in general we should initialize objects using constructors when available, as that leaves any initialization logic up to the type. If no constructor is available, and there are fields to initialize, we should prefer `&T{}` to get the pointer to an object. Only in cases in which we don't want to initialize _anything_ (such as an empty map) should we use `new` and `make`, as they only allocate and do not initialize. Exceptions to these rules are of course possible.

database/json_file_storage.go
bernard-avalabs: I suspect golang's File object is a light wrapper around a file descriptor (created using an _open_ syscall). Writes to it are not buffered at the application level (unlike files opened with _fopen_) so they do not need to be flushed, but they are still written to the file system's buffercache first and later written back to disk.

An explicit sync is not necessary if the system does not abruptly power down. However, if there is a power outage, data in the buffercache that hasn't been written to disk can be lost. Most applications don't bother with calling sync as it hurts performance. However, databases often do, and we are building a type of database here.

With that said, a sync here might not be necessary if we assume that the filesystem will writeback the inode update (for the rename) after the writes to the file. This is probably a good assumption for most filesystems. We may also want to consider adding a checksum to the file to make sure we never use corrupted data.

relayer/relayer.go
```diff
@@ -116,7 +124,60 @@ func NewRelayer(
...
+		err = sub.ProcessFromHeight(latestSeenBlock)
+		if err != nil {
+			logger.Warn(
+				"Encountered an error when processing historical blocks. Continuing to normal relaying operation.",
+				zap.String("chainID", r.sourceChainID.String()),
+				zap.Error(err),
+			)
+		}
+		return &r, sub, nil
+	} else if errors.Is(err, database.ErrChainNotFound) || errors.Is(err, database.ErrKeyNotFound) {
```
bernard-avalabs, line 156: nit: I usually prefer to not have an "else if" if the previous "if" block ends with a return. I find it slightly clearer to start a new "if" block instead. Just a minor stylistic preference, you don't need to change this if you don't agree.

vms/evm/subscriber.go
```diff
@@ -161,58 +162,78 @@ func (s *subscriber) ProcessFromHeight(height *big.Int) error {
...
+		s.logsChan <- *messageInfo
+	}
 
-		// Queue each of the logs to be processed
-		s.logger.Info(
-			"Processing logs on initialization",
-			zap.String("fromBlockHeight", height.String()),
-			zap.String("toBlockHeight", toBlock.String()),
+	// Update the database with the latest seen block height
+	err = s.db.Put(s.chainID, []byte(database.LatestSeenBlockKey), []byte(strconv.FormatUint(latestBlock, 10)))
```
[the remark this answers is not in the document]
bernard-avalabs: I think Geoff is right. We should add a waitgroup or some other synchronization mechanism to signal when it is safe to write the latest seen block to disk.

bernard-avalabs (commented):
Generally looks good. I've left a few comments.

vms/evm/subscriber.go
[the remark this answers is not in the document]
cam-schultz: We discussed this offline, but the same issue also arises in normal operation as well. The summary of the issue is that individual messages (possible from different blocks) are processed by the relayer concurrently, but we always save the latest seen block in the database. If the relayer crashes while processing a message, it is possible that the database will have stored a later block, in which case the relayer will fail to process the message on restart. The fix for this is to have the relayer process messages from the same source chain in serial, rather concurrently, and not update the database until after the message is successfully processed. We can therefore guarantee that when processing a given log, the block before that log is fully processed.

We also identified some followup items to tackle in the future, one of which is to add back in concurrent processing of messages from the same chain.
- https://github.com/ava-labs/awm-relayer/issues/30
- https://github.com/ava-labs/awm-relayer/issues/31

relayer/relayer.go
cam-schultz: Agreed. Updated.

geoff-vball (commented):
The current structure looks good to me! There are a bunch of places in the code where I think the logging could be improved, where we're currently just logging a string, but none of the parameters involved in the call. Not sure if you want to tackle that in this PR or a new one.

geoff-vball:
One final question... I know we confirmed that logs from `eth.subscribe` are guaranteed to come in order, but do we know if the call to `ethClient.FilterLogs` will give the logs in block-sorted order?

geoff-vball (dismissed):
👍 Looks great!

minghinmatthewlam (approved):
LGTM thanks for covering this case

relayer/relayer.go
```diff
@@ -154,32 +190,29 @@ func (r *Relayer) RelayMessage(warpLogInfo *vmtypes.WarpLogInfo, metrics *Messag
...
-	// This allows RouteToMessageChannel to forward inbound messages to the correct channel
-	// Note: It's technically possible to block indefinitely here if the inbound handler is processing an flood of non-relevant app responses. This would occur if
-	// the RLocks are acquired more quickly than they are released in RouteToMessageChannel. However, given that inbound messages are rate limited and that the
-	// RLock is only held for a single map access, this is likely not something we need to worry about in practice.
-	messageResponseChan := make(chan message.InboundMessage, peers.InboundMessageChannelSize)
-	r.messageChanLock.Lock()
-	r.messageChanMap[requestID] = messageResponseChan
-	r.messageChanLock.Unlock()
+	// Increment the request ID to use for this message
+	r.currentRequestID++
```
michaelkaplan13, line 212: It looks like previously we used the current value before incrementing, but now we will increment first and then use the value. I would assume this is inconsequential, but just want to call it out to confirm.

vms/evm/subscriber.go
```diff
@@ -26,40 +30,84 @@ const (
...
 	maxResubscribeAttempts      = 10
+	MaxBlocksToProcess          = 200
 )
 
 var (
 	// Errors
 	ErrInvalidLog = errors.New("invalid warp message log")
 )
 
+// The filter query used to match logs emitted by the Warp precompile
```
michaelkaplan13, line 41:
```suggestion
// The filter query used to match logs emitted by the Warp precompile.
```

michaelkaplan13: Actually, looks like we haven't been 100% consistent with punctuation to date. I don't have strong preference either way, but lets decide on a convention.

vms/evm/subscriber.go
```diff
@@ -69,22 +117,150 @@ func NewWarpLogInfo(log types.Log) (*vmtypes.WarpLogInfo, error) {
...
 			)
 			continue
 		}
-		s.log <- *messageInfo
+		s.logsChan <- *messageInfo
+	}
+}
+
+// Process logs from the given block height to the latest block
+// Cap the number of blocks requested from the client to MaxBlocksToProcess,
```
michaelkaplan13, line 140: May want to consider adding a TODO to remove the cap on the number of blocks processed here for now, given that we plan to follow up with that subsequent update.

vms/evm/subscriber.go
```diff
@@ -69,22 +117,150 @@ func NewWarpLogInfo(log types.Log) (*vmtypes.WarpLogInfo, error) {
...
+		return err
+	}
+
+	// Sort the logs in ascending block order. Order logs by index within blocks
+	sort.SliceStable(logs, func(i, j int) bool {
+		if logs[i].BlockNumber == logs[j].BlockNumber {
+			return logs[i].TxIndex < logs[j].TxIndex
+		}
+		return logs[i].BlockNumber < logs[j].BlockNumber
+	})
```
michaelkaplan13, line 204: There can potentially be multiple logs in a single transaction. I think we need to handle that case by checking the log index.

michaelkaplan13 (commented):
I have a few clean up suggestions and questions, but generally looks good to me.

peers/external_handler.go
```diff
@@ -71,17 +71,8 @@ func NewRelayerExternalHandler(
...
-// once the cross-chain message is delivered. Messages are routed through the resulting tree structure:
-//
-// 								  TeleporterRelayer/source subnet       teleporterMessageRelayer/Teleporter message
-//
-//                       		{ TeleporterRelayer.responseChan --...  { teleporterMessageRelayer.messageResponseChan (consumer)
-// HandleInbound (producer) --->{ TeleporterRelayer.responseChan ------>{ teleporterMessageRelayer.messageResponseChan (consumer)
-//                       		{			  ...					    {						...
-//                       		{ TeleporterRelayer.responseChan --...  { teleporterMessageRelayer.messageResponseChan (consumer)
-
+// On startup, one Relayer goroutine is created per source subnet, which listens to the subscriber for cross-chain messages
```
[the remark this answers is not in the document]
cam-schultz: Here, it's capitalized to denote the type `Relayer`. Elsewhere when using the term less precisely, it is not capitalized.

vms/evm/subscriber.go
cam-schultz: Agreed. I'm going to leave the comments as is for now, then circle back across the whole repo.

relayer/relayer.go
```diff
@@ -188,15 +221,28 @@ func (r *Relayer) RelayMessage(warpLogInfo *vmtypes.WarpLogInfo, metrics *Messag
...
+			"Failed to run message relayer",
+			zap.String("chainID", r.sourceChainID.String()),
+			zap.String("warpMessageID", warpMessageInfo.WarpUnsignedMessage.ID().String()),
+			zap.Error(err),
+		)
+		return err
+	}
+
+	// Update the database with the latest seen block height
+	// We cannot store the latest processed block height, because we do not know if a given log is the last log in a block
```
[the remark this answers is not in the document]
cam-schultz: What if we just removed this comment and added the comment you mentioned? It sounds like the potential source of confusion would be the use of both terms, not necessarily one over the other.

vms/evm/subscriber.go
cam-schultz: We should just be able to order by log index instead of tx index. I've made this change.

relayer/relayer.go
cam-schultz: Good catch. It shouldn't matter, since the request ID is only used by the relayer in memory, but I've moved the increment to after it's used to remain consistent.

database/json_file_storage.go
```diff
@@ -0,0 +1,187 @@
...
```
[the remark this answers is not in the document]
cam-schultz: Good point. I added a field for the current chain state, which Put writes to before writing the whole object to disk. We can of course use the cached state in `Get` as well, but I think we should move that to its own issue.

vms/evm/subscriber.go
cam-schultz: I think that cap will still be used in the future if we paginate calls to `eth_getLogs`. I'll leave this comment as is, and have captured that future work [here](https://github.com/ava-labs/awm-relayer/issues/29)

geoff-vball (dismissed):

bernard-avalabs (dismissed):
LGTM

database/database.go
```diff
@@ -0,0 +1,25 @@
...
+
+var (
+	ErrKeyNotFound              = errors.New("key not found")
+	ErrChainNotFound            = errors.New("no database for chain")
+	ErrDatabaseMisconfiguration = errors.New("database misconfiguration")
+)
+
+// RelayerDatabase is a key-value store for relayer state, with each chainID maintaining its own state
+type RelayerDatabase interface {
+	Get(chainID ids.ID, key []byte) ([]byte, error)
```
gwen917, line 23: Why do we return []byte instead of string directly? When we write in to the file, we convert []byte to string. Also, when we use the result, we convert []byte to string.

gwen917 (dismissed):

database/database.go
cam-schultz: We discussed this offline, but for visibility, I chose `[]byte` in the interface so that the keys and values could be as general as possible. The underlying JSON databse uses strings so that the resulting file is human readable.

relayer/relayer.go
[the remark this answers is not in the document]
michaelkaplan13: I generally think that "processed" is a more accurate name than "seen" in this case given that the relayer may see a block but not store it in its database for a couple of reason such as the block not having any warp messages or an error in relaying a message from the block. 

That being said, I don't think it's too significant either way. Will leave up to you if you think it's worth updating or not 👍

michaelkaplan13 (approved):

relayer/relayer.go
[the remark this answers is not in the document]
cam-schultz: No problem. Renamed all variables, symbols, and comments accordingly.

vms/subscriber.go
```diff
@@ -20,8 +20,8 @@ type Subscriber interface {
 	// ProcessFromHeight processes events from {height} to the latest block
 	ProcessFromHeight(height *big.Int) error
 
-	// UpdateLatestSeenBlock retrieves the latest block from the chain and updates the database
-	UpdateLatestSeenBlock() error
+	// UpdateLatestProcessedBlock retrieves the latest block from the chain and updates the database
+	UpdateLatestProcessedBlock() error
```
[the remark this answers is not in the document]
cam-schultz: I'm more in favor of omitting `Initial` from the function name, since the interface imposes no requirement that this function be called at any particular time in the object lifecycle. I agree it could be more descriptive as to what it's actually updating the database value _to_. I've renamed it `SetProcessedBlockHeightToLatest`; let me know if you think another name more accurately reflects the behavior.

relayer/relayer.go
```diff
@@ -116,34 +115,71 @@ func NewRelayer(
...
+	// If neither of the above conditions are met, then we return an error
+	r.logger.Warn(
+		"failed to get latest block from database",
+		zap.String("chainID", r.sourceChainID.String()),
+		zap.Error(err),
+	)
+	return nil, nil, err
 }
 
 // RelayMessage relays a single warp message to the destination chain. Warp message relay requests are concurrent with each other,
```
bernard-avalabs, line 175: "Warp message relay requests are concurrent with each other"

Does this need to be updated after the recent change to process messages from a given source serially?

bernard-avalabs (approved):
LGTM

vms/subscriber.go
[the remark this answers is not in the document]
michaelkaplan13: Nice, yeah totally agree. Good call 🙏

relayer/relayer.go
cam-schultz: Nice catch - updated now

michaelkaplan13 (dismissed):
LGTM!

geoff-vball (dismissed):

geoff-vball (approved):

michaelkaplan13 (approved):

PR #21, opened 2023-08-30
Unit tests

### LINKED ISSUE #11: Add unit tests
**Context and scope**
Unit test coverage is sparse. We should strive for full test coverage in everything below the application entry point.

**Open questions**
For mocks, should we try to use `mockgen` or any other standardized approach for this?

minghinmatthewlam:
## Why this should be merged
Adding test coverage for relayer repo and fixes #11 

cc @geoff-vball for prior contribution to unit tests as well.

## How this works
- Replaces `assert` with `require` checks, since require fails on the first error and makes it easier to debug
- `mockgen` binary to generate mocks for interfaces

## How this was tested
- CI 

## How this was documented
README update


messages/teleporter/config_test.go
cam-schultz, line 1: Can you add license header comments to the new files? I think we're ok to omit them for the generated mock files, since the code under license is in the input to mockgen, not the output.

utils/utils.go
```diff
@@ -26,8 +26,7 @@ const (
 )
 
 var (
-	Uint256Max = (&big.Int{}).SetBytes(common.Hex2Bytes("FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF"))
-
+	MaxHashInputLength = 32
```
cam-schultz, line 29: We can remove this and use `common.HashLength` instead

utils/utils.go
```diff
@@ -38,15 +37,15 @@ var (
...
+func CheckStakeWeightExceedsThreshold(accumulatedSignatureWeight *big.Int, totalWeight uint64, quorumNumerator uint64, quorumDenominator uint64) bool {
 	if accumulatedSignatureWeight == nil {
 		return false
 	}
 
 	// Verifies that quorumNum * totalWeight <= quorumDen * sigWeight
-	scaledTotalWeight := new(big.Int).SetUint64(totalWeight)
-	scaledTotalWeight.Mul(scaledTotalWeight, new(big.Int).SetUint64(DefaultQuorumNumerator))
-	scaledSigWeight := new(big.Int).Mul(accumulatedSignatureWeight, new(big.Int).SetUint64(DefaultQuorumDenominator))
+	totalWeightBI := new(big.Int).SetUint64(totalWeight)
```
cam-schultz, line 46: Is there a specific reason to introduce the new variable `totalWeightBI`? When compared to operating on `scaledTotalWeight` in place, there's an additional allocation incurred.

utils/utils.go
```diff
@@ -58,16 +57,18 @@ func CheckStakeWeightExceedsThreshold(accumulatedSignatureWeight *big.Int, total
 
 // BigToHashSafe ensures that a bignum value is able to fit into a 32 byte buffer before converting it to a common.Hash
 // Returns an error if overflow/truncation would occur by trying to perfom this operation.
+// TODO is this function still needed? It works for negative numbers now.
```
cam-schultz, line 60: Looking at `common.BigToHash`, I think we can remove this function. `big.Int` doesn't guarantee the internal bytes will be representable as a uint256 (though the getters will truncate to the requested type), so we may want to add redundant validity checks at the callsites.

cam-schultz (commented):
A handful of comments, but this generally looks good. Great to see mocks being leveraged in the unit tests 👍

utils/utils.go
geoff-vball: I made this change to make the code a little more clear. Previously we were assigning `scaledTotalWeight := new(big.Int).SetUint64(totalWeight)`, but it didn't actually represent the scaled total weight until the operation on the next line.

If we think the performance hit of the allocation is a big deal, we can change it back. We could also eliminate another allocation on line 48 by re-using `scaledSigWeight`.

utils/utils.go
minghinmatthewlam: > (though the getters will truncate to the requested type), so we may want to add redundant validity checks at the callsites.

can you clarify this, didn't get it. We aren't using `BigToHashSafe` anywhere anyways, so I think we can remove if we don't imagine using it.

messages/teleporter/message_test.go
```diff
@@ -35,49 +35,41 @@ func testTeleporterMessage(messageID int64) TeleporterMessage {
...
-	return hashes, err
+func packSendCrossChainMessageEvent(destinationChainID common.Hash, message TeleporterMessage) ([]common.Hash, []byte, error) {
+	return EVMTeleporterContractABI.PackEvent("SendCrossChainMessage", destinationChainID, message.MessageID, message)
 }
 
 func TestPackUnpackTeleporterMessage(t *testing.T) {
-	message := testTeleporterMessage(4)
+	var (
+		messageID          int64 = 4
+		destinationChainID       = common.HexToHash("0x03")
```
geoff-vball, line 45: nit: formatting here is off

vms/evm/contract_message_test.go
```diff
@@ -62,14 +61,11 @@ func TestUnpack(t *testing.T) {
...
-		if err != nil {
-			t.Errorf("failed to decode test input: %v", err)
-		}
+		require.NoErrorf(t, err, "failed to decode test input: %v", err)
+
 		msg, err := m.UnpackWarpMessage(input)
-		if err != nil {
-			t.Errorf("failed to unpack message: %v", err)
-		}
+		require.NoErrorf(t, err, "failed to unpack message: %v", err)
```
geoff-vball, line 67: I don't think the two comments here are necessary either.

geoff-vball (commented):
Just a few small comments. This looks good! Thanks for taking this over 🙏

config/config_test.go
```diff
@@ -118,8 +118,10 @@ func TestGetDestinationRPCEndpoint(t *testing.T) {
 	}
 
 	for i, testCase := range testCases {
-		res := testCase.s.GetNodeRPCEndpoint()
-		require.Equal(t, testCase.expectedResult, res, fmt.Sprintf("test case %d failed", i))
+		t.Run(fmt.Sprintf("test_%d", i), func(t *testing.T) {
+			res := testCase.s.GetNodeRPCEndpoint()
+			require.Equal(t, testCase.expectedResult, res)
+		})
```
geoff-vball, line 124: I do like this format a lot more. If we can add names to the tests describing what they do, that would be even better than a number.

utils/utils_test.go
```diff
@@ -45,16 +51,16 @@ func TestConvertProtocol(t *testing.T) {
...
-		}
+		t.Run(fmt.Sprintf("test_%d", i), func(t *testing.T) {
+			actualUrl, err := ConvertProtocol(testCase.urlString, testCase.protocol)
+
+			if testCase.expectedError {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+			}
+			require.Equal(t, testCase.expectedUrl, actualUrl)
```
gwen917, line 62: do we want to move this line into the else block? since we only care about the return value when there is no error.

vms/evm/contract_message_test.go
```diff
@@ -46,9 +46,8 @@ func createUnsignedMessage() *warp.UnsignedMessage {
 }
 
 func TestUnpack(t *testing.T) {
-	ctrl := gomock.NewController(t)
-
-	m := NewContractMessage(logging.NewMockLogger(ctrl), config.SourceSubnet{})
+	ctrl := logging.NewMockLogger(gomock.NewController(t))
```
gwen917, line 49: nit: rename `ctrl` to `mockLogger`

vms/evm/contract_message_test.go
```diff
@@ -62,14 +61,11 @@ func TestUnpack(t *testing.T) {
...
-		if err != nil {
-			t.Errorf("failed to decode test input: %v", err)
-		}
+		require.NoError(t, err)
+
 		msg, err := m.UnpackWarpMessage(input)
-		if err != nil {
-			t.Errorf("failed to unpack message: %v", err)
-		}
+		require.NoError(t, err)
```
gwen917, line 67: Can we test error case?

vms/evm/destination_client.go
```diff
@@ -29,6 +31,11 @@ const (
 	MaxPriorityFeePerGas = 2500000000 // 2.5 gwei
 )
 
+// Client interface wraps the ethclient.Client interface for mocking purposes.
+type Client interface {
+	ethclient.Client
```
gwen917, line 36: Can we mock the ethclient.Client directly?

minghinmatthewlam: if we mock the interface directly from the imported pkg we have to use mock reflect, which I didn't see a way to still work with `go generate`, so prefer to leave as is.

utils/utils.go
minghinmatthewlam: I think the extra allocation for clarify in variable naming is okay

utils/utils.go
minghinmatthewlam: hmm rereading this comment, I actually prefer leaving this utils function. Right now we don't use `common.BigTohash` or the `BigToHashSafe`, but want to leave for future use, since `common.BigToHash` will just truncate it to 32 bytes, I think it's better to check for error.

README.md
```diff
@@ -37,3 +37,15 @@ The relayer consists of the following components:
 <div align="center">
   <img src="resources/relayer-diagram.png?raw=true">
 </div>
+
+## Testing
+
+---
```
michaelkaplan13, line 43: Nit: The horizontal line will be included below the headings automatically. It looks like we've doubled up in this case. Could we remove the "---" from the README here and above?

gwen917 (dismissed):
LGTM

michaelkaplan13 (dismissed):
All of the tests and mocks added here LGTM. My only question would be if we want to include test coverage reports/requirements such that we can track our test coverage going forward and ensure it's at a level we feel comfortable with.

geoff-vball (dismissed):
LGTM! 👍

michaelkaplan13 (approved):

minghinmatthewlam:
> All of the tests and mocks added here LGTM. My only question would be if we want to include test coverage reports/requirements such that we can track our test coverage going forward and ensure it's at a level we feel comfortable with.

I'm definitely for printing out a report on our test coverage, not sure fi we'd want to require it to above a minimum threshold. Right now our % of test coverage is still low, and want to avoid tackling it all in this PR. Would be good with first printing out the test coverage %, thoughts @michaelkaplan13 ?

michaelkaplan13:
> > All of the tests and mocks added here LGTM. My only question would be if we want to include test coverage reports/requirements such that we can track our test coverage going forward and ensure it's at a level we feel comfortable with.
> 
> I'm definitely for printing out a report on our test coverage, not sure fi we'd want to require it to above a minimum threshold. Right now our % of test coverage is still low, and want to avoid tackling it all in this PR. Would be good with first printing out the test coverage %, thoughts @michaelkaplan13 ?

Yeah I agree, it's best left for a separate PR either way. 👍

minghinmatthewlam:
Created new issue to address test coverage: https://github.com/ava-labs/awm-relayer/issues/44

gwen917 (approved):
LGTM👍

Issue #22, opened 2023-08-30
Do not relay messages that have already been delivered


cam-schultz:
**Context and scope**
When https://github.com/ava-labs/awm-relayer/issues/14 is implemented, it will be possible for relayers to attempt to relay messages that have already been delivered. We should first check if the message (as specificied by the origin chain ID and Teleporter message ID) has already been delivered, and stop early if so.

**Discussion and alternatives**
The most likely place for this will be in `MessageManager:ShouldSendMessage`

PR #23, opened 2023-08-31
Check if message already delivered

cam-schultz:
## Why this should be merged
- Checks if the Teleporter message has already been delivered before aggregating signatures & constructing the transaction. 
- Gives `MessageManager` and `DestinationClient` interfaces more clearly defined roles.
- Fixes https://github.com/ava-labs/awm-relayer/issues/22

## How this works
- Adds a call to the `TeleporterMessenger` contract method `messageReceived` (https://github.com/ava-labs/teleporter/blob/main/contracts/src/Teleporter/TeleporterMessenger.sol#L492) in `MessageManager:ShouldSendMessage`. If the message has already been received, return early
- Moves message delivery decision logic out of `DestinationClient` and into `MessageManager`. `DestinationClient` now only manages interaction with the destination chain via the relayer EOA, while `MessageManager` manages protocol unpacking and delivery decision logic.

## How this was tested
CI, Teleporter integration tests, & manual testing.


messages/teleporter/message_manager.go
```diff
@@ -87,13 +109,59 @@ func (m *messageManager) ShouldSendMessage(warpMessageInfo *vmtypes.WarpMessageI
...
+	if err != nil {
+		return false, err
+	}
+	if received {
+		m.logger.Info(
+			"Message already deliverd to destination.",
+			zap.String("destinationChainID", destinationChainID.String()),
+			zap.String("teleporterMessageID", teleporterMessage.MessageID.String()),
+		)
+		return false, nil
```
bernard-avalabs, line 162: It looks like the caller should look at the err to distinguished between the message has been received and error conditions. Do the callers do that currently? How should the caller handle errors?

bernard-avalabs (commented):
Generally looks good to me. I left a question about error handling for the receive message check.

messages/teleporter/message_manager.go
```diff
@@ -70,6 +74,24 @@ func NewMessageManager(
 	}, nil
 }
 
+func isDestination(messageDestinationID ids.ID, allowedDestinationID ids.ID) bool {
+	return messageDestinationID == allowedDestinationID
+}
+
+func isAllowedRelayer(allowedRelayers []common.Address, eoa common.Address) bool {
```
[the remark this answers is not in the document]
cam-schultz: At the code level, this particular check is ensuring that the relayer EOA that will send the tx to the destination is included in the Teleporter message's list of allowed relayers. The function above checks that the EVM client that will send the tx is sending to the destination chain ID included in the Teleporter message.

At the architectural level, `DestinationClient` is the entity responsible for sending the tx to the correct destination chain. `MessageRelayer`s should be composed of exactly one `DestinationClient`, so in that sense, relayers are destination specific.

messages/teleporter/message_manager.go
cam-schultz: The call to `ShouldSendMessage` is valid regardless of the value of the boolean, as long as the error is nil. We perform that check here: https://github.com/ava-labs/awm-relayer/blob/2060ab680cbd1a2501a87ac3ec01d6cd9da79f37/relayer/message_relayer.go#L80

I added a comment to the `MessageManager` interface noting that the boolean should be ignored if a error is returned.

messages/teleporter/message.go
```diff
@@ -34,6 +35,13 @@ type ReceiveCrossChainMessageInput struct {
 	RelayerRewardAddress common.Address `json:"relayerRewardAddress"`
 }
 
+// MessageReceivedInput is the input to the MessageReceived
+// in the contract deployed on the receiving chain
+type MessageReceivedInput struct {
+	OriginChainID ids.ID  `json:"relayerRewardAddress"`
+	MessageID	 *big.Int `json:"messageID"`
```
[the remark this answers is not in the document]
cam-schultz: Fixed

messages/teleporter/message_manager.go
```diff
@@ -87,13 +109,59 @@ func (m *messageManager) ShouldSendMessage(warpMessageInfo *vmtypes.WarpMessageI
 	if !ok {
 		return false, fmt.Errorf("relayer not configured to deliver to destination. destinationChainID=%s", destinationChainID.String())
 	}
-	if !destinationClient.Allowed(destinationChainID, teleporterMessage.AllowedRelayerAddresses) {
+	senderAddress := destinationClient.SenderAddress()
+	if !isAllowedRelayer(teleporterMessage.AllowedRelayerAddresses, senderAddress) {
+		m.logger.Info("Relayer EOA not allowed to deliver this message.")
```
michaelkaplan13, line 114: Should we include the message ID in the log here?

messages/teleporter/message_manager.go
```diff
@@ -87,13 +109,59 @@ func (m *messageManager) ShouldSendMessage(warpMessageInfo *vmtypes.WarpMessageI
...
+		return false, nil
+	}
+	if !isDestination(destinationChainID, destinationClient.DestinationChainID()) {
```
michaelkaplan13, line 117: This check seems redundant given that we look up the destination client by the chain ID already. If we choose to keep it, could we move this to right after the `destinationClient` is looked up from the `destinationClients` map?

messages/teleporter/message_manager.go
```diff
@@ -87,13 +109,59 @@ func (m *messageManager) ShouldSendMessage(warpMessageInfo *vmtypes.WarpMessageI
...
 		m.logger.Info(
-			"Relayer not allowed to deliver to chain.",
+			"Destination chain ID for message not supported by relayer.",
 			zap.String("destinationChainID", destinationChainID.String()),
 		)
 		return false, nil
 	}
+
+	// Check if the message has already been delivered to the destination chain
```
michaelkaplan13, line 125: Could we move the section below to a `messageAlreadyDelivered` helper? (or whatever name you think is best)

michaelkaplan13 (commented):
Makes sense to me at a high level. I like the `ShouldSendMessage` interface a lot. I left a few minor suggestions.

messages/teleporter/message_manager.go
cam-schultz: Added

messages/teleporter/message_manager.go
cam-schultz: Agreed. I've removed it altogether, but left the corresponding method in `DestinationClient`, as it makes sense as a getter for that interface.

messages/teleporter/message_manager.go
cam-schultz: Sure thing, changed

messages/teleporter/message.go
```diff
@@ -34,6 +35,13 @@ type ReceiveCrossChainMessageInput struct {
 	RelayerRewardAddress common.Address `json:"relayerRewardAddress"`
 }
 
+// MessageReceivedInput is the input to the MessageReceived
+// in the contract deployed on the receiving chain
```
minghinmatthewlam, line 39:
```suggestion
// MessageReceivedInput is the input to the MessageReceived call
// in the contract deployed on the destination chain
```

messages/teleporter/message_manager.go
```diff
@@ -70,6 +74,24 @@ func NewMessageManager(
 	}, nil
 }
 
+func isDestination(messageDestinationID ids.ID, allowedDestinationID ids.ID) bool {
```
minghinmatthewlam, line 77: dont see us using this, can you also change it to `isDestinationChain`

messages/teleporter/message_manager.go
```diff
@@ -87,18 +109,95 @@ func (m *messageManager) ShouldSendMessage(warpMessageInfo *vmtypes.WarpMessageI
...
+		m.logger.Error(
+			"Failed calling messageReceived method on destination chain.",
+			zap.String("destinationChainID", destinationChainID.String()),
+		)
+		return false, err
+	}
+	// check the contract call result
+	delivered, err := unpackMessageReceivedResult(result)
+	if err != nil {
+		return false, err
```
minghinmatthewlam, line 193: can we add a log error here

messages/teleporter/message_manager.go
minghinmatthewlam: just saw Michael's comment, agree we can remove the check, can we remove this helper function?

minghinmatthewlam (commented):
LGTM, just nits for changing comment and removing `isDestination` since not used

messages/teleporter/message_manager.go
```diff
@@ -87,18 +109,95 @@ func (m *messageManager) ShouldSendMessage(warpMessageInfo *vmtypes.WarpMessageI
...
+	}
+	if delivered {
+		return true, nil
+	}
+	return false, nil
+}
+
 // SendMessage extracts the gasLimit and packs the call data to call the receiveCrossChainMessage method of the Teleporter contract,
 // and dispatches transaction construction and broadcast to the destination client
 func (m *messageManager) SendMessage(signedMessage *warp.Message, parsedVmPayload []byte, destinationChainID ids.ID) error {
```
geoff-vball, line 199: I know it's not part of this PR, but the two declarations below here in the `var` block are not needed.

messages/teleporter/message_manager.go
cam-schultz: Removed

geoff-vball (dismissed):
Left one small comment on something mostly unrelated. The PR looks good to me 👍

messages/teleporter/message_manager.go
cam-schultz: Sdded, and also improved logging in the rest of this file.

messages/teleporter/message_manager.go
cam-schultz: Removed. Those were leftover from a previous iteration where values were being set in an inner scope and I needed to avoid shadowing.

geoff-vball (dismissed):
LGTM!

messages/teleporter/message_manager.go
```diff
@@ -87,36 +105,126 @@ func (m *messageManager) ShouldSendMessage(warpMessageInfo *vmtypes.WarpMessageI
...
+	if err != nil {
+		m.logger.Error(
+			"Failed packing messageReceived call data.",
+			zap.String("destinationChainID", destinationChainID.String()),
+			zap.Error(err),
+		)
+		return false, err
+	}
+	protocolAddress := common.BytesToAddress(m.protocolAddress[:])
+	callMessage := interfaces.CallMsg{
```
gwen917, line 182: you mentioned that gas params are not need for eth_call [here](https://github.com/ava-labs/cctp-integration/pull/61#discussion_r1323596017). Base on [eth_call API](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_call), it seems that the gas parameter is used for eth_call itself not for the transaction we want to send. Is it correct?

cam-schultz: `eth_call` won't consume any gas, so gas params are generally unneeded. One exception is if a function implements control flow based on gas parameters (for example by calling `gasLeft`), but that's not the case for the Teleporter contract.

gwen917 (dismissed):

messages/teleporter/message_manager.go
```diff
@@ -87,36 +105,126 @@ func (m *messageManager) ShouldSendMessage(warpMessageInfo *vmtypes.WarpMessageI
...
 }
 
+// Helper to check if a message has been delivered to the destination chain
+// Returns true if the message has been delivered, false if not
+// On error, the boolean result should be ignored
+func (m *messageManager) messageDelivered(
+	destinationClient vms.DestinationClient,
+	warpMessageInfo *vmtypes.WarpMessageInfo,
+	teleporterMessage *TeleporterMessage,
+	senderAddress common.Address,
```
michaelkaplan13, line 157: Nit: I think `senderAddress` is an optional parameter for calling a contract off-chain outside of the transaction. In our case, it doesn't matter who is calling `messageReceived`, so we can likely remove this as a parameter here all together.

messages/teleporter/message_manager.go
```diff
@@ -87,36 +105,126 @@ func (m *messageManager) ShouldSendMessage(warpMessageInfo *vmtypes.WarpMessageI
...
+			"Failed unpacking messageReceived result.",
+			zap.String("destinationChainID", destinationChainID.String()),
+			zap.Error(err),
+		)
+		return false, err
+	}
+	if delivered {
+		return true, nil
+	}
+	return false, nil
```
michaelkaplan13, line 209:
```suggestion
	return delivered, nil
```

geoff-vball (approved):

michaelkaplan13 (approved):
LGTM!

messages/teleporter/message_manager.go
cam-schultz: Removed 👍

PR #25, opened 2023-09-01
Portable BLST

cam-schultz:
## Why this should be merged
Fixes cross-platform application and unit test execution by adding portable BLST flags. Otherwise, execution fails with `SIGILL`

## How this works
- Adds portable BLST flags to `constants.sh`
- Adds test entrypoint script `test.sh` that sources `constants.sh`

## How this was tested
CI, Teleporter integration tests

### DIFF
--- a/scripts/constants.sh
+++ b/scripts/constants.sh
@@ -22,3 +22,12 @@ relayer_dockerhub_repo=${DOCKER_REPO:-"awm-relayer"}
 current_branch=$(git symbolic-ref -q --short HEAD || git describe --tags --exact-match || true)
 
 git_commit=${RELAYER_COMMIT:-$( git rev-list -1 HEAD )}
+
+# Set the CGO flags to use the portable version of BLST
+#
+# We use "export" here instead of just setting a bash variable because we need
+# to pass this flag to all child processes spawned by the shell.
+export CGO_CFLAGS="-O -D__BLST_PORTABLE__"
+# While CGO_ENABLED doesn't need to be explicitly set, it produces a much more
+# clear error due to the default value change in go1.20.
+export CGO_ENABLED=1
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -1,0 +1,14 @@
+#!/usr/bin/env bash
+# Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+# See the file LICENSE for licensing terms.
+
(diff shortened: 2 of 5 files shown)


scripts/test.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+# Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+# See the file LICENSE for licensing terms.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# Directory above this script
+RELAYER_PATH=$( cd "$( dirname "${BASH_SOURCE[0]}" )"; cd .. && pwd )
```
minghinmatthewlam, line 10: nit: can you make this into two lines like above

scripts/constants.sh
```diff
@@ -22,3 +22,12 @@ relayer_dockerhub_repo=${DOCKER_REPO:-"awm-relayer"}
 current_branch=$(git symbolic-ref -q --short HEAD || git describe --tags --exact-match || true)
 
 git_commit=${RELAYER_COMMIT:-$( git rev-list -1 HEAD )}
+
+# Set the CGO flags to use the portable version of BLST
+#
+# We use "export" here instead of just setting a bash variable because we need
+# to pass this flag to all child processes spawned by the shell.
+export CGO_CFLAGS="-O -D__BLST_PORTABLE__"
```
minghinmatthewlam, line 30: do we need this for teleporter as well?

minghinmatthewlam (approved):
LGTM, just some questions

gwen917 (dismissed):
LGTM

scripts/constants.sh
michaelkaplan13: The teleporter repo itself doesn't have any Go applications that use BLST. Those should just be avalanchego, subnet-evm, and the relayer.

scripts/test.sh
michaelkaplan13: Lets also do the same in `constants.sh`

michaelkaplan13 (commented):
LGTM, just the one nit from @minghinmatthewlam would be good to keep uniform.

scripts/test.sh
cam-schultz: Fixed

michaelkaplan13 (approved):

PR #26, opened 2023-09-01
E2E tests

### LINKED ISSUE #15: Set up end-to-end testing framework

minghinmatthewlam:
## Why this should be merged
Provides an e2e test that makes sure the relayer is running properly from start up, subscribing, and delivering the message to destination chain and address. Fixes #15 

## How this works
- Runs as part of CI
- Builds a local network with warp enabled subnets
- Configures a test relayer config
- Builds and runs the relayer separately 
- Simulates a transaction from the Teleporter contract to `sendWarpMessage` with a Teleporter message payload
- Verifies message gets delivered properly to destination with right values.

This can be extended for other future message protocols we support. 

## How this was tested
CI, unit tests, locally

## How is this documented
README


.github/workflows/test.yml
cam-schultz, line 1: It looks like the tab spacing in this file was changed. Can you change it back to two spaces to remain consistent with the other workflow files in this repo?

.github/workflows/test.yml
```diff
@@ -1,36 +1,64 @@
...
+            - name: Checkout subnet-evm repository
+              uses: actions/checkout@v3
+              with:
+                  repository: ava-labs/subnet-evm
+                  ref: v0.5.4
+
+            - name: Setup Go
+              uses: actions/setup-go@v4
+              with:
+                  go-version: "1.20.7"
```
cam-schultz, line 52: We can set the go version as a global environment, rather than repeating it in the two jobs.

README.md
```diff
@@ -37,3 +37,31 @@ The relayer consists of the following components:
...
+## Testing
+
+---
+
+### Unit tests
+
+Unit tests can be ran locally by running the command in root of the project:
+
+```
+go test ./...
```
cam-schultz, line 50: Let's change this to call `scripts/test.sh` rather than `go test` directly.

geoff-vball (commented):
Left some small comments after looking over everything but the main testing file.

README.md
```diff
@@ -37,3 +37,31 @@ The relayer consists of the following components:
...
+```
+
+### E2E tests
+
+E2E tests are ran as part of CI, but can also be ran locally. To run the E2E tests locally, you need to have the `avalanchego` build path and vm binary set up. For example with [subnet-evm](https://github.com/ava-labs/subnet-evm):
+
+```
+cd <SUBNET_EVM_PATH>
```
cam-schultz, line 58: nit: I think we can just put `cd subnet-evm` since these are example commands.

messages/teleporter/message.go
```diff
@@ -35,7 +35,7 @@ type ReceiveCrossChainMessageInput struct {
 }
 
 // unpack Teleporter message bytes according to EVM ABI encoding rules
-func unpackTeleporterMessage(messageBytes []byte) (*TeleporterMessage, error) {
+func UnpackTeleporterMessage(messageBytes []byte) (*TeleporterMessage, error) {
```
cam-schultz, line 38: Does this need to be exported? I'm not seeing any references to it outside `package teleporter`.

messages/teleporter/utils.go
```diff
@@ -47,3 +48,10 @@ func CalculateReceiveMessageGasLimit(numSigners int, executionRequiredGasLimit *
 
 	return res, nil
 }
+
+// Pack the SendCrossChainMessage event type. PackEvent is documented as not supporting struct types, so this should be used
+// with caution. Here, we only use it for testing purposes. In a real setting, the Teleporter contract should pack the event.
+func PackTeleporterMessage(destinationChainID common.Hash, message TeleporterMessage) ([]byte, error) {
```
cam-schultz, line 54: Similarly, does this need to be exported?

scripts/e2e_test.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+# Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+# See the file LICENSE for licensing terms.
+
+set -e
+
+AWM_RELAYER_PATH=$(
```
cam-schultz, line 7: Let's name this `RELAYER_PATH` to be consistent with the other scripts (those changes may not have all been merged)

tests/e2e/e2e_test.go
```diff
@@ -0,0 +1,441 @@
...
+
+const fundedKeyStr = "56289e99c94b6912bfc12adc093c9b51124f0dc54ac7a766b2bc5ccf558d8027"
+const teleporterKeyStr = "d26c3344074df322e7c66ad0d2357d96f0d07d0f40038264d1d64d276709da41"
+
+var (
+	anrConfig                 = runner.NewDefaultANRConfig()
+	manager                   = runner.NewNetworkManager(anrConfig)
+	warpChainConfigPath       string
+	relayerConfigPath         string
+	teleporterContractAddress = common.HexToAddress("1dD31B5351e76d51F4B152ce64fE5cf594694De5")
```
cam-schultz, line 50: nit: can you add leading 0x to this address string?

tests/e2e/e2e_test.go
```diff
@@ -0,0 +1,441 @@
...
+	ginkgo.RunSpecs(t, "Relayer e2e test")
+}
+
+func toWebsocketURI(uri string, blockchainID string) string {
+	return fmt.Sprintf("ws://%s/ext/bc/%s/ws", strings.TrimPrefix(uri, "http://"), blockchainID)
+}
+
+func toRPCURI(uri string, blockchainID string) string {
+	return fmt.Sprintf("http://%s/ext/bc/%s/rpc", strings.TrimPrefix(uri, "http://"), blockchainID)
+}
```
cam-schultz, line 77: Can you move this to a test utils file?

messages/teleporter/message.go
minghinmatthewlam: this is used in `e2e_test.go`

messages/teleporter/utils.go
minghinmatthewlam: this is used in `e2e_test.go`

scripts/e2e_test.sh
```diff
@@ -0,0 +1,28 @@
...
+
+source "$AWM_RELAYER_PATH"/scripts/constants.sh
+
+source "$AWM_RELAYER_PATH"/scripts/versions.sh
+
+# Build ginkgo
+# to install the ginkgo binary (required for test build and run)
+go install -v github.com/onsi/ginkgo/v2/ginkgo@${GINKGO_VERSION}
+
+ACK_GINKGO_RC=true ginkgo build ./tests/e2e
```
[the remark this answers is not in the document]
minghinmatthewlam: removed, it was to silence a RC advertisement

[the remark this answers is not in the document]
minghinmatthewlam: advertisement for ginkgo 2.0 but we're already on v2.2

vms/evm/destination_client.go
```diff
@@ -134,32 +134,24 @@ func (tdc *destinationClient) SendTx(signedMessage *avalancheWarp.Message,
...
-		Value:     big.NewInt(0),
-		Data:      callData,
-		AccessList: types.AccessList{
-			{
-				Address:     warp.ContractAddress,
-				StorageKeys: predicateutils.BytesToHashSlice(predicateBytes),
-			},
-		},
-	})
+	tx := predicateutils.NewPredicateTx(
```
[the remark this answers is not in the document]
minghinmatthewlam: the `predicateutils.NewPredicateTx` just helps form a predicate tx that packs the predicate bytes, and forms the access list with the predicate address. Will perform the same as we previously had, but I think it's good to use the `predicateutils` in case of any changes on subnet-evm side.

.github/workflows/test.yml
minghinmatthewlam: yup had a wrong config in my `setting.json` that set all tab spacing to be 4 spaces, removed that setting. Still have yaml default formatting on for better indentation.

geoff-vball (commented):
This is looking pretty good! This is going to be really helpful for testing, thanks for putting it together.

tests/e2e/e2e_test.go
```diff
@@ -0,0 +1,441 @@
...
+		RequiredGasLimit:        big.NewInt(1),
+		AllowedRelayerAddresses: []common.Address{},
+		Receipts:                []teleporter.TeleporterMessageReceipt{},
+		Message:                 []byte{1, 2, 3, 4},
+	}
+)
+
+func TestE2E(t *testing.T) {
+	if os.Getenv("RUN_E2E") == "" {
+		t.Skip("Environment variable RUN_E2E not set; skipping E2E tests")
```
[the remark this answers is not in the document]
geoff-vball: There's no case for running this file without the flag, it's just that the current setup in `test.sh` will run all test files in the repo. I'm not sure if there is an easy way to exclude a directory from `go test ./...`, but if you have one that might be easier. We just want a way to skip E2E tests and only run unit tests.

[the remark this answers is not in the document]
minghinmatthewlam: yup exactly that's added right now because `go test ./...` will run the e2e test as well

tests/e2e_test.go
```diff
@@ -0,0 +1,422 @@
...
+
+		gomega.Expect(err).Should(gomega.BeNil())
+
+		time.Sleep(15 * time.Second)
+		log.Info("Subscribing to new heads")
+
+		newHeadsB := make(chan *types.Header, 10)
+		sub, err := chainBWSClient.SubscribeNewHead(ctx, newHeadsB)
+		gomega.Expect(err).Should(gomega.BeNil())
+		defer sub.Unsubscribe()
```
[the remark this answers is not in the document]
minghinmatthewlam: for the test network no new blocks should be created unless the test sends a tx, but will move it to above `SendTransaction` regardless.

tests/e2e_test.go
```diff
@@ -0,0 +1,422 @@
...
+	predicateutils "github.com/ava-labs/subnet-evm/utils/predicate"
+	warpPayload "github.com/ava-labs/subnet-evm/warp/payload"
+	"github.com/ava-labs/subnet-evm/x/warp"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+)
+
```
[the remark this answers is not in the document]
minghinmatthewlam: I think the comment section above the `ginkgo.Describe` covers this comment.

tests/e2e_test.go
```diff
@@ -0,0 +1,422 @@
...
+		predicateBytes, err := predicateutils.UnpackPredicate(packedPredicate)
+		gomega.Expect(err).Should(gomega.BeNil())
+		receivedWarpMessage, err = avalancheWarp.ParseMessage(predicateBytes)
+		gomega.Expect(err).Should(gomega.BeNil())
+
+		// Check that the transaction is no longer pending
+		txHash := block.Transactions()[0].Hash()
+		_, isPending, err := chainBWSClient.TransactionByHash(ctx, txHash)
+		gomega.Expect(err).Should(gomega.BeNil())
+		gomega.Expect(isPending).Should(gomega.BeFalse())
```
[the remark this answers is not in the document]
minghinmatthewlam: agree that this check was extra, and more of a `subnet-evm` test. Removed as it doesn't really test relayer/teleporter. The transaction `signedTx` is different from the `txHash` we receive here, the first one is submitted on source chain which gets picked up by the relayer to submit a transaciton on the destination chain.

geoff-vball:
How do you feel about adding `gomega` as a dot-import? I think the current assertions are very difficult to read, and removing the need for `gomega.` prefixes everywhere would help a lot.

README.md
```diff
@@ -37,3 +37,31 @@ The relayer consists of the following components:
...
+```
+
+### E2E tests
+
+E2E tests are ran as part of CI, but can also be ran locally. To run the E2E tests locally, you need to have the `avalanchego` build path and vm binary set up. For example with [subnet-evm](https://github.com/ava-labs/subnet-evm):
+
+```bash
+cd subnet-evm
+BASEDIR=/tmp/e2e-test AVALANCHEGO_BUILD_PATH=/tmp/e2e-test/avalanchego ./scripts/install_avalanchego_release.sh
+./scripts/build.sh /tmp/e2e-test/avalanchego/plugins/srEXiWaHuhNyGwPUi444Tu47ZEDwxTWrbQiuD7FmgSAQ6X7Dy
```
geoff-vball, line 60: To get the tests to work on my ubuntu box, I had to use a different directory other than `/tmp` to install avalanchego. I don't think my system will let me run executables from there. I just used '~/tmp` instead for all these steps.

README.md
```diff
@@ -37,3 +37,31 @@ The relayer consists of the following components:
...
+
+### Unit tests
+
+Unit tests can be ran locally by running the command in root of the project:
+
+```bash
+./scripts/test.sh
+```
+
+### E2E tests
```
cam-schultz, line 53: Can you add a note here to install Gingko following the instructions [here](https://onsi.github.io/ginkgo/#installing-ginkgo)?

README.md
```diff
@@ -37,3 +37,31 @@ The relayer consists of the following components:
...
+
+E2E tests are ran as part of CI, but can also be ran locally. To run the E2E tests locally, you need to have the `avalanchego` build path and vm binary set up. For example with [subnet-evm](https://github.com/ava-labs/subnet-evm):
```
cam-schultz, line 55: Can you note that the you need clone subnet-evm locatlly and check out the correct version? Maybe we could leverage `versions.sh` to automate this.

tests/e2e_test.go
```diff
@@ -0,0 +1,419 @@
...
+	})
+
+	ginkgo.It("Set up relayer config", ginkgo.Label("Relayer", "Setup Relayer"), func() {
+		hostA, portA, err := getURIHostAndPort(chainANodeURIs[0])
+		gomega.Expect(err).Should(gomega.BeNil())
+
+		hostB, portB, err := getURIHostAndPort(chainBNodeURIs[0])
+		gomega.Expect(err).Should(gomega.BeNil())
+
+		log.Info("Setting up relayer config", "hostA", hostA, "portA", portA, "blockChainA", blockchainIDA.String(), "hostB", hostB, "portB", portB, "blockChainB", blockchainIDB.String(), "subnetA", subnetA.String(), "subnetB", subnetB.String())
```
cam-schultz, line 235: Can we break this into multiple lines for readability?

tests/e2e_test.go
```diff
@@ -0,0 +1,419 @@
...
+
+		relayerConfig := config.Config{
```
cam-schultz, line 237: It might be useful for authoring additional tests to have a global `defaultRelayerConfig` and modify it for individual test cases.

tests/e2e_test.go
```diff
@@ -0,0 +1,419 @@
...
+
+		packedInput, err := warp.PackSendWarpMessage(warp.SendWarpMessageInput{
+			DestinationChainID: common.Hash(blockchainIDB),
+			DestinationAddress: teleporterContractAddress,
+			Payload:            payload,
+		})
+		gomega.Expect(err).Should(gomega.BeNil())
+
+		// Send a transaction that simulates the Teleporter contract calling sendWarpMessage with a Teleporter message payload
+		tx := types.NewTx(&types.DynamicFeeTx{
```
cam-schultz, line 342: Similar to the above comment about relayer config, it might be nice to have a default teleporter message at global scope. Ideally relayer e2e tests shouldn't depend on teleporter message content, as that would be better suited for the teleporter integration tests. There are probably exceptions of course.

tests/e2e_test.go
```diff
@@ -0,0 +1,419 @@
...
+		newHeadsB := make(chan *types.Header, 10)
+		sub, err := chainBWSClient.SubscribeNewHead(ctx, newHeadsB)
+		gomega.Expect(err).Should(gomega.BeNil())
+		defer sub.Unsubscribe()
+
+		log.Info("Sending sendWarpMessage transaction", "destinationChainID", blockchainIDB, "txHash", signedTx.Hash())
+		err = chainAWSClient.SendTransaction(ctx, signedTx)
+		gomega.Expect(err).Should(gomega.BeNil())
+
+		log.Info("Waiting for new block confirmation")
```
cam-schultz, line 369: Can you add a comment above this block? Something like "Get the latest block from subnet B"

tests/e2e_test.go
```diff
@@ -0,0 +1,419 @@
...
+		gomega.Expect(receipt.Status).Should(gomega.Equal(types.ReceiptStatusSuccessful))
+
+		log.Info("Finished sending warp message, closing down output channel")
+
+		// Cancel the command and stop the relayer
+		relayerCancel()
+		_ = cmd.Wait()
+	})
+
+	ginkgo.It("Verify Warp Message", ginkgo.Label("Relay", "VerifyWarp"), func() {
```
cam-schultz, line 403: nit: Can you rename this step? "Verify warp message" sounds like the aggregate signature verification within Avalanchego, which is not what we're checking here.

cam-schultz (commented):
A handful of small comments in the test, but overall looks great!

tests/e2e_test.go
```diff
@@ -0,0 +1,610 @@
...
+		receivedWarpMessage, err = avalancheWarp.ParseMessage(predicateBytes)
+		gomega.Expect(err).Should(gomega.BeNil())
+
+		// Check that the transaction has successful receipt status
+		txHash := block.Transactions()[0].Hash()
+		receipt, err = chainBRPCClient.TransactionReceipt(ctx, txHash)
+		gomega.Expect(err).Should(gomega.BeNil())
+		gomega.Expect(receipt.Status).Should(gomega.Equal(types.ReceiptStatusSuccessful))
+
+		// // Try sending the same Teleporter message again. This should fail to be delivered
```
gwen917, line 498: do we want to remove the commented code?

tests/utils.go
```diff
@@ -0,0 +1,126 @@
...
+		Value:     defaultTeleporterMessageValue,
+		Data:      data,
+	})
+}
+
+func readHexTextFile(filename string) []byte {
+	fileData, err := os.ReadFile(filename)
+	gomega.Expect(err).Should(gomega.BeNil())
+	hexString := string(fileData)
+	hexString = hexString[2:] // remove 0x prefix
```
geoff-vball, line 122: `Utils.go` has a function `SanitizeHashString`. I think we should rename it `SanitizeHexString` and use it here.

tests/e2e_test.go
```diff
@@ -0,0 +1,610 @@
...
+	"github.com/ava-labs/subnet-evm/tests/utils/runner"
+	predicateutils "github.com/ava-labs/subnet-evm/utils/predicate"
+	warpPayload "github.com/ava-labs/subnet-evm/warp/payload"
+	"github.com/ava-labs/subnet-evm/x/warp"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
```
geoff-vball, line 40: How do people feel about using a dot-import for `gomega`? I find the assertions really hard to read with the package path in every statement at least twice.

cam-schultz: I'm fine dot importing `gomega` in this context. I'd want to avoid them almost anywhere else in the repo though, especially in the application code.

tests/utils.go
cam-schultz: Changed 👍

tests/e2e_test.go
geoff-vball: Yes for sure :)

geoff-vball (dismissed):

tests/e2e_test.go
```diff
@@ -0,0 +1,555 @@
...
+		log.Info("Finished setting up e2e test subnet variables")
+	})
+
+	ginkgo.It("Deploy Teleporter Contract", ginkgo.Label("Relayer", "Deploy Teleporter"), func() {
+		ctx := context.Background()
+
+		// Read in the Teleporter contract information
+		teleporterContractAddress = common.BytesToAddress(readHexTextFile("./tests/UniversalTeleporterMessengerContractAddress.txt"))
+		teleporterDeployerAddress := common.BytesToAddress(readHexTextFile("./tests/UniversalTeleporterDeployerAddress.txt"))
+		teleporterDeployerTransaction := readHexTextFile("./tests/UniversalTeleporterDeployerTransaction.txt")
```
michaelkaplan13, line 242: It looks like we read the hex string, convert it to a `[]byte`, but then the only place we use it we first convert it back to a hex string. Could we just read the string from the file and use it directly?

cam-schultz: We could, but if we wanted to still use `readHexTextFile`, we'd have to alter it to return `string`, then convert that `string` to `common.Address` for the other two vars here. IMO, the current implementation is more readable in the test file, even though there are unnecessary conversions as you pointed out.

geoff-vball: You can use `common.HexToAddress` to get the same readability if you want to change `readHexTextFile` to return a string.

tests/e2e_test.go
```diff
@@ -0,0 +1,555 @@
...
+		relayerCmd                       *exec.Cmd
+		relayerCancel                    context.CancelFunc
+	)
+
+	fundedKey, err = crypto.HexToECDSA(fundedKeyStr)
+	if err != nil {
+		panic(err)
+	}
+
+	ginkgo.It("Setup subnet URIs", ginkgo.Label("Relayer", "Setup"), func() {
```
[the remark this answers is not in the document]
cam-schultz: My understanding is that we'd want all common setup code (standing up the subnets, deploying Teleporter, etc.) to be in the `BeforeSuite` call. Individual tests can then live in their own `Describe` specs (AKA a suite), which can further be broken down into `It` specs. 

IIUC in an ordered suite `BeforeAll` is no different than the first `It` spec within the `Describe` suite.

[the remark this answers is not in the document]
michaelkaplan13: Cool, makes sense to me. That is my understand based on that documentation as well.

As is, just wanted to call out that `Setup subnet URIs` and `Deploy Teleporter Contract` are their own `It` specs rather than in the `BeforeSuite` implementation. I think that is fine for now though, as not all AWM relayer tests will necessarily depend on Teleporter moving forward.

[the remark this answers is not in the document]
cam-schultz: Updated to move those stages to `BeforeSuite` 👍

tests/e2e_test.go
minghinmatthewlam: thoughts on dot importing `ginkgo` as well? That seems to be the standard from ginkgo bootstrapping https://onsi.github.io/ginkgo/#bootstrapping-a-suite. No strong pref though

geoff-vball: I'd slightly lean against it. Doing it for `gomega` is great because it's very repetitive and hurts readability. In general I like being explicit about which package something is from, and I don't think `ginkgo` is getting in the way.

minghinmatthewlam: makes sense, sgtm

minghinmatthewlam:
LGTM, going to refresh reviews

messages/teleporter/message.go
```diff
@@ -65,16 +67,26 @@ func UnpackTeleporterMessage(messageBytes []byte) (*TeleporterMessage, error) {
...
 
+// UnpackMessageReceivedResult attempts to unpack result bytes to a bool indicating whether the message was received
 func UnpackMessageReceivedResult(result []byte) (bool, error) {
 	var success bool
 	err := EVMTeleporterContractABI.UnpackIntoInterface(&success, "messageReceived", result)
 	return success, err
 }
+
+// PackSendCrossChainMessageEvent packs the SendCrossChainMessage event type. PackEvent is documented as not supporting struct types, so this should be used
+// with caution. Here, we only use it for testing purposes. In a real setting, the Teleporter contract should pack the event.
```
geoff-vball, line 88: Can we make this warning a little more obvious? Maybes start one line of the comment with "CAUTION". I would totally miss this most of the time.

cam-schultz: Good call, clarified that. I also thought about moving it to `package tests`, but that unfortunately creates an import cycle. If we decide to move these Teleporter packing/unpacking utils to `teleporter`, we can spend more energy putting up better guardrails, but IMO a more explicit comment is sufficient for now.

geoff-vball (approved):

cam-schultz (dismissed):

geoff-vball (approved):

cam-schultz (approved):

PR #28, opened 2023-09-06
Add a Go linter to CI

gwen917:
## Why this should be merged
- Add a linter for go files with set lint rules and a github workflow.

## How this works
- Use Github Action for running [golangci-lint](https://github.com/golangci/golangci-lint) 

## How this was tested
- CI

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -9,7 +9,7 @@ on:
       - '*'
 
 jobs:
-  golang-linter:
+  golangci:
     runs-on: ubuntu-20.04
 
     steps:
@@ -28,5 +28,5 @@ jobs:
       uses: golangci/golangci-lint-action@v3
       with:
         version: v1.51
-        working-directory: .
+        working-directory: ./awm-relayer
         args: --timeout 10m
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -9,18 +9,17 @@ on:
       - '*'
 
 jobs:
-  lint_test:
-    name: Lint
+  golang-linter:
     runs-on: ubuntu-20.04
 
     steps:
     - name: Checkout repository
-    uses: actions/checkout@v3
-    with:
-      path: awm-relayer
+      uses: actions/checkout@v3
+      with:
+        path: awm-relayer
 
     - name: Setup Go
-    - uses: actions/setup-go@v4
+      uses: actions/setup-go@v4
       with:
         go-version: '1.20.7'
         check-latest: true
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -1,0 +1,33 @@
+# Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+# See the file LICENSE for licensing terms.
+
+name: Run Golang Linter
+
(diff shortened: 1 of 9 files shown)


vms/evm/contract_message_test.go
```diff
@@ -7,44 +7,12 @@ import (
...
-	"github.com/ava-labs/avalanchego/vms/platformvm/warp"
 	"github.com/ava-labs/awm-relayer/config"
-	warpPayload "github.com/ava-labs/subnet-evm/warp/payload"
-	"github.com/ethereum/go-ethereum/common"
 	"github.com/stretchr/testify/assert"
 	"go.uber.org/mock/gomock"
 )
 
-// Used to create a valid unsigned message for testing. Should not be used directly in tests.
-func createUnsignedMessage() *warp.UnsignedMessage {
```
cam-schultz, line 21: Let's keep function. For context, it's used to generate a valid unsigned warp message, which can be hardcoded as the expected value in `TestUnpack`. If the warp message format changes, then `TestUnpack` should fail, signalling that we should make corresponding changes in `awm-relayer`. We can remove this function once the warp message format is stable.

gwen917: sure. since it causes the unused function linter error, I will try to update the config so that it will not error on this.

geoff-vball: I think you can use `//nolint:unused` above the function signature and it should let the linter pass

scripts/lint.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+# Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+# See the file LICENSE for licensing terms.
+
+set -o errexit
```
minghinmatthewlam, line 5: can you add a comment about what these do

minghinmatthewlam (commented):
Mostly LGTM matches what I see in other company lint configs, nit for comment and adding back `createUnsignedMessage`

.github/workflows/linter.yml
```diff
@@ -0,0 +1,32 @@
+# Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+# See the file LICENSE for licensing terms.
+
+name: Run Golang Linter
+
+on:
+  push:
+    branches:
+      - '*'
```
geoff-vball, line 9: I think we should run this in the same way as most of our other checks - on all PRs and also pushes to main.

gwen917: oh, does this not check all PRs and push to main? The [test.yml](https://github.com/ava-labs/awm-relayer/blob/main/.github/workflows/test.yml#L6-L9) did the same.

vms/evm/contract_message_test.go
gwen917: fixed.

cam-schultz (approved):
LGTM!

bernard-avalabs (approved):
LGTM

.github/workflows/linter.yml
gwen917: Discussed in the standup yesterday. The lint is light weighted. It should should be fine running for every push.

PR #42, opened 2023-09-20
Integrate warp non compatible changes

### LINKED ISSUE #24: Update relayer to support warp changes
**Context and scope**
Warp is having non compatible changes, and we need to make corresponding changes to the relayer, including changes for handling Teleporter changes.

Pending warp pr https://github.com/ava-labs/subnet-evm/pull/823
Teleporter issue: https://github.com/ava-labs/teleporter/issues/19

We would want a new release version for deployment

minghinmatthewlam:
## Why this should be merged
Warp has non compatible changes recently added in `v0.5.6` that changes how Teleporter integrates with the precompile. We need to update relayer with any corresponding changes and update subnet-evm version. Fixes #24 

## How this works
Updates subnet-evm version dependency, uses `predicateutils` for predicate tx

## How this was tested
Building local docker image, and running e2e test with corresponding teleporter changes https://github.com/ava-labs/teleporter/pull/29


geoff-vball (dismissed):
👍

michaelkaplan13 (commented):
All these changes LGTM, but do we need to encode the signed message as an array now that the warp predicate can contain multiple messages? 

I would have also expected there to be a `messageIndex` parameter added somewhere since it's now a parameter to the `getVerifiedWarpMessage` precompile call. I'm not sure if we're planning on allowing that to be a parameter in `receiveCrossChainMessage` in Teleporter or not, though seems like it would likely make sense to.

minghinmatthewlam:
> All these changes LGTM, but do we need to encode the signed message as an array now that the warp predicate can contain multiple messages?
> 
> I would have also expected there to be a `messageIndex` parameter added somewhere since it's now a parameter to the `getVerifiedWarpMessage` precompile call. I'm not sure if we're planning on allowing that to be a parameter in `receiveCrossChainMessage` in Teleporter or not, though seems like it would likely make sense to.

The slices are handled by the `AccessList` where each access tuple would correspond to a predicate's bytes, and `predicateutils.NewPredicateTx` will help add the message as the first index.

Agreed `receiveCrossChainMessage` can accept an index, I had hardcoded it to 0, but ill add that in case dapps or future relayer improvements leverages that. Putting back in draft until I address that.

michaelkaplan13:
> > All these changes LGTM, but do we need to encode the signed message as an array now that the warp predicate can contain multiple messages?
> > I would have also expected there to be a `messageIndex` parameter added somewhere since it's now a parameter to the `getVerifiedWarpMessage` precompile call. I'm not sure if we're planning on allowing that to be a parameter in `receiveCrossChainMessage` in Teleporter or not, though seems like it would likely make sense to.
> 
> The slices are handled by the `AccessList` where each access tuple would correspond to a predicate's bytes, and `predicateutils.NewPredicateTx` will help add the message as the first index.
> 
> Agreed `receiveCrossChainMessage` can accept an index, I had hardcoded it to 0, but ill add that in case dapps or future relayer improvements leverages that. Putting back in draft until I address that.

Oh nice! I didn't realize that about how the predicates are encoded within the `AccessList`. I like that much better than an array. 

And sounds good about the index. 🙏

messages/teleporter/message.go
```diff
@@ -33,6 +33,7 @@ type TeleporterMessageReceipt struct {
 // in the contract deployed on the destination chain
```
bernard-avalabs, line 33: Can the comment be expanded to describe the new Index field?

bernard-avalabs (approved):
LGTM.

messages/teleporter/utils.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 const (
-	receiveMessageGasLimitBufferAmount = 100_000
+	receiveMessageGasLimitBufferAmount = 100_000_0
```
michaelkaplan13, line 15: Was there a reason this needed to be increased? 

1,000,000 extra gas is a lot, so ideally we don't have to increase it this much. Either way, lets format the number as `1_000_000` to make it easier to read.

messages/teleporter/message.go
```diff
@@ -35,6 +35,7 @@ type TeleporterMessageReceipt struct {
...
 type ReceiveCrossChainMessageInput struct {
 	RelayerRewardAddress common.Address `json:"relayerRewardAddress"`
+	Index                uint32         `json:"index"`
```
michaelkaplan13, line 38: Nit: This is dependent on the corresponding Teleporter PR, but I think it would make more sense to put the `index` as the first parameter of `receiveCrossChainMessage` since it specifies which message is being received.

messages/teleporter/utils.go
minghinmatthewlam: I was taking a guess as to why the e2e_test failed, and if it's running out of gas. Removed.

cam-schultz (approved):
LGTM. Agree with Michael's comment to have `Index` as the first parameter to `receiveCrossChainMessage`

messages/teleporter/message.go
minghinmatthewlam: updated to `messageIndex` to match teleporter PR, and made first parmeter

michaelkaplan13 (approved):
LGTM

cam-schultz (approved):

geoff-vball (approved):
👍

cam-schultz (approved):

PR #50, opened 2023-09-26
use docker build path context

minghinmatthewlam:
## Why this should be merged
Release action for publish docker image is failing, this hopes to address and fix.

## How this works
for `build-push-action` [docs](https://github.com/docker/build-push-action/tree/v5/#git-context), specifies we use git context without the `checkout` action, or if we use `checkout` action we should pass in path `context`. So this PR is adding the path context since we do use `checkout` action.

## How this was tested
n/a

## How is this documented
n/a

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -21,7 +21,7 @@ jobs:
       # Needed for multi-platform builds
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
-      
+
       - name: Log in to Docker Hub
         uses: docker/login-action@v3
         with:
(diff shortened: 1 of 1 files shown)


cam-schultz (approved):

geoff-vball (approved):

michaelkaplan13:
Given that we don't modify any files after the checkout step, it seems like the recommended use here would be to not have the separate checkout step at all, and just use the default git context. I could be missing some context though.

minghinmatthewlam:
in the `build-push-actions` docs https://github.com/docker/build-push-action, by default it doesn't use the `checkout` action, and wouldn't need the additionial `context: . `, but think this is easier to read and more inline with the other workflows that use the `checkout` option. Discussed offline, will leave in this format.

PR #54, opened 2023-10-05
Block hash publisher

cam-schultz:
## Why this should be merged
Adds relayer support for publishing verified block hashes from subnets. See https://github.com/ava-labs/subnet-evm/issues/734 for more details on the `subnet-evm` feature and use cases. Also includes architecture and interface changes to better support one-to-many Warp message relaying for future anycast protocols.

## How this works
Adds `EVM_BLOCKHASH` VM type, and `BLOCK_HASH_PUBLISHER` message protocol. `EVM_BLOCKHASH` subscribes to all new blocks from the configured source, and passes it to the message relayer. `BLOCK_HASH_PUBLISHER` then creates a single aggregate signature and publishes the block hash on each of the configured destinations via `BlockHashReceiver.sol`, added here: https://github.com/ava-labs/teleporter/pull/51. 

Some notes:
- Processing of historical blocks is NOT included in this PR to limit this PRs scope.
- Publishing block hashes from the primary network is also NOT included in this PR. In that case, we need to create a separate aggregate signature for each destination, which I also elected to skip to limit this PRs scope.

## How this was tested
- [x] E2E test
- [x] Expand E2E to two destination chains
- [x] block_hash_publisher/config tests
- [x] block_hash_publisher/message_manager tests
- [ ] evm_block_hash/contract_message tests

## How is this documented
Code comments


config/config.go
```diff
@@ -232,6 +237,8 @@ func (s *SourceSubnet) Validate() error {
 				return fmt.Errorf("invalid message contract address in EVM source subnet: %s", messageContractAddress)
 			}
 		}
+	case EVM_BLOCKHASH:
```
bernard-avalabs, line 240: It seems odd that Blockhash is considered its own VM type. Are we overloading the term "VM" here?

messages/block_hash_publisher/message_manager.go
```diff
@@ -0,0 +1,170 @@
...
+	blockInterval       uint64
+
+	lastTimeSent uint64
+	lastBlock    uint64
+}
+
+func (d *destinationSenderInfo) shouldSend(blockTimestamp uint64, blockNumber uint64) bool {
+	if d.useTimeInterval {
+		interval := d.timeIntervalSeconds
+		if time.Unix(int64(blockTimestamp), 0).Sub(time.Unix(int64(d.lastTimeSent), 0)) < (time.Duration(interval) * time.Second) {
```
geoff-vball, line 37: Do we need to convert both of these to unix times?

messages/block_hash_publisher/message_manager.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/ava-labs/awm-relayer/config"
+	"github.com/ava-labs/awm-relayer/vms"
+	"github.com/ava-labs/awm-relayer/vms/vmtypes"
+	teleporter_block_hash "github.com/ava-labs/teleporter/abi-bindings/Teleporter/TeleporterBlockHashReceiver"
+	"github.com/ethereum/go-ethereum/common"
+	"go.uber.org/zap"
+)
+
+const (
+	publishBlockHashGasLimit = 275000
```
bernard-avalabs, line 19: Is this temporary, or do we know if this will be enough?

tests/publish_block_hashes.go
```diff
@@ -0,0 +1,336 @@
...
+	return blockHashReceiverAddress
+}
+
+func receiveBlockHash(
+	ctx context.Context,
+	blockHashReceiverAddress common.Address,
+	newHeads chan *types.Header,
+	subnetInfo teleporterTestUtils.SubnetTestInfo,
+	blockHashABI *abi.ABI, expectedHashes []common.Hash) {
+	newHead := <-newHeads
```
geoff-vball, line 98: nit: Can we put a newline here? this is hard to read.

messages/block_hash_publisher/message_manager.go
```diff
@@ -0,0 +1,170 @@
...
+	destinationClients map[ids.ID]vms.DestinationClient,
+) (*messageManager, error) {
+	// Marshal the map and unmarshal into the Block Hash Publisher config
+	data, err := json.Marshal(messageProtocolConfig.Settings)
+	if err != nil {
+		logger.Error("Failed to marshal Block Hash Publisher config")
+		return nil, err
+	}
+	var messageConfig Config
+	if err := json.Unmarshal(data, &messageConfig); err != nil {
```
bernard-avalabs, line 67: I'm not sure I understand why we are marshalling and then immediately unmarshalling the data here.

messages/block_hash_publisher/message_manager.go
```diff
@@ -0,0 +1,170 @@
...
+				"Failed to decode base-58 encoded destination chain ID",
+				zap.Error(err),
+			)
+			return nil, err
+		}
+		destinations[destinationID] = &destinationSenderInfo{
+			useTimeInterval:     destination.useTimeInterval,
+			timeIntervalSeconds: destination.timeIntervalSeconds,
+			blockInterval:       destination.blockInterval,
+			address:             common.HexToAddress(destination.Address),
```
bernard-avalabs, line 94: Can you confirm that destination.Address has already been verified so HexToAddress should never fail?

bernard-avalabs (commented):
Generally makes sense to me. I've left a few comments.

config/types.go
```diff
@@ -9,12 +9,15 @@ type VM int
 const (
 	UNKNOWN_VM VM = iota
 	EVM
+	EVM_BLOCKHASH
```
geoff-vball, line 12: I might be misunderstanding something, but this doesn't feel like it should be a distinct VM, I think it would make more sense as a flag? I also don't really understand how we're using it right now - Does an `EVM_BLOCKHASH` subscriber perform a superset of the actions performed byan `EVM` subscriber?

cam-schultz:
Closing as stale. May revisit in the future.

PR #59, opened 2023-10-13
Configuration support fully specified source to destination mapping

gwen917:
## Why this should be merged
- Allow relayer only relays the message between specific subnets, instead of relaying all the messages between `source subnets` to `destination subnets`
- Add `AllowedDestinations` chainIDs in `source subnets` to specify the allowed destinations.

## How this was tested
- Unit tests, CI


config/config.go
```diff
@@ -190,24 +191,35 @@ func (c *Config) Validate() error {
 	return nil
 }
 
-// GetSourceIDs returns the Subnet and Chain IDs of all subnets configured as a source
-func (cfg *Config) GetSourceIDs() ([]ids.ID, []ids.ID, error) {
+// GetSourceIDs returns the Subnet, Chain IDs and Allowed Destination Chain IDs of all subnets configured as a source
+func (cfg *Config) GetSourceIDs() ([]ids.ID, []ids.ID, map[ids.ID]map[ids.ID]bool, error) {
```
cam-schultz, line 195: I know this is slightly out of scope for this feature, but could we have `GetSourceIDs` be a standalone function that returns the values from the global configuration singleton?

cam-schultz, line 195: I would also recommend leaving the signature of `GetSourceIDs` as is, and instead adding a separate function to return the mapping of source chains to allowed destinations.

config/config.go
```diff
@@ -244,6 +256,13 @@ func (s *SourceSubnet) Validate() error {
 		}
 	}
 
+	// Validate the allowed destinations
```
cam-schultz, line 259: We should also validate that the allowed destinations are all present in `cfg.DestinationSubnets`

config/config.go
cam-schultz: We can call this function directly from `teleporter/message_manager.go` rather than having to pass it from `main`.

config/config.go
```diff
@@ -190,24 +191,35 @@ func (c *Config) Validate() error {
 	return nil
 }
 
-// GetSourceIDs returns the Subnet and Chain IDs of all subnets configured as a source
-func (cfg *Config) GetSourceIDs() ([]ids.ID, []ids.ID, error) {
+// GetSourceIDs returns the Subnet, Chain IDs and Allowed Destination Chain IDs of all subnets configured as a source
+func (cfg *Config) GetSourceIDs() ([]ids.ID, []ids.ID, map[ids.ID]map[ids.ID]bool, error) {
 	var sourceSubnetIDs []ids.ID
 	var sourceChainIDs []ids.ID
+	var allowedDestinationChaiIDMap map[ids.ID]map[ids.ID]bool
```
cam-schultz, line 198:
```suggestion
	var allowedDestinationChainIDMap map[ids.ID]map[ids.ID]bool
```

config/config.go
```diff
@@ -190,24 +191,35 @@ func (c *Config) Validate() error {
...
 		sourceSubnetIDs = append(sourceSubnetIDs, subnetID)
 
 		chainID, err := ids.FromString(s.ChainID)
 		if err != nil {
-			return nil, nil, fmt.Errorf("invalid subnetID in configuration. error: %v", err)
+			return nil, nil, nil, fmt.Errorf("invalid subnetID in configuration. error: %v", err)
 		}
 		sourceChainIDs = append(sourceChainIDs, chainID)
+
+		allowedDestinationChaiIDs := make(map[ids.ID]bool)
```
cam-schultz, line 212:
```suggestion
		allowedDestinationChainIDs := make(map[ids.ID]bool)
```

messages/teleporter/message_manager.go
```diff
@@ -105,6 +108,19 @@ func (m *messageManager) ShouldSendMessage(warpMessageInfo *vmtypes.WarpMessageI
 	if !ok {
 		return false, fmt.Errorf("relayer not configured to deliver to destination. destinationChainID=%s", destinationChainID.String())
 	}
+
+	// If allowedDestinations is empty, then all destinations are allowed
+	// If allowedDestinations is not empty, then only the allowed destinations are allowed
+	if _, exist := m.allowedDestinations[destinationChainID]; !exist && len(m.allowedDestinations) > 0 {
+		m.logger.Info(
+			"Destination chain not allowed to receive messages.",
```
cam-schultz, line 116: I would rephrase this as "Relayer not configured to relay between source and destination", and add the `sourceChainID` to the log.

messages/teleporter/message_manager.go
```diff
@@ -105,6 +108,19 @@ func (m *messageManager) ShouldSendMessage(warpMessageInfo *vmtypes.WarpMessageI
 	if !ok {
 		return false, fmt.Errorf("relayer not configured to deliver to destination. destinationChainID=%s", destinationChainID.String())
 	}
+
+	// If allowedDestinations is empty, then all destinations are allowed
+	// If allowedDestinations is not empty, then only the allowed destinations are allowed
+	if _, exist := m.allowedDestinations[destinationChainID]; !exist && len(m.allowedDestinations) > 0 {
```
cam-schultz, line 114: The map returned by `Config.GetSourceIDs` is keyed by `sourceChainID`, not `destinationChainID`.

cam-schultz, line 114: Also, I think the condition may be wrong. We want to return early here if either 1) `allowedDestinations` is empty OR 2) `allowedDestinations` is non-empty, and the `destinationChainID` is present. I think we want something like:

```
if dsts, exist := m.allowedDestinations[sourceChainID]; exists {
    if allowed, ok := dsts[destinationChainID]; !ok || !allowed {
        m.logger.Info(...)
        return false, nil
    }
}
```

gwen917: I think the key should be `destinationChainID`. The map return from `Config.GetSouceIDs` is `map[id.ID]map[id.ID]bool`. The first key is sourceChainID, and the second key is destinationChainID. When call [runRelayer](https://github.com/ava-labs/awm-relayer/pull/59/files#diff-327181d0a8c5e6b164561d7910f4eeffd41442d55b2a2788fda2aa2692f17ec0R180), we pass in `map[sourceChainID]`. So here the key of the map is `destinationChainID`.

gwen917: Same as above. The `m.allowedDestinations` is map[id.ID]bool and the key is `destinationChainID`.

config/config.go
gwen917: checked [here](https://github.com/ava-labs/awm-relayer/pull/59/files#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R191-R201)

messages/teleporter/message_manager.go
cam-schultz: Ah, I did not realize that this is the nested map.

cam-schultz: Gotcha. I think we still want to check against the inner condition in my comment above. We need to check the length of the map first though.
```
if len(m.allowedDestinations) > 0 {
    if allowed, ok := m.allowedDestinations[destinationChainID]; !ok || !allowed {
        m.logger.Info(...)
        return false, nil
    }
}
```

config/config.go
```diff
@@ -37,15 +37,16 @@ type MessageProtocolConfig struct {
...
-	ChainID           string                           `mapstructure:"chain-id" json:"chain-id"`
-	VM                string                           `mapstructure:"vm" json:"vm"`
-	APINodeHost       string                           `mapstructure:"api-node-host" json:"api-node-host"`
-	APINodePort       uint32                           `mapstructure:"api-node-port" json:"api-node-port"`
-	EncryptConnection bool                             `mapstructure:"encrypt-connection" json:"encrypt-connection"`
-	RPCEndpoint       string                           `mapstructure:"rpc-endpoint" json:"rpc-endpoint"`
-	WSEndpoint        string                           `mapstructure:"ws-endpoint" json:"ws-endpoint"`
-	MessageContracts  map[string]MessageProtocolConfig `mapstructure:"message-contracts" json:"message-contracts"`
+	SubnetID            string                           `mapstructure:"subnet-id" json:"subnet-id"`
+	ChainID             string                           `mapstructure:"chain-id" json:"chain-id"`
```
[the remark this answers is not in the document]
cam-schultz: Yes, this is the Avalanche blockchain ID.

messages/teleporter/message_manager.go
```diff
@@ -105,6 +108,21 @@ func (m *messageManager) ShouldSendMessage(warpMessageInfo *vmtypes.WarpMessageI
...
+	// If allowedDestinations is not empty, then only the allowed destinations are allowed
+	if len(m.allowedDestinations) > 0 {
+		if allowed, exist := m.allowedDestinations[destinationChainID]; !exist || !allowed {
+			m.logger.Info(
+				"Relayer not configured to relay between source and destination",
+				zap.String("destinationChainID", destinationChainID.String()),
+				zap.String("warpMessageID", warpMessageInfo.WarpUnsignedMessage.ID().String()),
+				zap.String("teleporterMessageID", teleporterMessage.MessageID.String()),
+			)
+			return false, nil
```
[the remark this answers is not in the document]
cam-schultz: Yeah, that case should not be an error. With the upcoming subnet-evm changes (as well as anycasting support in the future), I think `ShouldSendMessage` is the right place to check against the list of configured destinations.

config/config.go
cam-schultz: Whoops, the change that adds a global config is still in a branch. I think we should register the config as a global singleton, rather than passing it through function calls.

messages/teleporter/message_manager.go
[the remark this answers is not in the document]
cam-schultz: Thinking about it some more, we can clean up this implementation by using the configured allowed destinations as a filter in on the list of destination clients in `NewRelayer` _before_ the call to `NewMessageManager`. We'd then pass in the filtered list of destination clients. In [relayer.go](https://github.com/ava-labs/awm-relayer/blob/199b3e772ccec3babda3b8cb168109c9afe36f9a/relayer/relayer.go#L78), we'd add something like:

```
var filteredDestinationClients map[ids.ID]ethclient.Client
allowedDestinationChainIDs, err := sourceSubnetInfo.GetAllowedDestinations()
if len(m.allowedDestinations) > 0 { 
    filteredDestinationClients := make(map[ids.ID]ethclient.Client)
    for _, id := range m.allowedDestinations {
        filteredDestinationClients[id] = destinationClients[id]
    }
} else {
    filteredDestinationClients = destinationClients
}
...
messageManager, err := messages.NewMessageManager(logger, addressHash, config, filteredDestinationClients, allowedDestinationChainIDs)
```

This would move logic for deciding which destinations are and are not valid to the relayer constructor which ones once, rather than the message manager `ShouldSendMessage` implementation, which runs for each message. When processing a message log, we could then reference the allowed destinations for that source subnet [here](https://github.com/ava-labs/awm-relayer/blob/199b3e772ccec3babda3b8cb168109c9afe36f9a/relayer/relayer.go#L188)

config/config.go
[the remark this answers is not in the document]
cam-schultz: This is an unfortunate overloading of terms, but I agree we should make the distinction where we can. FWIW, `subnet-evm` isn't consistent either: `ethclient.Client.ChainID` returns the blockchainID, while the Warp precompile has the unambiguous method `getBlockchainID`. I don't think the EVM chain ID comes into play in `subnet-evm` or `awm-relayer`.

There are 479 hits for `chainID` in this repo, however... Do you mind if we save this for a separate PR to not pollute the diff?

config/config.go
```diff
@@ -37,15 +37,16 @@ type MessageProtocolConfig struct {
...
+	VM                  string                           `mapstructure:"vm" json:"vm"`
+	APINodeHost         string                           `mapstructure:"api-node-host" json:"api-node-host"`
+	APINodePort         uint32                           `mapstructure:"api-node-port" json:"api-node-port"`
+	EncryptConnection   bool                             `mapstructure:"encrypt-connection" json:"encrypt-connection"`
+	RPCEndpoint         string                           `mapstructure:"rpc-endpoint" json:"rpc-endpoint"`
+	WSEndpoint          string                           `mapstructure:"ws-endpoint" json:"ws-endpoint"`
+	MessageContracts    map[string]MessageProtocolConfig `mapstructure:"message-contracts" json:"message-contracts"`
+	AllowedDestinations []string                         `mapstructure:"allowed-destinations" json:"allowed-destinations"`
```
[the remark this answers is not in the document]
cam-schultz: Renamed to `SupportedDestinations`

messages/teleporter/message_manager.go
```diff
@@ -105,6 +108,21 @@ func (m *messageManager) ShouldSendMessage(warpMessageInfo *vmtypes.WarpMessageI
 	if !ok {
 		return false, fmt.Errorf("relayer not configured to deliver to destination. destinationChainID=%s", destinationChainID.String())
 	}
+
+	// If allowedDestinations is empty, then all destinations are allowed
```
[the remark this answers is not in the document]
cam-schultz: Added config options documentation

messages/teleporter/message_manager.go
```diff
@@ -34,6 +34,7 @@ type messageManager struct {
 	// The cache is keyed by the Warp message ID, NOT the Teleporter message ID
 	teleporterMessageCache *cache.LRU[ids.ID, *TeleporterMessage]
 	destinationClients     map[ids.ID]vms.DestinationClient
+	supportedDestinationss map[ids.ID]bool
```
bernard-avalabs, line 37: Is the extra "s" at the end intentional?

messages/teleporter/message_manager.go
```diff
@@ -105,6 +108,22 @@ func (m *messageManager) ShouldSendMessage(warpMessageInfo *vmtypes.WarpMessageI
 	if !ok {
 		return false, fmt.Errorf("relayer not configured to deliver to destination. destinationChainID=%s", destinationChainID.String())
 	}
+
+	// If supportedDestinationss is empty, then all destinations are allowed
```
bernard-avalabs, line 112: This seems opposite than the expected behavior (one normally assumes that if the allow list is empty, then nothing is allowed).

bernard-avalabs (commented):
Generally makes sense to me. I left a few minor comments.

messages/teleporter/message_manager.go
cam-schultz: This is similar to how the "allowedRelayerAddresses" field is handled in `TeleporterMessenger`: https://github.com/ava-labs/teleporter/blob/main/contracts/src/Teleporter/TeleporterMessenger.sol#L126

The expected use case here is that if the `supportedDestinations` JSON key is not provided, we don't filter any of the configured destinations. This would be useful if the relayer operator wants messages from subnet A to be relayed to anybody, but messages from subnet B to only be relayed to subnet C, for example.

messages/teleporter/message_manager.go
cam-schultz: Typo 🙂 . Fixed now.

README.md
```diff
@@ -25,6 +25,85 @@ The relayer binary accepts a path to a JSON configuration file as the sole argum
 ./build/awm-relayer --config-file path-to-config
 ```
 
+### Configuration
```
minghinmatthewlam, line 28: Love this new README section 🙌

config/config.go
```diff
@@ -373,3 +394,27 @@ func (s *DestinationSubnet) GetRelayerAccountInfo() (*ecdsa.PrivateKey, common.A
 	pkBytes = append(pkBytes, pk.PublicKey.Y.Bytes()...)
 	return pk, common.BytesToAddress(crypto.Keccak256(pkBytes)), nil
 }
+
+//
+// Global config getters
+//
+
+// GetSourceIDs returns the Subnet and Chain IDs of all subnets configured as a source
+func GetSourceIDs() ([]ids.ID, []ids.ID, error) {
```
minghinmatthewlam, line 403: why did we move this to using a global config variable vs a receiver function?

config/config.go
```diff
@@ -37,15 +40,19 @@ type MessageProtocolConfig struct {
...
+	SubnetID              string                           `mapstructure:"subnet-id" json:"subnet-id"`
+	ChainID               string                           `mapstructure:"chain-id" json:"chain-id"`
+	VM                    string                           `mapstructure:"vm" json:"vm"`
+	APINodeHost           string                           `mapstructure:"api-node-host" json:"api-node-host"`
+	APINodePort           uint32                           `mapstructure:"api-node-port" json:"api-node-port"`
+	EncryptConnection     bool                             `mapstructure:"encrypt-connection" json:"encrypt-connection"`
+	RPCEndpoint           string                           `mapstructure:"rpc-endpoint" json:"rpc-endpoint"`
+	WSEndpoint            string                           `mapstructure:"ws-endpoint" json:"ws-endpoint"`
+	MessageContracts      map[string]MessageProtocolConfig `mapstructure:"message-contracts" json:"message-contracts"`
+	SupportedDestinations []string                         `mapstructure:"supported-destinations" json:"allowed-destinations"`
```
minghinmatthewlam, line 52: should the json match the `mapstrcuture` value?

config/config.go
```diff
@@ -37,15 +40,19 @@ type MessageProtocolConfig struct {
...
+
+	// convenience field
```
minghinmatthewlam, line 54: can you add a short descrfiption of how this field is used?

config/config.go
```diff
@@ -187,27 +187,31 @@ func (c *Config) Validate() error {
...
-		sourceSubnetIDs = append(sourceSubnetIDs, subnetID)
+		if sourceChains.Contains(s.ChainID) {
+			return fmt.Errorf("configured source subnets must have unique chain IDs")
+		}
+		sourceChains.Add(s.ChainID)
 
-		chainID, err := ids.FromString(s.ChainID)
-		if err != nil {
-			return nil, nil, fmt.Errorf("invalid subnetID in configuration. error: %v", err)
+		for _, blockchainID := range s.SupportedDestinations {
```
minghinmatthewlam, line 201: thoughts on moving this to `sourceChain.Validate()`? In my mind these checks are still validating the source chain configs, we can pass in the destination chains to validate.

messages/teleporter/message_manager.go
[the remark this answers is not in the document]
minghinmatthewlam: would agree that we should move the destination client filtering logic to `NewRelayer` to prevent checking each message.

minghinmatthewlam (commented):
mostly LGTM, left some questions. If we have a case of source chains A,B,C, and destiantion chains D,E,F. Now we want to add chain G as an allowed destination for only A, we'd also have to add the allow destiantion field of DEF for chains B and C right? Since the chains B and C would otherwise deliver to chain G automatically.

cam-schultz:
> mostly LGTM, left some questions. If we have a case of source chains A,B,C, and destiantion chains D,E,F. Now we want to add chain G as an allowed destination for only A, we'd also have to add the allow destiantion field of DEF for chains B and C right? Since the chains B and C would otherwise deliver to chain G automatically.

Yes, that's correct. If no supported destinations are listed, all configured destination subnets are implicitly allowed. Otherwise, to disallow just a single destination, all other destinations must be listed.

config/config.go
cam-schultz: Yes, nice catch. Fixed now.

config/config.go
cam-schultz: Good call, moved.

config/config.go
cam-schultz: After it's constructed, the configuration is a static singleton, so this pattern makes access more straightforward. The alternative is passing the option from the config down the call stack to where it's used, unnecessarily complicating interfaces along the way.

config/config.go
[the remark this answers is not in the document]
geoff-vball: afaik `ethclient.Client.ChainID` returns the eth `chainID`. I don't think the eth client has any way of knowing the `blockchainID`. 

Also note that in the go code, `chainID` is a big.Int and `blockchainID` is a [32]byte

config/config.go
minghinmatthewlam: I have a preference, though not a strong preference, to stay away from the global variables and to instead pass in the configs, similar to previously how we had global var for destination clients. This also helps prevent in the future from accidentally ovewriting values in the global config

config/config.go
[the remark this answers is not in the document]
cam-schultz: Created an issue here: https://github.com/ava-labs/awm-relayer/issues/70

messages/teleporter/message_manager.go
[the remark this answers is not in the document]
cam-schultz: Moved 👍

config/config.go
cam-schultz: On second thought, I agree we should pass the config down the callstack rather than have a global variable. I've made this change.

config/config.go
```diff
@@ -244,6 +234,28 @@ func (s *SourceSubnet) Validate() error {
 		}
 	}
 
+	// Validate the allowed destinations
+	for _, chainIDs := range s.SupportedDestinations {
+		if _, err := ids.FromString(chainIDs); err != nil {
+			return fmt.Errorf("invalid chainID in source subnet configuration. Provided ID: %s", chainIDs)
+		}
+	}
```
michaelkaplan13, line 242: Is this loop needed? Looks like we have the same check in the iteration below already.

config/config.go
```diff
@@ -244,6 +234,28 @@ func (s *SourceSubnet) Validate() error {
...
+		chainID, err := ids.FromString(blockchainID)
+		if err != nil {
+			return fmt.Errorf("invalid chainID in configuration. error: %v", err)
+		}
+		if !destinationChainIDs.Contains(blockchainID) {
+			return fmt.Errorf("configured source subnet %s has a supported destination blockchain ID %s that is not configured as a destination blockchain",
+				s.SubnetID,
+				blockchainID)
+		}
+		s.supportedDestinations.Add(chainID)
```
michaelkaplan13, line 256: I'd consider moving setting the `supportedDestinations` value to a `initialize` method to be more obvious, but no strong preference either way

config/config.go
```diff
@@ -373,3 +385,27 @@ func (s *DestinationSubnet) GetRelayerAccountInfo() (*ecdsa.PrivateKey, common.A
 	pkBytes = append(pkBytes, pk.PublicKey.Y.Bytes()...)
 	return pk, common.BytesToAddress(crypto.Keccak256(pkBytes)), nil
 }
+
+//
+// Top-level config getters
+//
+
+// GetSourceIDs returns the Subnet and Chain IDs of all subnets configured as a source
+func (c *Config) GetSourceIDs() ([]ids.ID, []ids.ID, error) {
```
michaelkaplan13, line 394: It looks like this function is only called once from `main`, so it doesn't matter too much either way, but I would consider adding `sourceSubnetIDs` and `sourceChainIDs` to fields that are set once and then returned directly by `GetSourceIDs` rather than doing the iteration and construction each time it is called.

relayer/relayer.go
```diff
@@ -173,11 +187,20 @@ func NewRelayer(
 
 // RelayMessage relays a single warp message to the destination chain. Warp message relay requests from the same origin chain are processed serially
 func (r *Relayer) RelayMessage(warpLogInfo *vmtypes.WarpLogInfo, metrics *MessageRelayerMetrics, messageCreator message.Creator) error {
+	// Check that the destination chain ID is supported
+	if len(r.supportedDestinations) > 0 && !r.supportedDestinations.Contains(warpLogInfo.DestinationChainID) {
```
michaelkaplan13, line 191: Can we create a unit test for this case?

cam-schultz: I separated this check to its own function and added a unit test.

config/config.go
cam-schultz: Good call, added a caching step.

config/config.go
cam-schultz: Nope, it's redundant. Removed.

config/config.go
cam-schultz: I think that might be more trouble than it's worth. To do so properly, we'd want to iterate through all supported destinations in the validate method, then do so again to cache them in initialize. Here, we do it in a single pass.

relayer/relayer_test.go
```diff
@@ -6,19 +6,52 @@ package relayer
...
+			expectedResult:     false,
+		},
 	}
-	_, address, err := info.GetRelayerAccountInfo()
 
-	require.NoError(t, err)
-	require.Equal(t, expectedAddress, address.String())
+	for _, testCase := range testCases {
+		result := testCase.relayer.CheckSupportedDestination(testCase.destinationChainID)
+		if result != testCase.expectedResult {
```
michaelkaplan13, line 53: Is there a reason to prefer using `t.Fatalf` rather than `require.Equal`? I'm slightly behind on our desired golang unit test conventions, but am noticing the change here.

config/config.go
```diff
@@ -392,6 +389,11 @@ func (s *DestinationSubnet) GetRelayerAccountInfo() (*ecdsa.PrivateKey, common.A
 
 // GetSourceIDs returns the Subnet and Chain IDs of all subnets configured as a source
 func (c *Config) GetSourceIDs() ([]ids.ID, []ids.ID, error) {
+	// If we've already called this method, return the already constructed result
+	if len(c.sourceSubnetIDs) > 0 && len(c.sourceChainIDs) > 0 {
```
michaelkaplan13, line 393: Nit: We're introducing a couple different places where various of types of initialization occurs in this PR (in `SourceSubnet.Validate` and also here). I think it will be more maintainable to structure the initialization steps using the same patterns. Could we move this step to `Config.Validate` to match?

michaelkaplan13 (commented):
One small comment and one question, but otherwise LGTM

config/config.go
cam-schultz: Sure thing. Moved this to `Config.Validate`, and just return the stored values in this method.

relayer/relayer_test.go
cam-schultz: No reason other than my own inconsistency 😅 . Changed to `require.Equal`

config/config.go
```diff
@@ -187,30 +185,47 @@ func (c *Config) Validate() error {
...
+		sourceChains.Add(s.ChainID)
+	}
 
-// GetSourceIDs returns the Subnet and Chain IDs of all subnets configured as a source
-func (cfg *Config) GetSourceIDs() ([]ids.ID, []ids.ID, error) {
+	// Store the source subnet and chain IDs for future use
 	var sourceSubnetIDs []ids.ID
 	var sourceChainIDs []ids.ID
-	for _, s := range cfg.SourceSubnets {
+	for _, s := range c.SourceSubnets {
```
michaelkaplan13, line 203: Nit: Can we do this in the existing loop over `c.SourceSubnets` above instead of adding a separate iteration?

cam-schultz: Sure thing. Also clarified some of the surrounding comments.

michaelkaplan13 (approved):
LGTM

geoff-vball (approved):
LGTM!

PR #61, opened 2023-10-18
Remove Teleporter interfacing code

cam-schultz:
## Why this should be merged
https://github.com/ava-labs/teleporter/pull/52 removes the `teleporter` dependency on `awm-relayer` by adding ABI packing/unpacking functions and Teleporter utilities to `teleporter`. This PR removes the duplicate implementations of those functions from `awm-relayer`.

## How this works
Functions from `package teleporter` are removed; callsites instead reference the versions in the `teleporter` repository.

## How this was tested
CI


michaelkaplan13 (approved):
🙌 Seeing this cleanup gives me more confidence that the golang directory structure/contents of the `teleporter` repo are correct now.

minghinmatthewlam (approved):
LGTM

PR #67, opened 2023-10-24
separate info API cfg

cam-schultz:
## Why this should be merged
Currently the P-Chain API URL passed in the config is used for two purposes:
- To get the canonical validator set for a subnet
- To establish peers in the app request network via the info API

The latter case requires the info API be enabled on the node. For the public RPC (`api.avax.network`), the info API is disabled.

## How this works
The change separates the Info API URL from the P-Chain API URL, giving each of the above use cases a dedicated URL. They may still be the same node, but we no longer require the info API be enabled on the P-Chain API node. 

## How this was tested
CI

## How is this documented
Modified `Configuration` section in README


cam-schultz:
Closing in favor of https://github.com/ava-labs/awm-relayer/pull/68

PR #68, opened 2023-10-24
Simplify info client usage

cam-schultz:
## Why this should be merged
Currently the P-Chain API URL passed in the config is used for two purposes:

- To get the canonical validator set for a subnet
- To establish peers in the app request network via the info API
The latter case requires that the info API expose the `GetNodeIP` method on the node. For the public RPC (api.avax.network), this method is disabled.

## How this works
Removes the calls to the methods `GetNodeIP` and `GetNodeID` on the configured P-Chain API node. Instead, we get the list of peer IPs and IDs from `GetPeers`, which is enabled on the public RPC server. 

Also fixes an edge case in which it's possible to connect to fewer than the intended number of peers on startup.

## How this was tested
CI

## How is this documented
N/A

### DIFF
--- a/peers/app_request_network.go
+++ b/peers/app_request_network.go
@@ -103,24 +103,6 @@ func NewNetwork(
 		)
 		return nil, nil, err
 	}
-	ip, err := infoClient.GetNodeIP(context.Background())
-	if err != nil {
-		logger.Error(
-			"Failed to get ip",
-			zap.Error(err),
-		)
-		return nil, nil, err
-	}
-	id, _, err := infoClient.GetNodeID(context.Background())
-	if err != nil {
-		logger.Error(
-			"Failed to get node id",
-			zap.Error(err),
-		)
-		return nil, nil, err
-	}
-	beaconIPs = append(beaconIPs, ip)
-	beaconIDs = append(beaconIDs, id.String())
 
 	var indices []uint64
 	// If we have more peers than numInitialTestPeers, take a random sample of numInitialTestPeers peers
--- a/peers/app_request_network.go
+++ b/peers/app_request_network.go
@@ -6,6 +6,7 @@ package peers
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"sync"
 	"time"
 
@@ -17,7 +18,6 @@ import (
 	"github.com/ava-labs/avalanchego/utils/constants"
 	"github.com/ava-labs/avalanchego/utils/ips"
 	"github.com/ava-labs/avalanchego/utils/logging"
-	"github.com/ava-labs/avalanchego/utils/sampler"
 	"github.com/ava-labs/avalanchego/utils/set"
 	"github.com/prometheus/client_golang/prometheus"
 	"go.uber.org/zap"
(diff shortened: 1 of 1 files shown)


michaelkaplan13 (approved):
🙌 This is a great improvement to make. The changes all LGTM, but do want to note that when trying to use the public API in the relayer configuration now with these changes, I hit the follow error when it is attempting to relay a message:

```
{"level":"error","timestamp":"2023-10-24T14:46:09.656-0400","logger":"awm-relayer","caller":"relayer/message_relayer.go:411","msg":"Failed to get the canonical subnet validator set","subnetID":"1M439meCk9SDQYJSrDBJJeRntk1FtmSaaZakSeLFvcJonxGUM","error":"failed to fetch validator set (P-Chain Height: 123702, SubnetID: 1M439meCk9SDQYJSrDBJJeRntk1FtmSaaZakSeLFvcJonxGUM): failed to decode client response: the method platform.getValidatorsAt is not available"}
```

I think these changes are self-standing and still worth moving forward with as is though.

peers/app_request_network.go
```diff
@@ -103,44 +103,33 @@ func NewNetwork(
...
+		}
+	}
+	if len(beaconIPs) == 0 {
+		logger.Error(
+			"Failed to find any peers to connect to",
+			zap.Error(err),
+		)
+		return nil, nil, err
+	}
+	if len(beaconIPs) < numInitialTestPeers {
```
minghinmatthewlam, line 127: why do we need to limit the number of peers to 5 again? IIUC when we do the signature aggregation process it could take longer in terms of us sending out more signature requests, but would increase likelihood of having enough stake threshold in the peers. Are 5 peers generally enough for exceeding stake threshold?

minghinmatthewlam:
separate of PR changes, but was looking at the relayer's [network manual track](https://github.com/ava-labs/awm-relayer/blob/main/peers/app_request_network.go#L168). Supposedly the network will keep attempting the connection with exponential backoffs, should we look at timing out this operation?

minghinmatthewlam (approved):
LGTM, left some questions for clarification

peers/app_request_network.go
```diff
@@ -103,44 +103,33 @@ func NewNetwork(
...
+		logger.Warn(
+			"Failed to find a full set of peers to connect to on startup",
+			zap.Int("connectedPeers", len(beaconIPs)),
+			zap.Int("expectedConnectedPeers", numInitialTestPeers),
+		)
```
bernard-avalabs, line 132: Should we abort if the number of connectedPeers is 0?

peers/app_request_network.go
```diff
@@ -103,44 +103,33 @@ func NewNetwork(
...
 	}
 
 	for i, beaconIDStr := range beaconIDs {
```
bernard-avalabs, line 135: What if we are unable to connect to any of the initial test peers? Will we select another sample of peers?

bernard-avalabs (commented):
Generally looks good to me. Left a few questions.

peers/app_request_network.go
cam-schultz: This is just the list of initial peers to manually connect to in order to initialize the app request network. When requesting signatures, we connect to all peers from which we are requesting signatures.

peers/app_request_network.go
cam-schultz: In that case, we return an error a few lines above here.

peers/app_request_network.go
cam-schultz: That's a good question. As Matt mentioned [above](https://github.com/ava-labs/awm-relayer/pull/68#issuecomment-1777882938), `ManuallyTrack` will continuously attempt to connect. We should look into what the failure modes for that function call are, and consider how to handle them. Right now if we are unable to connect to any peers on initialization, we will attempt to connect to the subnet validators when relaying a message. So there is some form of redundancy, but an explicit resampling on startup would be more robust.

PR #76, opened 2023-10-31
Integrate subnet-evm 0.5.8

gwen917:
## Why this should be merged
- Integrate with subnet-evm 0.5.8 and avalancheGo 1.10.14

## How this was tested
- Unit tests
- CI

## How is this documented


messages/message_manager.go
```diff
@@ -23,7 +23,7 @@ import (
 type MessageManager interface {
 	// ShouldSendMessage returns true if the message should be sent to the destination chain
 	// If an error is returned, the boolean should be ignored by the caller.
-	ShouldSendMessage(warpMessageInfo *vmtypes.WarpMessageInfo, destinationChainID ids.ID) (bool, error)
+	ShouldSendMessage(warpMessageInfo *vmtypes.WarpMessageInfo) (bool, ids.ID, error)
```
michaelkaplan13, line 26: Could you update the comment that explains the return values here.

messages/teleporter/message_manager_test.go
```diff
@@ -49,6 +49,17 @@ var (
 	}
 )
 
+func getChainIDBytes(id string) ([32]byte, error) {
+	chainID, err := ids.FromString(id)
```
michaelkaplan13, line 53: If I'm reading this helper correctly, `decodeChainID` may be a more descriptive name. 

Also, I think we should be able to have the implementation be just
```
return ids.FromString(id)
```

because `ids.ID` is an alias for `[32]byte` already. If that's the case, it may be easiest to remove it all together since it's already a single function call.

relayer/message_relayer.go
```diff
@@ -95,45 +92,45 @@ func (r *messageRelayer) relayMessage(warpMessageInfo *vmtypes.WarpMessageInfo,
...
+func (r *messageRelayer) createSignedMessage(requestID uint32, destinationChainID ids.ID) (*warp.Message, error) {
 	r.logger.Info(
 		"Starting relayer routine",
-		zap.String("destinationChainID", r.destinationChainID.String()),
+		zap.String("destinationChainID", destinationChainID.String()),
 	)
 
 	// Get the current canonical validator set of the source subnet.
-	validatorSet, totalValidatorWeight, err := r.getCurrentCanonicalValidatorSet()
+	validatorSet, totalValidatorWeight, err := r.getCurrentCanonicalValidatorSet(destinationChainID)
```
michaelkaplan13, line 133: Digging in to the implementation of `getCurrentCanonicalValidatorSet`, I see why it's necessary to now pass the `destinationChainID`, but looking at the call here, I was confused why we pass the destination chain ID when  trying to look up the source subnet validator set.

Could we make this more intuitive by creating a helper along the lines of "determineSigningSubnetD" that performs the check on whether or not the message was sent from the primary network, and then have `getCurrentCanonicalValidatorSet` take that subnet ID as a parameter to look up the validator set for?

tests/basic_relay.go
```diff
@@ -284,3 +284,12 @@ func BasicRelay() {
 	relayerCancel()
 	_ = relayerCmd.Wait()
 }
+
+// HashSliceToBytes serializes a []common.Hash into a tightly packed byte array.
+func HashSliceToBytes(hashes []common.Hash) []byte {
```
michaelkaplan13, line 289: I'd recommend moving this into `tests/utils` somewhere.

messages/teleporter/message_manager_test.go
```diff
@@ -49,6 +49,17 @@ var (
...
+	if err != nil {
+		return [32]byte{}, err
+	}
+	chainIDBytes := common.Hex2Bytes(chainID.Hex())
+	var array32 [32]byte
+	copy(array32[:], chainIDBytes)
+	return array32, nil
+}
+
 func TestShouldSendMessage(t *testing.T) {
```
cam-schultz, line 63: In this test, let's use `PackTeleporterMessage` instead of `PackSendCrossChainMessageEvent`

messages/teleporter/message_manager_test.go
```diff
@@ -154,18 +152,33 @@ func TestShouldSendMessage(t *testing.T) {
 	}
 	for _, test := range testCases {
 		t.Run(test.name, func(t *testing.T) {
+			validTeleporterMessage.DestinationChainID, err = getChainIDBytes(test.destinationChainID.String())
```
cam-schultz, line 155: I think we can just do `validTeleporterMessage.DestinationChainID = test.destinationChainID`, since they're both `[32]byte`. We might have to do some casts, but should be lightweight.

messages/teleporter/message_manager.go
```diff
@@ -91,21 +91,27 @@ func isAllowedRelayer(allowedRelayers []common.Address, eoa common.Address) bool
 }
 
 // ShouldSendMessage returns true if the message should be sent to the destination chain
-func (m *messageManager) ShouldSendMessage(warpMessageInfo *vmtypes.WarpMessageInfo, destinationChainID ids.ID) (bool, error) {
+func (m *messageManager) ShouldSendMessage(warpMessageInfo *vmtypes.WarpMessageInfo) (bool, ids.ID, error) {
```
cam-schultz, line 94: Rather than returning the destination ID in this method, I think we should add a separate method to `MessageManager` called `GetDestinationID` that returns an `ids.ID`. That would make more sense logically, and would make this interface more general for future use cases such as anycasting. For the anycast case, we can generalize `GetDestinationID` to something like `GetDestinations`, but I think just adding `GetDestinationID` is fine for now.

cam-schultz: We could also move the call to `messageManager.GetDestinationID()` to `RelayMessage()`, then pass in the `destinationChainID` to `NewMessageRelayer`. That way we can still compose the message relayer with the `destinationChainID`, instead of having to pass the non-member variable around the function calls like you do in this PR.

relayer/relayer.go
```diff
@@ -230,7 +220,7 @@ func (r *Relayer) RelayMessage(warpLogInfo *vmtypes.WarpLogInfo, metrics *Messag
 	}
 
 	// Create and run the message relayer to attempt to deliver the message to the destination chain
-	messageRelayer := newMessageRelayer(r.logger, metrics, r, warpMessageInfo.WarpUnsignedMessage, warpLogInfo.DestinationChainID, r.responseChan, messageCreator)
+	messageRelayer := newMessageRelayer(r.logger, metrics, r, warpMessageInfo.WarpUnsignedMessage, r.responseChan, messageCreator)
```
cam-schultz, line 223: See my comment above about refactoring this call

relayer/relayer.go
```diff
@@ -187,16 +187,6 @@ func NewRelayer(
 
 // RelayMessage relays a single warp message to the destination chain. Warp message relay requests from the same origin chain are processed serially
 func (r *Relayer) RelayMessage(warpLogInfo *vmtypes.WarpLogInfo, metrics *MessageRelayerMetrics, messageCreator message.Creator) error {
-	// Check that the destination chain ID is supported
-	if !r.CheckSupportedDestination(warpLogInfo.DestinationChainID) {
```
cam-schultz, line 191: We should keep this in. If we extract the destination chain ID from `messageManager` in this function, we can then call `CheckSupportedDestination` with that value.

tests/utils/UniversalTeleporterDeployerAddress.txt
cam-schultz, line 1: Would you mind adding a README in `tests/utils` describing what these three files are, as well as the `teleporter` git commit from which they are sourced?

messages/teleporter/message_manager_test.go
gwen917: Fixed.

messages/teleporter/message_manager_test.go
gwen917: fixed

relayer/relayer.go
gwen917: fixed

messages/teleporter/message_manager_test.go
gwen917: fixed

messages/teleporter/message_manager.go
gwen917: fixed

tests/utils/UniversalTeleporterDeployerAddress.txt
gwen917: added in the ReadMe

README.md
```diff
@@ -148,6 +148,13 @@ Next, provide the path to the `subnet-evm` repository and the path to a writeabl
 ```bash
 ./scripts/e2e_test.sh --local --subnet-evm ~/subnet-evm --data-dir ~/tmp/e2e-test
 ```
+
+If a new version of Teleporter contract is deployed, we need to update the following files:
+- `tests/utils/UniversalTeleporterDeployerAddress.txt`
+- `tests/utils/UniversalTeleporterDeployerTransaction.txt`
+- `tests/utils/UniversalTeleporterMessagerContractAddress.txt`
+We use the values in these files to depoly the Teleporter contract used in the tests.
```
michaelkaplan13, line 156:
```suggestion
The E2E tests use the `TeleporterMessenger` contract deployment transaction specified in the following files:
- `tests/utils/UniversalTeleporterDeployerAddress.txt`
- `tests/utils/UniversalTeleporterDeployerTransaction.txt`
- `tests/utils/UniversalTeleporterMessagerContractAddress.txt`
To update the version of Teleporter used by the E2E tests, update these values with the latest contract deployment information.
```

Nit: To be consistent with the wording of the rest of the README/documentation, lets avoid the usage of "we".

messages/teleporter/message_manager.go
```diff
@@ -90,6 +90,20 @@ func isAllowedRelayer(allowedRelayers []common.Address, eoa common.Address) bool
 	return false
 }
 
+func (m *messageManager) GetDestinationChainID(warpMessageInfo *vmtypes.WarpMessageInfo) (ids.ID, error) {
+	// Unpack the teleporter message and add it to the cache
```
michaelkaplan13, line 94: This comment likely needs updating.

README.md
```diff
@@ -148,6 +148,13 @@ Next, provide the path to the `subnet-evm` repository and the path to a writeabl
 ```bash
 ./scripts/e2e_test.sh --local --subnet-evm ~/subnet-evm --data-dir ~/tmp/e2e-test
 ```
+
+The E2E tests use the `TeleporterMessenger` contract deployment transaction specified in the following files:
+- `tests/utils/UniversalTeleporterDeployerAddress.txt`
+- `tests/utils/UniversalTeleporterDeployerTransaction.txt`
+- `tests/utils/UniversalTeleporterMessagerContractAddress.txt`
+To update the version of Teleporter used by the E2E tests, update these values with the latest contract deployment information.
```
cam-schultz, line 156: Can you link to the deployment utils README in `teleporter`?

messages/teleporter/message_manager.go
```diff
@@ -90,6 +90,20 @@ func isAllowedRelayer(allowedRelayers []common.Address, eoa common.Address) bool
 	return false
 }
 
+func (m *messageManager) GetDestinationChainID(warpMessageInfo *vmtypes.WarpMessageInfo) (ids.ID, error) {
+	// Unpack the teleporter message to get the destination chain ID
```
cam-schultz, line 94:
```suggestion
	// Unpack the Teleporter message to get the destination chain ID
```

messages/teleporter/message_manager.go
```diff
@@ -90,6 +90,20 @@ func isAllowedRelayer(allowedRelayers []common.Address, eoa common.Address) bool
...
+	teleporterMessage, err := teleportermessenger.UnpackTeleporterMessage(warpMessageInfo.WarpPayload)
```
cam-schultz, line 95: Can you put the unpacked Teleporter message into `teleporterMessageCache`, and update `ShouldSendMessage` to check the cache first? That way we don't need to unpack the message in both methods.

cam-schultz (commented):
A couple more cleanup comments 👍

messages/teleporter/message_manager.go
gwen917: fixed

README.md
gwen917: This [contract deployment link](https://github.com/ava-labs/teleporter/blob/main/contract-deployment/README.md) is not found. Where should I link in `teleporter`?

cam-schultz: https://github.com/ava-labs/teleporter/tree/main/utils/contract-deployment

messages/teleporter/message_manager.go
```diff
@@ -101,19 +101,32 @@ func (m *messageManager) GetDestinationChainID(warpMessageInfo *vmtypes.WarpMess
...
 
 // ShouldSendMessage returns true if the message should be sent to the destination chain
 func (m *messageManager) ShouldSendMessage(warpMessageInfo *vmtypes.WarpMessageInfo, destinationChainID ids.ID) (bool, error) {
-	// Unpack the teleporter message and add it to the cache
-	teleporterMessage, err := teleportermessenger.UnpackTeleporterMessage(warpMessageInfo.WarpPayload)
-	if err != nil {
-		m.logger.Error(
-			"Failed unpacking teleporter message.",
-			zap.String("warpMessageID", warpMessageInfo.WarpUnsignedMessage.ID().String()),
+	warpMssageID := warpMessageInfo.WarpUnsignedMessage.ID()
```
cam-schultz, line 112: nit: typo should be `warpMessageID`

messages/teleporter/message_manager.go
```diff
@@ -90,16 +90,43 @@ func isAllowedRelayer(allowedRelayers []common.Address, eoa common.Address) bool
...
 	if err != nil {
 		m.logger.Error(
 			"Failed unpacking teleporter message.",
 			zap.String("warpMessageID", warpMessageInfo.WarpUnsignedMessage.ID().String()),
 		)
-		return false, err
+		return ids.ID{}, err
+	}
+
+	// Cache the message so it can be reused in SendMessage
```
cam-schultz, line 104:
```suggestion
	// Cache the message so it can be reused in ShouldSendMessage and SendMessage
```

vms/evm/contract_message_test.go
```diff
@@ -59,15 +53,15 @@ func TestUnpack(t *testing.T) {
...
+			networkID:     0x539,
```
cam-schultz, line 57: Can you add a comment explaining what this network ID is?

cam-schultz (commented):
Just a few small nits

messages/teleporter/message_manager.go
```diff
@@ -90,16 +90,43 @@ func isAllowedRelayer(allowedRelayers []common.Address, eoa common.Address) bool
 	return false
 }
 
-// ShouldSendMessage returns true if the message should be sent to the destination chain
-func (m *messageManager) ShouldSendMessage(warpMessageInfo *vmtypes.WarpMessageInfo, destinationChainID ids.ID) (bool, error) {
-	// Unpack the teleporter message and add it to the cache
+func (m *messageManager) GetDestinationChainID(warpMessageInfo *vmtypes.WarpMessageInfo) (ids.ID, error) {
+	// Unpack the Teleporter message to get the destination chain ID
```
minghinmatthewlam, line 94: shoudl we check the cache here first to see if there's a matching entry?

minghinmatthewlam: I see that the order is for `GetDestinationChainID` to be called first, but think fine to add a check in case this isn't always the case now or in future.

messages/teleporter/message_manager.go
```diff
@@ -90,16 +90,43 @@ func isAllowedRelayer(allowedRelayers []common.Address, eoa common.Address) bool
...
+		var err error
+		teleporterMessage, err = teleportermessenger.UnpackTeleporterMessage(warpMessageInfo.WarpPayload)
+		if err != nil {
+			m.logger.Error(
+				"Failed unpacking teleporter message.",
+				zap.String("destinationChainID", destinationChainID.String()),
+				zap.String("warpMessageID", warpMssageID.String()),
+			)
+			return false, err
+		}
```
minghinmatthewlam, line 129: we should put in cache here

minghinmatthewlam: Actually we should only put to cache if `cache.Get` didn't return an entry anyways, so no need to put at the end in case we already have cache entry.

minghinmatthewlam: Think we can also put to cache for `sendMessage` if no cache entry already.

messages/teleporter/message_manager.go
gwen917: We added cache check for `ShouldSendMessage`.

minghinmatthewlam: can we add a check for `cache.Get` here too? Commented below, but we can do a helper function for checking:
```
get warp message id
try get cache
if !ok 
  unpack teleporter message
  if err handle error
  put teleporter message in cache
return TeleporterMessage
```

README.md
gwen917: added the link

messages/teleporter/message_manager.go
gwen917: fixed

messages/teleporter/message_manager.go
gwen917: added a helper function that checks the cache first and parse from message if not exist in cache. `GetDestinationID`, `ShouldSend` and `SendMessage` will use the helper function to get teleporter message

messages/teleporter/message_manager.go
gwen917: Commented above.

messages/teleporter/message_manager.go
```diff
@@ -90,14 +90,54 @@ func isAllowedRelayer(allowedRelayers []common.Address, eoa common.Address) bool
 	return false
 }
 
+// Helper to get the teleporter message from the cache or from the warp message payload
```
minghinmatthewlam, line 93:
```suggestion
// getTeleporterMessage returns the Warp message's corresponding Teleporter message from the cache if it exists. Otherwise parses the Warp message payload.
```

minghinmatthewlam: nit: can we put this helper function below the interface implementation functions

gwen917: fixed

messages/teleporter/message_manager.go
gwen917: fixed

messages/teleporter/message_manager.go
```diff
@@ -291,3 +299,27 @@ func (m *messageManager) SendMessage(signedMessage *warp.Message, parsedVMPayloa
 	)
 	return nil
 }
+
+// getTeleporterMessage returns the Warp message's corresponding Teleporter message from the cache if it exists.
+// Otherwise parses the Warp message payload.
+func (m *messageManager) getTeleporterMessage(warpMessageID ids.ID, warpPayload []byte) (*teleportermessenger.TeleporterMessage, error) {
```
michaelkaplan13, line 305: Nit: `parseTeleporterMessage` may be a clearer name for this function.

scripts/versions.sh
```diff
@@ -4,7 +4,7 @@
 
 # Set up the versions to be used
 AWM_RELAYER_VERSION=${AWM_RELAYER_VERSION:-'v0.2.2'}
-SUBNET_EVM_VERSION=${SUBNET_EVM_VERSION:-'v0.5.6'}
+SUBNET_EVM_VERSION=${SUBNET_EVM_VERSION:-'v0.5.8'}
 # Don't export them as they're used in the context of other calls
-AVALANCHE_VERSION=${AVALANCHE_VERSION:-'v1.10.10'}
+AVALANCHE_VERSION=${AVALANCHE_VERSION:-'v1.10.14'}
```
michaelkaplan13, line 9: Does it work to go straight to AvalancheGo v1.10.15 to avoid and extra update?

messages/teleporter/message_manager.go
gwen917: fixed

scripts/versions.sh
gwen917: We can update it to v1.10.15. But in the go.mod, we still need to use v1.10.14 since subnet-evm v0.5.8 uses v1.10.14. Do we want keep the same version?

vms/evm/contract_message_test.go
gwen917: I got the data from the integration test. When we did integration test, we sent cross chain message. The data is from the log event of sendCrossChain message. Then unpack the data, we can get [unsigned message](https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/warp/unsigned_message.go#L15) that has networkID.

cam-schultz: Gotcha. Decoding the hex, this is `1337`, the network ID used by avalanche-cli for local networks. This is defined already in the variable [LocalNetworkID](https://github.com/ava-labs/awm-relayer/blob/main/peers/app_request_network.go#L27). Let's just use that directly here.

gwen917: cool, fixed to use LocalNetworkID

scripts/versions.sh
michaelkaplan13: Right. Lets keep `v1.10.14` for now 👍 

We'll later update both subnet-evm and avlanchego to the latest versions, which hopefully don't have any further breaking changes.

michaelkaplan13 (dismissed):

tests/utils/utils.go
```diff
@@ -55,3 +56,12 @@ func ReadHexTextFile(filename string) string {
 	Expect(err).Should(BeNil())
 	return string(fileData)
 }
+
+// HashSliceToBytes serializes a []common.Hash into a tightly packed byte array.
+func HashSliceToBytes(hashes []common.Hash) []byte {
```
minghinmatthewlam, line 61: do we need this? We can use subnet-evms

minghinmatthewlam (approved):
LGTM!

tests/utils/utils.go
gwen917: Fixed. Previously, I thought it was not provided in the 0.5.8.

cam-schultz (approved):
LGTM

PR #89, opened 2023-11-15
subscriber.ProcessFromHeight: paginate `eth_getLogs` requests

### LINKED ISSUE #29: Consider catch up behavior if relayer has missed a large number of blocks
**Context and scope**
With https://github.com/ava-labs/awm-relayer/pull/20, the relayer has the ability to retrieve historical Warp logs and process them on startup. This is done through a call to `eth_getLogs`, which most clients limit to 500-2000 blocks. As a conservative bound, we currently bound the number of blocks to retrieve from `eth_getLogs` to 200 blocks.

We should consider the case in which the relayer needs to catch up more than the hardcoded maximum number of blocks (currently 200). Some possible approaches are:
- Split the blocks into multiple `eth_getLogs` calls, each of which is bounded to a certain number blocks
- Allow the user to configure the maximum lookback window size, allowing them to tailor it to the client running the API node

**Open questions**
In order to determine reasonable bounds, we should consider what the expected maximum downtime for _all relayer replications_ would be. With enough fault tolerance, message deliver would not be constrained by an individual relayer's lookback window at all.

feuGeneA:
I recommend reviewing each commit one at a time, because there are a significant amount of whitespace changes, but those are largely in the "just `gofmt -w`" commit, and each of the other commits are quite tiny and easy to review independently.

- [x] Paginate `eth_getLogs` requests
- [x] Provide a configuration option to bypass catching up on historical blocks

## Why this should be merged

Because it fixes #29 

## How this works

When catching up on missed blocks, limits the number of blocks per `eth_getLogs` request (to the value of a constant variable), making multiple `eth_getLogs` requests if necessary.

## How this was tested

Via the included unit tests

## How is this documented

In the comments describing the function `ProcessFromHeight`


vms/evm/subscriber.go
```diff
@@ -154,63 +157,71 @@ func (s *subscriber) ProcessFromHeight(height *big.Int) error {
...
-		Topics:    warpFilterQuery.Topics,
-		Addresses: warpFilterQuery.Addresses,
-		FromBlock: height,
-		ToBlock:   toBlock,
-	}
-	logs, err := ethClient.FilterLogs(context.Background(), initializationFilterQuery)
-	if err != nil {
-		s.logger.Error(
-			"Failed to get logs on initialization",
+	processBlockRange := func(fromBlock *big.Int, toBlock *big.Int) error {
```
michaelkaplan13, line 163: I'd recommend making `processBlockRange` it's own named (unexported) function in this file, just for readability.

feuGeneA: i like that, thanks. addressed in [4de3aba](https://github.com/ava-labs/awm-relayer/pull/89/commits/4de3aba8b1e1d3fd3098804d7ac57bf18946e0ea)

relayer/relayer.go
```diff
@@ -129,6 +130,34 @@ func NewRelayer(
...
+				zap.String("chainID", r.sourceChainID.String()),
+				zap.Error(err),
+			)
+		}
+	}
+
+	return &r, sub, nil
+}
+
+func processMissedBlocks(
```
cam-schultz, line 156: Can we have `r *Relayer` be a receiver rather than an argument?

config/config.go
```diff
@@ -64,13 +64,14 @@ type DestinationSubnet struct {
...
-	SourceSubnets      []SourceSubnet      `mapstructure:"source-subnets" json:"source-subnets"`
-	DestinationSubnets []DestinationSubnet `mapstructure:"destination-subnets" json:"destination-subnets"`
+	LogLevel            string              `mapstructure:"log-level" json:"log-level"`
+	NetworkID           uint32              `mapstructure:"network-id" json:"network-id"`
+	PChainAPIURL        string              `mapstructure:"p-chain-api-url" json:"p-chain-api-url"`
+	EncryptConnection   bool                `mapstructure:"encrypt-connection" json:"encrypt-connection"`
+	StorageLocation     string              `mapstructure:"storage-location" json:"storage-location"`
+	SourceSubnets       []SourceSubnet      `mapstructure:"source-subnets" json:"source-subnets"`
+	DestinationSubnets  []DestinationSubnet `mapstructure:"destination-subnets" json:"destination-subnets"`
+	ProcessMissedBlocks bool                `mapstructure:"process-missed-blocks" json:"process-missed-blocks"`
```
cam-schultz, line 74: This config option is a good idea. Previously to disable missed block processing, the database itself would need to be updated outside of the relayer application.

vms/evm/subscriber.go
```diff
@@ -61,6 +61,7 @@ type subscriber struct {
 
 	logger logging.Logger
 	db     database.RelayerDatabase
+	dial   func(url string) (ethclient.Client, error)
```
cam-schultz, line 64: What's the motivation for adding this as a field? It doesn't look like we're adding any abstraction on top of `ethclient.Client`.

cam-schultz: Oh I see, we're using it to mock `ethclient.Dial` in the unit test. Can you add a comment explaining this?

vms/evm/subscriber_test.go
```diff
@@ -0,0 +1,130 @@
...
+				Addresses: []common.Address{
+					warp.ContractAddress,
+				},
+				FromBlock: big.NewInt(fromBlock),
+				ToBlock:   big.NewInt(toBlock),
+			},
+		).Return([]types.Log{}, nil).Times(1)
+	}
+
+	min := func(a, b int64) int64 {
```
cam-schultz, line 103: Not recommending we update in this PR, but in Go 1.21 we can use the built-in `min`: https://tip.golang.org/ref/spec#Min_and_max

vms/evm/subscriber.go
```diff
@@ -154,25 +157,37 @@ func (s *subscriber) ProcessFromHeight(height *big.Int) error {
...
-		s.logger.Warn(
-			fmt.Sprintf("Requested to process too many blocks. Processing only the most recent %d blocks", MaxBlocksToProcess),
-			zap.String("requestedBlockHeight", height.String()),
-			zap.String("latestBlockHeight", toBlock.String()),
-			zap.String("chainID", s.chainID.String()),
-		)
-		height = big.NewInt(0).Add(toBlock, big.NewInt(-MaxBlocksToProcess))
+	bigLatestBlock := big.NewInt(0).SetUint64(latestBlock)
+
+	for fromBlock := height; fromBlock.Cmp(bigLatestBlock) <= 0; /*see post statement in body*/ {
```
cam-schultz, line 162: Since `height` is a `*big.Int`, incrementing `fromBlock` will also increment `height`. This isn't a problem now, but could cause a bug in the future if we expand this function to use `height` after this loop. We should initialize the loop using `fromBlock := new(big.Int).Set(height)

I also think having the post statement in the loop body is error prone, as there's nothing stopping somebody from adding code below the post statement. One workaround would be to put the loop body in an anonymous function and execute the post statement in a `defer`, guaranteeing that it's the last statement executed in the body. We'd want to comment that clearly though. Neither way is particularly clean, so I leave it up to you to decide.

cam-schultz (commented):
Generally LGTM! Left a few cleanup comments and documentation requests. 

I'll also think a bit more about how to mock `ethclient.Dial`, as I generally am not a fan of modifying a class' implementation to make it testable. There might be an opportunity for a cleaner design. In any case, it's all behind the Subscriber interface anyway, so not a huge deal either way.

relayer/relayer.go
feuGeneA: Thanks for this. Still getting used to being in Go land.  :) Addressed in https://github.com/ava-labs/awm-relayer/pull/89/commits/48f95aa6e25ec99bb654f26ba5ffa861588a2bfc

vms/evm/subscriber.go
feuGeneA: added a comment in https://github.com/ava-labs/awm-relayer/pull/89/commits/6618b41f81d0bc8edfda6b83c1f21b9a7856ceca

vms/evm/subscriber_test.go
feuGeneA: I realized this while implementing, but thank you for pointing it out, because it inspired me to document this via a `TODO: delete this` type comment in the code, in https://github.com/ava-labs/awm-relayer/pull/89/commits/6618b41f81d0bc8edfda6b83c1f21b9a7856ceca

vms/evm/subscriber.go
feuGeneA: Good call on the pointer side effect, I'm still getting used to Go land, so this was a great lesson, thank you.

Regarding the error prone for loop, thank you also for pushing on that. What I submitted here was my second iteration, and that was better than the first, but I still wasn't completely happy with it. Your workaround suggestion was good food for thought; I didn't take it, but it made me find a way to do it that _is_ particularly clean (though with a tiny bit of computational inefficiency (calling the `toBlock()` function variable twice per loop iteration), but I thought the tradeoff for cleanliness over maximum efficiency was a good one). I revamped this loop in https://github.com/ava-labs/awm-relayer/pull/89/commits/c4807558f8e8a15905d1854a03bb8660ea2370ba . let me know what you think.

feuGeneA:
> I'll also think a bit more about how to mock `ethclient.Dial`, as I generally am not a fan of modifying a class' implementation to make it testable. There might be an opportunity for a cleaner design. In any case, it's all behind the Subscriber interface anyway, so not a huge deal either way.

In my humble opinion, I'm okay with modifying a class' implementation to make it testable. Where I draw the line is at modifying a class' _interface_ to make it testable. Given that the `dial` field on this struct is "private", I felt this was a reasonable way to introduce a seam, but I'm totally open to ideas on a cleaner design.

michaelkaplan13 (commented):
**edited** Stale comment from older commit. Will add comment on current commit now.

vms/evm/subscriber.go
```diff
@@ -154,25 +159,39 @@ func (s *subscriber) ProcessFromHeight(height *big.Int) error {
...
+	bigLatestBlock := big.NewInt(0).SetUint64(latestBlock)
+
+	// helper function to determine the upper bound of a block range given
+	// the lower bound:
+	toBlock := func(fromBlock *big.Int) (toBlock *big.Int) {
+		toBlock = big.NewInt(0).Add(fromBlock, big.NewInt(MaxBlocksPerRequest))
+		if toBlock.Cmp(bigLatestBlock) > 0 {
+			toBlock.Set(bigLatestBlock)
+		}
+		return
```
michaelkaplan13, line 171: If we're going to have this as a helper function, can we move it out to it's own function and call it `calculateToBlock`? That should hopefully make easier to read and easier to unit test.

That being said, I don't think a helper function is needed in this case. Could we make the for loop:
```
blockIncrement := big.NewInt(MaxBlocksPerRequest + 1)
for fromBlock := big.NewInt(0).Set(height); fromBlock.Cmp(bigLatestBlock) <= 0; fromBlock.Add(fromBlock, blockIncrement) {

**Comment by geoff-vball** [comment]
@feuGeneA I'm going to change your life 
Screenshot 2023-11-22 at 11 07 51 AM

File: vms/evm/subscriber.go (hunk 7 above)
Comment (geoff-vball) [reply vms/evm/subscriber.go:171]: Doesn't this get us `MaxBlocksPerRequest + 1` blocks per request?

File: vms/evm/subscriber.go [hunk 8]
```diff
@@ -154,25 +159,39 @@ func (s *subscriber) ProcessFromHeight(height *big.Int) error {
...
+func (s *subscriber) processBlockRange(
+	ethClient ethclient.Client,
+	fromBlock, toBlock *big.Int,
+) error {
 	initializationFilterQuery := interfaces.FilterQuery{
 		Topics:    warpFilterQuery.Topics,
 		Addresses: warpFilterQuery.Addresses,
-		FromBlock: height,
+		FromBlock: fromBlock,
 		ToBlock:   toBlock,
```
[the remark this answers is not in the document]
feuGeneA: I played around with some public ethereum nodes and found that calling `eth_getLogs` with a higher-than-latest `toBlock` simply gives logs to the latest block. (It doesn't error.) I suppose that in theory this means we might be able to simplify some of the logic here, but given that we're subscribing before catching up (see [here](https://github.com/ava-labs/awm-relayer/blob/main/relayer/relayer.go#L121-L130)), I'd suggest that we leave that clamping logic in place, so that we don't doubly queue any blocks.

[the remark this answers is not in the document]
feuGeneA: I played around with some public ethereum nodes and found that calling `eth_getLogs` with a higher-than-latest `toBlock` simply gives logs to the latest block. (It doesn't error.) I suppose that in theory this means we might be able to simplify some of the logic here, but given that we're subscribing before catching up (see [here](https://github.com/ava-labs/awm-relayer/blob/main/relayer/relayer.go#L121-L130)), I'd suggest that we leave that clamping logic in place, so that we don't doubly queue any blocks.

vms/evm/subscriber.go
feuGeneA: @michaelkaplan13 thank you for pushing this further. I took your suggestion in https://github.com/ava-labs/awm-relayer/pull/89/commits/20d14f5d376b3acd141b36f0e649b89c99eb37af

@geoff-vball no because this is just with regards to iterating the `fromBlock` param from one request to the next, not the `toBlock` param

vms/evm/subscriber.go
```diff
@@ -154,25 +159,33 @@ func (s *subscriber) ProcessFromHeight(height *big.Int) error {
...
-			zap.String("chainID", s.chainID.String()),
-		)
-		height = big.NewInt(0).Add(toBlock, big.NewInt(-MaxBlocksToProcess))
+	bigLatestBlock := big.NewInt(0).SetUint64(latestBlock)
+
+	for fromBlock := big.NewInt(0).Set(height); fromBlock.Cmp(bigLatestBlock) <= 0; fromBlock.Add(fromBlock, big.NewInt(MaxBlocksPerRequest+1)) {
+		toBlock := big.NewInt(0).Add(fromBlock, big.NewInt(MaxBlocksPerRequest))
+		if toBlock.Cmp(bigLatestBlock) > 0 {
+			toBlock.Set(bigLatestBlock)
+		}
```
cam-schultz, line 168: Based on your findings discussed in https://github.com/ava-labs/awm-relayer/pull/89/files#r1402587975, it sounds this flooring is not strictly necessary. I understand the reason for keeping it, but can you leave a comment explaining so?

vms/evm/subscriber.go
```diff
@@ -154,25 +159,33 @@ func (s *subscriber) ProcessFromHeight(height *big.Int) error {
...
-			fmt.Sprintf("Requested to process too many blocks. Processing only the most recent %d blocks", MaxBlocksToProcess),
-			zap.String("requestedBlockHeight", height.String()),
-			zap.String("latestBlockHeight", toBlock.String()),
-			zap.String("chainID", s.chainID.String()),
-		)
-		height = big.NewInt(0).Add(toBlock, big.NewInt(-MaxBlocksToProcess))
+	bigLatestBlock := big.NewInt(0).SetUint64(latestBlock)
+
+	for fromBlock := big.NewInt(0).Set(height); fromBlock.Cmp(bigLatestBlock) <= 0; fromBlock.Add(fromBlock, big.NewInt(MaxBlocksPerRequest+1)) {
+		toBlock := big.NewInt(0).Add(fromBlock, big.NewInt(MaxBlocksPerRequest))
```
cam-schultz, line 165: I think there's an off-by-one-error here. Say we have `MaxBlocksPerRequest = 20`, and `fromBlock = 0`. Then `toBlock` would equal `20`, and since `eth_getLogs` takes an inclusive block range, we'd process `21` blocks instead of `20`. The fix here is to use `MaxBlocksPerRequest-1` on this line, and `MaxBlocksPerRequest` in the loop condition.

cam-schultz: (This is assuming `eth_getLogs` does in fact take an inclusive block range)

michaelkaplan13: `eth_getLogs` is inclusive on both side, great call out. I don't think it makes a _huge_ difference either way, but it is odd if `MaxBlockPerRequest` isn't the true effective value.

vms/evm/subscriber.go
michaelkaplan13: I think @geoff-vball is correct here actually. @cam-schultz also called it out and suggested the fix in this comment https://github.com/ava-labs/awm-relayer/pull/89/files?diff=split&w=1#r1402690097

vms/evm/subscriber_test.go
```diff
@@ -0,0 +1,131 @@
...
+		"awm-relayer-test",
+		logging.NewWrappedCore(
+			logging.Info,
+			os.Stdout,
+			logging.JSON.ConsoleEncoder(),
+		),
+	)
+
+	subnetId, err := ids.FromString(sourceSubnet.ChainID)
+	if err != nil {
```
minghinmatthewlam, line 42: lets use `require.NoError` with error message instead of `t.Fatalf`

vms/evm/subscriber.go
minghinmatthewlam: should we set `toBlock := fromBlock + maxBlockPerRequest - 1`? Since right now I think we'd be processing maxBlockPerRequest + 1

vms/evm/subscriber.go
```diff
@@ -154,25 +159,33 @@ func (s *subscriber) ProcessFromHeight(height *big.Int) error {
...
-	toBlock := big.NewInt(0).SetUint64(latestBlock)
-	if height.Cmp(big.NewInt(0).Sub(toBlock, big.NewInt(MaxBlocksToProcess))) < 0 {
-		s.logger.Warn(
-			fmt.Sprintf("Requested to process too many blocks. Processing only the most recent %d blocks", MaxBlocksToProcess),
-			zap.String("requestedBlockHeight", height.String()),
-			zap.String("latestBlockHeight", toBlock.String()),
-			zap.String("chainID", s.chainID.String()),
-		)
-		height = big.NewInt(0).Add(toBlock, big.NewInt(-MaxBlocksToProcess))
+	bigLatestBlock := big.NewInt(0).SetUint64(latestBlock)
```
minghinmatthewlam, line 162: nit: can we rename to `latestBlockHeight`

vms/evm/subscriber.go
feuGeneA: addressed in https://github.com/ava-labs/awm-relayer/pull/89/commits/f967bdd8c80e1b06cba033330fd2f9a358513665

vms/evm/subscriber.go
feuGeneA: Thanks guys, really appreciate the scrutiny here. Addressed in https://github.com/ava-labs/awm-relayer/pull/89/commits/a59dd14cecbf9e05463522ceb1b775c537fc0128

vms/evm/subscriber_test.go
feuGeneA: thanks, addressed in https://github.com/ava-labs/awm-relayer/pull/89/commits/75c965b79d0cb866b61ac1757909f7da700ab936

vms/evm/subscriber.go
feuGeneA: addressed in https://github.com/ava-labs/awm-relayer/pull/89/commits/64715d642dc9280e5ef15c0631cb1f7d8d96cf1a

vms/evm/subscriber.go
feuGeneA: thanks for connecting the dots here; conversation continued over there (https://github.com/ava-labs/awm-relayer/pull/89#discussion_r1408132909)

geoff-vball (approved):
LGTM!

cam-schultz (approved):
LGTM!

michaelkaplan13 (approved):

PR #99, opened 2023-12-04
Single source of truth for go version

geoff-vball:
## Why this should be merged
Creates a single source of truth for pulling the golang version, as well as deps like ginkgo and subnet-evm.

## How is this documented

### DIFF
--- a/scripts/versions.sh
+++ b/scripts/versions.sh
@@ -2,9 +2,24 @@
 # Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
 # See the file LICENSE for licensing terms.
 
-# Set up the versions to be used
-AWM_RELAYER_VERSION=${AWM_RELAYER_VERSION:-'v0.2.2'}
-SUBNET_EVM_VERSION=${SUBNET_EVM_VERSION:-'v0.5.8'}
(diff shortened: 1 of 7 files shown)


scripts/versions.sh
```diff
@@ -2,9 +2,24 @@
 # Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
 # See the file LICENSE for licensing terms.
 
-# Set up the versions to be used
-AWM_RELAYER_VERSION=${AWM_RELAYER_VERSION:-'v0.2.2'}
-SUBNET_EVM_VERSION=${SUBNET_EVM_VERSION:-'v0.5.8'}
+# The go version for this project is set from a combination of major.minor from go.mod and the patch version set here.
+GO_PATCH_VERSION=10
```
minghinmatthewlam, line 6: why do we set a go patch version of 10 here?

geoff-vball: Could you clarify what you mean?

cam-schultz (approved):

scripts/versions.sh
minghinmatthewlam: nvm I misunderstood it

minghinmatthewlam (approved):

PR #108, opened 2023-12-12
Set go version properly in workflows

geoff-vball:
## Why this should be merged
`GO_VERSION=$GO_VERSION >> $GITHUB_ENV` is not a proper bash command, so was failing, but silently for some reason. This was causing the env var not to be set, so the `go-version` in `actions/setup-go@v5` was defaulting to `1.20.11`

## How is this documented

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -17,12 +17,6 @@ jobs:
     runs-on: ubuntu-20.04
 
     steps:
-      - name: Checkout subnet-evm repository
-        uses: actions/checkout@v4
-        with:
-          repository: ava-labs/subnet-evm
-          ref: v0.5.9
-
       - name: Set Go version
         run: |
           source ./scripts/versions.sh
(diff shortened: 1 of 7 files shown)


cam-schultz (dismissed):

.github/workflows/e2e.yml
```diff
@@ -17,22 +17,26 @@ jobs:
...
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
           go-version: ${{ env.GO_VERSION }}
 
+      - name: Checkout subnet-evm repository
+        uses: actions/checkout@v4
+        with:
+          repository: ava-labs/subnet-evm
+          ref: v0.5.8 # TODO set with env.SUBNET_EVM_VERSION
```
michaelkaplan13, line 38: Is there a reason we can't do this TODO now?

geoff-vball: There's a breaking change, not exactly sure what it is, but created a ticket to investigate. https://github.com/ava-labs/awm-relayer/issues/109

minghinmatthewlam (approved):

michaelkaplan13 (approved):

PR #114, opened 2023-12-15
Catch up block height config option

### LINKED ISSUE #105: Specify catch up block height in config
**Context and scope**
The catch up mechanism reads from the JSON database to get the starting block from which to begin processing. This starting block is set automatically by the relayer as blocks are processed during runtime. There's no easy way for the user to specify the starting block; we should add a config option to do so.

**Open questions**
Should we limit the range of blocks that we process in one go? We don't currently do this in the automated approach, but this assumes a finite range since it's based on relayer downtime. Here, the user can specify an arbitrarily large range.

How should we "clean up" the config once we've already processed from the configured height? One idea would be to only process from the maximum of the configured block height and the stored block height, and then store the block height whenever we process a block (which we already do). The behavior in the following scenarios would be:
1. No block height is stored in the DB, configured block height is 100, and the current block height is 1000
  - We'd process blocks 100-1000 on startup, and store 1000 in the DB
2. Block height of 99 is stored in the DB, configured block height is 100, and the current block height is 1000
  - We'd process blocks 100-1000 on startup, and store 1000 in the DB
3. Block height of 101 is stored in the DB, configured block height is 100, and the current block height is 1000
  - We'd process blocks 101-1000 on startup, and store 1000 in the DB

All the above scenarios would be equavalent to scenario 3 after initialization & restart.

cam-schultz:
## Why this should be merged
Fixes #105

## How this works
Adds a `CatchUpBlockHeight` config option. On startup, the relayer will process from the maximum of the latest processed block height in the database, and the configured `CatchUpBlockHeight`. If neither value exists, then no catch up will occur.

## How this was tested
TODO

## How is this documented
README


tests/basic_relay.go
```diff
@@ -248,15 +242,61 @@ func BasicRelay(network interfaces.LocalNetwork) {
...
 
-func relayBasicMessage(
-	ctx context.Context,
+func sendBasicTeleporterMessage(ctx context.Context,
 	source interfaces.SubnetTestInfo,
 	destination interfaces.SubnetTestInfo,
 	fundedKey *ecdsa.PrivateKey,
-	fundedAddress common.Address,
-) {
+	fundedAddress common.Address) (teleportermessenger.TeleporterMessage, *big.Int) {
```
geoff-vball, line 299: nit: Can we keep the previous formatting, with the `ctx` on a newline and also the return types on a newline

README.md
```diff
@@ -83,6 +83,9 @@ The relayer is configured via a JSON file, the path to which is passed in via th
   `"supported-destinations": []string` 
   - List of destination subnet IDs that the source subnet supports. If empty, then all destinations are supported.
 
+  `"catch-up-block-height": integer`
+  - The block height at which to back-process transactions from the source subnet. If the database already contains a later block height for the source subnet, then that will be used instead.
+
```
geoff-vball, line 88: I've gone back and forth a bit on whether I think specifying a height of a max number of blocks to process would be better here. I think for release, being able to set the height of the warp upgrade makes the most sense, but in the future, setting a max number of blocks might make more sense.

geoff-vball (approved):
One nit and one thing to think about. Overall I think this is great

relayer/relayer.go
```diff
@@ -160,27 +161,47 @@ func NewRelayer(
...
+		// Use the max of the latest processed block and the catch up block height
 		latestProcessedBlock, success := new(big.Int).SetString(string(latestProcessedBlockData), 10)
 		if !success {
 			r.logger.Error("failed to convert latest block to big.Int", zap.Error(err))
 			return err
 		}
+		height = utils.MaxBigInt(latestProcessedBlock, catchUpBlockHeight)
+	} else if errors.Is(err, database.ErrChainNotFound) || errors.Is(err, database.ErrKeyNotFound) {
+		// If the database does not contain the latest processed block data, then we check if a catch up block height is provided.
+		if catchUpBlockHeight != nil {
```
michaelkaplan13, line 189: It looks like `catchUpBlockHeight` will never be `nil` since we set it using `big.NewInt(0).SetUint64(sourceSubnetInfo.CatchUpBlockHeight)`. Is there a way to distinguish a block height of 0 versus the value being unset?

config/config.go
```diff
@@ -47,6 +47,7 @@ type SourceSubnet struct {
 	WSEndpoint            string                           `mapstructure:"ws-endpoint" json:"ws-endpoint"`
 	MessageContracts      map[string]MessageProtocolConfig `mapstructure:"message-contracts" json:"message-contracts"`
 	SupportedDestinations []string                         `mapstructure:"supported-destinations" json:"supported-destinations"`
+	CatchUpBlockHeight    uint64                           `mapstructure:"catch-up-block-height" json:"catch-up-block-height"`
```
michaelkaplan13, line 50: I'm a bit worried we're providing too many options around the start block that the relayer will use on each initialization.

For instance, I think this is a weird possible case:
1. New relayer is started with no catch up block height and no DB. Uses current block and writes it to disk.
2. Now user realizes they wanted to relay a transaction from before starting it, so they specify a catch up block height and restart. 

In this case, the catch up block height has no effect because it was run without a DB or initial setting. 

I'd be in favor of simplifying the flow by removing the final case of "if there is no stored or configured start block, start from the tip of the chain", and renaming `CatchUpBlockHeight` to `StartBlockHeight`. If neither is present, we could print a warning log saying that the relayer will relay from the start of the chain, and to set the `StartBlockHeight` value if this is not the desired behavior.

cam-schultz: Is your suggestion to have `StartBlockHeight` potentially re-process blocks if the provided value is lower than the height stored in the DB?

I think that's reasonable, since we check that the message isn't already marked as delivered in the Teleporter contract before sending a tx.

cam-schultz: Discussed offline, and the approach we're going to go with is:
- When intializing a relayer for the first time (empty DB), if `StartBlockHeight` is provided, start from there, otherwise start from block 0 and emit a warning
- Otherwise, choose between the max of the `StartBlockHeight` and the block height in the DB

If the user wishes to start from a specific block earlier than the one stored in the DB, then the workaround would be to delete that entry from the DB and provide the desired block as the `StartBlockHeight` argument.

In the future, we'll add a config option to specify Warp message IDs to relay, which will provide the user fine-grained control for catching up on missed messages that were not automatically processed, without having to manually modify the DB. Issue here: https://github.com/ava-labs/awm-relayer/issues/126

README.md
cam-schultz: That's an interesting idea. Especially since processing from height 0 is perfectly valid, specifying a limit (and having a default value) would be helpful. Created an issue to track this: https://github.com/ava-labs/awm-relayer/issues/127

README.md
```diff
@@ -83,6 +83,9 @@ The relayer is configured via a JSON file, the path to which is passed in via th
   `"supported-destinations": []string` 
   - List of destination subnet IDs that the source subnet supports. If empty, then all destinations are supported.
 
+  `"catch-up-block-height": integer`
+  - The block height at which to back-process transactions from the source subnet. If the database already contains a later block height for the source subnet, then that will be used instead.
```
michaelkaplan13, line 87:
```suggestion
  `"start-block-height": unsigned integer`
  - The block height at which to begin checking for Warp message to process from the source subnet. If the database already contains a later block height for the source subnet, then that will be used instead.
```

main/main.go
```diff
@@ -96,7 +96,12 @@ func main() {
 		panic(err)
 	}
 
-	network, responseChans, err := peers.NewNetwork(logger, registerer, cfg.NetworkID, sourceSubnetIDs, sourceBlockchainIDs, cfg.PChainAPIURL)
+	// Unless the log level is debug, set the network log level to error
+	networkLogLevel := logging.Error
+	if logLevel == logging.Debug {
+		networkLogLevel = logging.Debug
+	}
```
michaelkaplan13, line 103: Why wouldn't we use the same log level in all cases here? (Could you add as a comment either way)

relayer/relayer.go
```diff
@@ -160,27 +162,53 @@ func NewRelayer(
...
 
-	// The following cases are treated as successful:
+	// First, determine the height to process from. There are two cases:
 	// 1) The database contains the latest processed block data for the chain
-	//    - In this case, we parse the block height and process warp logs from that height to the current block
+	//    - In this case, we process from the maximum of the latest processed block and the configured start block height to the latest block
 	// 2) The database has been configured for the chain, but does not contain the latest processed block data
-	//    - In this case, we save the current block height in the database, but do not process any historical warp logs
+	//    - In this case, if a start block height is provided, we process from the start block height to the latest block
+	//    - Otherwise, we save the current block height in the database, but do not process any historical warp logs
```
michaelkaplan13, line 178: I think this comment needs updating

relayer/relayer.go
```diff
@@ -189,9 +217,8 @@ func (r *Relayer) processMissedBlocks(
 			)
 		}
 		return nil
-	}
-	if errors.Is(err, database.ErrChainNotFound) || errors.Is(err, database.ErrKeyNotFound) {
-		// Otherwise, latestProcessedBlock is nil, so we instead store the latest block height.
+	} else {
+		// Otherwise, do not process any historical blocks, and store the latest block height
```
michaelkaplan13, line 221: I thought in this case we would log a warning and process starting from block 0?

cam-schultz: The way this function is called, we should never hit this case since `startBlockHeight` will never be `nil`.

Looking at it again though, it'd probably be cleaner to have this function require `startBlockHeight` to be non-nil, and remove this branch altogether.

main/main.go
cam-schultz: The p2p network generates logs that aren't necessarily pertinent to the core functionality of the relayer. In order to reduce the amount of log noise, we provide different log levels for each. Added a comment explaining this

relayer/relayer.go
```diff
@@ -136,7 +136,9 @@ func NewRelayer(
 	}
 
 	if shouldProcessMissedBlocks {
-		err = r.processMissedBlocks(sub)
+		// If StartBlockHeight is not set in the config, then we process from height 0
+		startHeight := big.NewInt(0).SetUint64(sourceSubnetInfo.StartBlockHeight)
+		err = r.processMissedBlocks(sub, startHeight)
```
geoff-vball, line 141: Would it make sense to to look up the configs/upgrades for this chain, and only start from the first block where warp was enabled?

geoff-vball (commented):
Just one comment about where we start indexing from. Otherwise LGTM

main/main.go
```diff
@@ -96,7 +92,17 @@ func main() {
...
 
-	network, responseChans, err := peers.NewNetwork(logger, registerer, cfg.NetworkID, sourceSubnetIDs, sourceBlockchainIDs, cfg.PChainAPIURL)
+	// Initialize the global app request network
+	logger.Info("Initializing app request network")
+	sourceSubnetIDs, sourceBlockchainIDs := cfg.GetSourceIDs()
+
+	// The app request network generates P2P networking logs that are verbose at the info level.
+	// Unless the log level is debug, set the network log level to error to avoid spamming the logs.
+	networkLogLevel := logging.Error
+	if logLevel == logging.Debug {
```
minghinmatthewlam, line 102: The `Verbo` log level is even lower than Debug, should the comparison be `<=`? Also how about log level `Trace`?

peers/app_request_network.go
```diff
@@ -41,13 +42,22 @@ type AppRequestNetwork struct {
...
-	logger logging.Logger,
+	logLevel logging.Level,
 	registerer prometheus.Registerer,
 	networkID uint32,
 	subnetIDs []ids.ID,
 	blockchainIDs []ids.ID,
 	APINodeURL string,
 ) (*AppRequestNetwork, map[ids.ID]chan message.InboundMessage, error) {
+	logger := logging.NewLogger(
+		"awm-relayer",
```
minghinmatthewlam, line 53: should we differentiate that this is the network layer, and not main app

relayer/relayer.go
minghinmatthewlam: not sure how easy/hard it is to find that block, but seems to make sense to me.

relayer/relayer.go
```diff
@@ -158,63 +160,61 @@ func NewRelayer(
...
-		if err != nil {
-			r.logger.Warn(
-				"Encountered an error when processing historical blocks. Continuing to normal relaying operation.",
-				zap.String("blockchainID", r.sourceBlockchainID.String()),
-				zap.Error(err),
-			)
-		}
-		return nil
+	// Process from the determined height to the latest block.
+	if height.Cmp(big.NewInt(0)) == 0 {
```
minghinmatthewlam, line 203: If teh last processed block height is not in db, should we require that the config starting block height is set, instead of automatically starting from 0 and sending a warning?

relayer/relayer.go
```diff
@@ -158,63 +160,61 @@ func NewRelayer(
...
-		// Note that the retrieved latest processed block may have already been partially (or fully) processed by the relayer on a previous run. When
-		// processing a warp message in real time, which is when we update the latest processed block in the database, we have no way of knowing
-		// if that is the last warp message in the block
+		// Use the max of the latest processed block and the start block height
 		latestProcessedBlock, success := new(big.Int).SetString(string(latestProcessedBlockData), 10)
 		if !success {
 			r.logger.Error("failed to convert latest block to big.Int", zap.Error(err))
 			return err
 		}
+		height = utils.MaxBigInt(latestProcessedBlock, height)
```
minghinmatthewlam, line 191: can we add a log here about using latest process block height from db

relayer/relayer.go
```diff
@@ -158,63 +160,61 @@ func NewRelayer(
...
+	} else if !errors.Is(err, database.ErrChainNotFound) && !errors.Is(err, database.ErrKeyNotFound) {
+		// Otherwise, we've encountered an unknown database error
+		r.logger.Warn(
+			"failed to get latest block from database",
+			zap.String("blockchainID", r.sourceBlockchainID.String()),
+			zap.Error(err),
+		)
+		return err
+	}
```
minghinmatthewlam, line 201: can you add a log here for `if err != nil` that relayer is using starting block height

relayer/relayer.go
cam-schultz: See the discussion here: https://github.com/ava-labs/awm-relayer/pull/114#discussion_r1443050931

relayer/relayer.go
cam-schultz: Do you feel strongly about logging the case when we pull the height from the DB specifically? We already log the height that we're processing from [here](https://github.com/ava-labs/awm-relayer/blob/008e978175449f32f24b35b8c6430e11e7365dd0/vms/evm/subscriber.go#L139), and to add the logging as described we wouldn't be able to use `utils.MaxBigInt`

relayer/relayer.go
cam-schultz: I think that's already covered in the logs [here](https://github.com/ava-labs/awm-relayer/blob/008e978175449f32f24b35b8c6430e11e7365dd0/vms/evm/subscriber.go#L139)

peers/app_request_network.go
cam-schultz: Yup, good call

main/main.go
cam-schultz: Good call, fixed

relayer/relayer.go
minghinmatthewlam: I mean a log that tells us we're using the block ehgiht from the config setting `startBlockheight` and not the latest db value

relayer/relayer.go
minghinmatthewlam: Same as below, I meant a log that just tells user we are using latest value from db instead of starting block height config value.

minghinmatthewlam: Can be debug too, whichever you think is mroe appropriate

cam-schultz: Gotcha. Added.

relayer/relayer.go
cam-schultz: Depends on https://github.com/ava-labs/awm-relayer/pull/124, so created a separate issue to track here: https://github.com/ava-labs/awm-relayer/issues/129

relayer/relayer.go
minghinmatthewlam: yup I agree with a majority on that approach, but didn't see the reason for starting from 0 if both are not set, probably was discussed offline.

cam-schultz: Thinking about this again, your suggestion might be a better UI. When starting a new relayer for the first time, I think it's more likely that a missing `StartBlockHeight` value is user error rather than intentionally starting processing blocks from genesis. If the user does wish to process from genesis, they can always set `StartBlockHeight` to 0 explicitly. @michaelkaplan13 curious what your thoughts are in case I'm missing something, otherwise I'm going to take @minghinmatthewlam's suggestion.

cam-schultz: Implemented this in a separate PR for ease of review: https://github.com/ava-labs/awm-relayer/pull/133

michaelkaplan13: Yep I agree with @minghinmatthewlam's suggestion here.

cam-schultz:
Closing in favor of https://github.com/ava-labs/awm-relayer/pull/133

Issue #8, opened 2023-08-28
Get Quorum numerator/denominator from VM


minghinmatthewlam:
**Context and scope**
Warp quorum numerator and denominators are set to decide what signature threshold needs to be exceeded to count as a valid warp message.

The quorum numerator and denominator are provided as a warp precompile config in the genesis. We should use those values rather than the ones hardcoded in the relayer.

https://github.com/ava-labs/awm-relayer/blob/main/utils/utils.go#L18

cam-schultz:
We can get the Warp numerator specified in genesis using `eth_getChainConfig` and parsing `warpConfig.quorumNumerator`. The Warp denominator is fixed to 100 for `subnet-evm` and `coreth`, but for robustness we may want to attempt to parse that value as well and default to 100 if the value doesn't exist.

PR #115, opened 2023-12-18
C-Chain compatibility

### LINKED ISSUE #113: Implement C-Chain <> subnet integration test
Context and scope
As of Avalanchego 1.10.17, Warp is activated by default on local networks. There is therefore no special setup required to author integration tests that send Teleporter messages between the C-Chain and subnets.

Discussion and alternatives
https://github.com/ava-labs/teleporter/issues/193 already covers the test cases on the Teleporter side, so the changes involved here will be to extend the `basic_relay` test to relay to/from the C-Chain in addtion to between two subnets.

### LINKED ISSUE #117: Support AppRequests for the C-Chain
**Context and scope**
The relayer gathers signatures from validators directly via AppRequest. This currently only works for subnet-evm, and we will need to get it working for coreth as well in order to gather signatures from the C-Chain. In the meantime, the Warp client provides an endpoint to get the aggregate signature for both subnet-evm chains as well as the C-Chain, but this is not as generalizable as the AppRequest approach.

cam-schultz:
## Why this should be merged
With Avalanchego v1.10.17, Warp is enabled by default on the C-Chain for local networks, opening a path to easily test Teleporter and AWM Relayer. This PR adds E2E test cases to test relaying between the C-Chain and a subnet, as well as modifications to the relayer signature aggregation logic.

Fixes #113
Fixes #117 

## How this works
- Extends the Basic Relay E2E test case to use the C-Chain/Primary Network as one of the subnets, and relay a message in each direction between the C-Chain and a subnet
- ~~Modifies the internal signature aggregation logic to use the Warp client included in `subnet-evm` and `coreth`, rather than the custom AppRequest logic. There are further changes needed to get the AppRequest approach working with the C-Chain, meanwhile the Warp client works out-of-the-box (issue here: https://github.com/ava-labs/awm-relayer/issues/117)~~
- EDIT: AppRequests to the C-Chain are working in this PR. I'll leave in the logic to use the Warp API in place, and track adding a config option to support that here: https://github.com/ava-labs/awm-relayer/issues/119

## How this was tested
Extended E2E tests.

## How is this documented
N/A


utils/utils.go
```diff
@@ -97,3 +93,16 @@ func SanitizeHexString(hex string) string {
 	}
 	return hex
 }
+
+func StripFromString(input, substring string) string {
```
geoff-vball, line 97: Leave a comment here that it removes the first occurance of the substring AND anything after it.

utils/utils.go
```diff
@@ -8,16 +8,12 @@ import (
...
 
 	"github.com/ethereum/go-ethereum/common"
 )
 
 const (
-	// Default stake threshold for aggregate signature verification. (67%)
-	// TODO: This should be made configuration on the VM level.
-	DefaultQuorumNumerator   = 67
-	DefaultQuorumDenominator = 100
-
```
geoff-vball, line 20: Do we need to complete https://github.com/ava-labs/awm-relayer/issues/8 for release?

cam-schultz: Good call. I'll update that issue now.

relayer/message_relayer.go
```diff
@@ -125,8 +144,79 @@ func (r *messageRelayer) relayMessage(warpMessageInfo *vmtypes.WarpMessageInfo,
...
+	if r.relayer.sourceSubnetID == constants.PrimaryNetworkID {
+		signingSubnetID, err = r.relayer.pChainClient.ValidatedBy(context.Background(), r.destinationBlockchainID)
+		if err != nil {
+			r.logger.Error(
+				"failed to get validating subnet for destination chain",
+				zap.String("destinationBlockchainID", r.destinationBlockchainID.String()),
+				zap.Error(err),
+			)
+			return nil, err
+		}
```
michaelkaplan13, line 167: Could this be simplified just to:
```
signingSubnetID = r.destinationBlockchainID
```

I'm guessing not, but I'm not sure why we'd need to make an RPC call to the P-chain to determine the blockchain ID of the signer set.

relayer/message_relayer.go
```diff
@@ -125,8 +144,79 @@ func (r *messageRelayer) relayMessage(warpMessageInfo *vmtypes.WarpMessageInfo,
 	return nil
 }
 
-// Run collects signatures from nodes by directly querying them via AppRequest, then aggregates the signatures, and constructs the signed warp message.
-func (r *messageRelayer) createSignedMessage(requestID uint32) (*warp.Message, error) {
+func (r *messageRelayer) createSignedMessage() (*warp.Message, error) {
+	r.logger.Info("Fetching aggregate signature from the source chain validators")
```
michaelkaplan13, line 148: Can we this log to make it clear its using the API endpoint?

relayer/message_relayer.go
```diff
@@ -125,8 +144,79 @@ func (r *messageRelayer) relayMessage(warpMessageInfo *vmtypes.WarpMessageInfo,
...
+				"failed to get validating subnet for destination chain",
+				zap.String("destinationBlockchainID", r.destinationBlockchainID.String()),
+				zap.Error(err),
+			)
+			return nil, err
+		}
+	}
+
+	var signedWarpMessageBytes []byte
+	for attempt := 1; attempt <= maxRelayerQueryAttempts; attempt++ {
```
michaelkaplan13, line 171:
```suggestion
	for attempt := 0; attempt < maxRelayerQueryAttempts; attempt++ {
```

Minor nit for the other folks who haven't worked at MathWorks 😅 


** Edit: Ah nevermind I see why you did this to make the "last iteration" check simpler below. I think in the past we've typically done -1 on the last iteration check rather than 1 indexing to start with, but no strong preference.

michaelkaplan13: I would consider creating a "request with retries" `utils` method either way so that we can re-use this logic though.

relayer/message_relayer.go
```diff
@@ -125,8 +144,79 @@ func (r *messageRelayer) relayMessage(warpMessageInfo *vmtypes.WarpMessageInfo,
...
+			)
+		}
+		if attempt != maxRelayerQueryAttempts {
+			// Sleep such that all retries are uniformly spread across totalRelayerQueryPeriodMs
+			// TODO: We may want to consider an exponential back off rather than a uniform sleep period.
+			time.Sleep(time.Duration(signatureRequestRetryWaitPeriodMs/maxRelayerQueryAttempts) * time.Millisecond)
+		}
+	}
+	r.logger.Warn(
+		"Failed to collect a threshold of signatures",
```
michaelkaplan13, line 209:
```suggestion
		"Failed to get aggregate signature from node endpoint",
```

relayer/message_relayer.go
```diff
@@ -125,8 +144,79 @@ func (r *messageRelayer) relayMessage(warpMessageInfo *vmtypes.WarpMessageInfo,
...
 }
 
-// Run collects signatures from nodes by directly querying them via AppRequest, then aggregates the signatures, and constructs the signed warp message.
-func (r *messageRelayer) createSignedMessage(requestID uint32) (*warp.Message, error) {
+func (r *messageRelayer) createSignedMessage() (*warp.Message, error) {
+	r.logger.Info("Fetching aggregate signature from the source chain validators")
+	warpClient, err := warpBackend.NewClient(r.relayer.apiNodeURI, r.relayer.sourceBlockchainID.String())
+	if err != nil {
+		r.logger.Error(
+			"Failed to create warp client",
```
minghinmatthewlam, line 152: can we add the api used

relayer/message_relayer.go
```diff
@@ -125,8 +144,79 @@ func (r *messageRelayer) relayMessage(warpMessageInfo *vmtypes.WarpMessageInfo,
...
+			zap.Error(err),
+		)
+		return nil, err
+	}
+	signingSubnetID := r.relayer.sourceSubnetID
+	if r.relayer.sourceSubnetID == constants.PrimaryNetworkID {
+		signingSubnetID, err = r.relayer.pChainClient.ValidatedBy(context.Background(), r.destinationBlockchainID)
+		if err != nil {
+			r.logger.Error(
+				"failed to get validating subnet for destination chain",
```
minghinmatthewlam, line 162: nit: capitalize error msg here

relayer/message_relayer.go
```diff
@@ -125,8 +144,79 @@ func (r *messageRelayer) relayMessage(warpMessageInfo *vmtypes.WarpMessageInfo,
...
+			warpMsg, err := avalancheWarp.ParseMessage(signedWarpMessageBytes)
+			if err != nil {
+				r.logger.Error(
+					"Failed to parse signed warp message",
+					zap.Error(err),
+				)
+				return nil, err
+			}
+			return warpMsg, err
+		} else {
```
minghinmatthewlam, line 195: don't think we need this `else`

relayer/message_relayer.go
```diff
@@ -173,10 +263,20 @@ func (r *messageRelayer) createSignedMessage(requestID uint32) (*warp.Message, e
...
+	} else {
+		req := msg.MessageSignatureRequest{
+			MessageID: r.warpMessage.ID(),
+		}
+		reqBytes, err = msg.RequestToBytes(codec, req)
 	}
-	reqBytes, err := msg.RequestToBytes(codec, req)
 	if err != nil {
 		r.logger.Error(
 			"Failed to marshal request bytes",
```
minghinmatthewlam, line 282: nit: think might be better to log Warp message ID over destination blockchain ID

relayer/message_relayer.go
```diff
@@ -311,7 +411,7 @@ func (r *messageRelayer) createSignedMessage(requestID uint32) (*warp.Message, e
 					}
 
 					// As soon as the signatures exceed the stake weight threshold we try to aggregate and send the transaction.
-					if utils.CheckStakeWeightExceedsThreshold(accumulatedSignatureWeight, totalValidatorWeight, utils.DefaultQuorumNumerator, utils.DefaultQuorumDenominator) {
+					if utils.CheckStakeWeightExceedsThreshold(accumulatedSignatureWeight, totalValidatorWeight, params.WarpDefaultQuorumNumerator, params.WarpQuorumDenominator) {
```
minghinmatthewlam, line 414: should we by default be using the `coreth.params`? What if `subnet-evm` has different values, and the message we're sending doesn't interact with the primary network

relayer/message_relayer.go
cam-schultz: For sending from the primary network, we need to get the destination subnet ID, not the blockchainID, so that we can query the right set of validators. 

The message relayer has the following information:
- destination blockchain ID (from the warp message)
- source subnet ID (from the configuration)
- source blockchainID (also from the configuration)

So we need to get the destination subnet ID via the P-Chain

relayer/message_relayer.go
cam-schultz: Yeah, we should make that distinction. I added a comment to https://github.com/ava-labs/awm-relayer/issues/8 to fetch the numerator config from the VM, rather than using a hardcoded value.

relayer/message_relayer.go
cam-schultz: Good call. Removed.

geoff-vball (approved):

minghinmatthewlam (dismissed):

relayer/message_relayer.go
michaelkaplan13: Ah got it, thanks

geoff-vball (approved):

relayer/message_relayer.go
michaelkaplan13: Discussed offline, we'll defer the retry utility for now because it's more complicated than it's worth for the time being.

michaelkaplan13 (approved):

PR #124, opened 2024-01-04
Fetch warp numerator

cam-schultz:
## Why this should be merged
Fixes https://github.com/ava-labs/awm-relayer/issues/8

## How this works
Fetches the Warp Quorum from the Warp Config on startup. Hardcodes the quorum value for the primary network, since that will never change.
Depends on https://github.com/ava-labs/subnet-evm/pull/1040; CI will not pass until that is merged.

## How this was tested
Added unit tests

## How is this documented
N/A

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -251,7 +251,7 @@ func setQuorumNumerator(cfgNumerator uint64) uint64 {
 // Differentiates between subnet-evm and coreth RPC internally
 func getWarpQuorum(
 	subnetID ids.ID,
-	rpc string,
+	client ethclient.Client,
 ) (WarpQuorum, error) {
 	if subnetID == constants.PrimaryNetworkID {
 		return WarpQuorum{
@@ -260,10 +260,6 @@ func getWarpQuorum(
 		}, nil
 	} else {
 		// Fetch the subnet's chain config
-		client, err := ethclient.Dial(rpc)
-		if err != nil {
-			return WarpQuorum{}, fmt.Errorf("failed to dial subnet %s: %v", subnetID, err)
-		}
 		chainConfig, err := client.ChainConfig(context.Background())
 		if err != nil {
 			return WarpQuorum{}, fmt.Errorf("failed to fetch chain config for subnet %s: %v", subnetID, err)
@@ -279,7 +275,8 @@ func getWarpQuorum(
 		}
 
 		// If we didn't find the Warp config in the genesis precompile list, check the upgrade precompiles
-		for _, precompile := range chainConfig.UpgradeConfig.PrecompileUpgrades {
+		upgrades := chainConfig.ToWithUpgradesJSON().UpgradeConfig.PrecompileUpgrades
+		for _, precompile := range upgrades {
 			warpConfig, ok := precompile.Config.(*warp.Config)
 			if ok {
 				return WarpQuorum{
@@ -302,10 +299,15 @@ func (c *Config) initializeWarpQuorum() error {
 			return fmt.Errorf("invalid subnetID in configuration. error: %v", err)
 		}
 
-		quorum, err := getWarpQuorum(subnetID, sourceSubnet.GetNodeRPCEndpoint())
+		client, err := ethclient.Dial(sourceSubnet.GetNodeRPCEndpoint())
+		if err != nil {
+			return fmt.Errorf("failed to dial source subnet %s: %v", subnetID, err)
+		}
+		quorum, err := getWarpQuorum(subnetID, client)
 		if err != nil {
 			return err
 		}
+
 		c.warpQuorum[subnetID] = quorum
 	}
 
@@ -321,10 +323,15 @@ func (c *Config) initializeWarpQuorum() error {
 			continue
 		}
 
-		quorum, err := getWarpQuorum(subnetID, destinationSubnet.GetNodeRPCEndpoint())
+		client, err := ethclient.Dial(destinationSubnet.GetNodeRPCEndpoint())
+		if err != nil {
+			return fmt.Errorf("failed to dial destination subnet %s: %v", subnetID, err)
+		}
+		quorum, err := getWarpQuorum(subnetID, client)
 		if err != nil {
 			return err
 		}
+
 		c.warpQuorum[subnetID] = quorum
 	}
 	return nil
--- a/config/config.go
+++ b/config/config.go
@@ -239,6 +239,14 @@ func (c *Config) Validate() error {
 	return nil
 }
 
+// If the numerator in the Warp config is 0, use the default value
+func setQuorumNumerator(cfgNumerator uint64) uint64 {
+	if cfgNumerator == 0 {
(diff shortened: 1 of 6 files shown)


config/config.go
```diff
@@ -223,6 +239,104 @@ func (c *Config) Validate() error {
...
+func getWarpQuorum(
+	subnetID ids.ID,
+	client ethclient.Client,
+) (WarpQuorum, error) {
+	if subnetID == constants.PrimaryNetworkID {
+		return WarpQuorum{
+			QuorumNumerator:   params.WarpDefaultQuorumNumerator,
+			QuorumDenominator: params.WarpQuorumDenominator,
+		}, nil
+	} else {
```
geoff-vball, line 261: nit: you could get rid of the else clause here.

config/config.go
```diff
@@ -223,6 +239,104 @@ func (c *Config) Validate() error {
...
+		for _, precompile := range upgrades {
+			warpConfig, ok := precompile.Config.(*warp.Config)
+			if ok {
+				return WarpQuorum{
+					QuorumNumerator:   setQuorumNumerator(warpConfig.QuorumNumerator),
+					QuorumDenominator: params.WarpQuorumDenominator,
+				}, nil
+			}
+		}
+		return WarpQuorum{}, fmt.Errorf("failed to find warp config for subnet %s", subnetID)
```
geoff-vball, line 288: Is it possible for a subnet to change the warp quorum in an upgrade after it has already been set? If so, we would need to find the most recent value.

config/config.go
```diff
@@ -223,6 +239,104 @@ func (c *Config) Validate() error {
...
+	c.warpQuorum = make(map[ids.ID]WarpQuorum)
+
+	// Fetch the Warp quorum values for each source subnet
+	for _, sourceSubnet := range c.SourceSubnets {
+		subnetID, err := ids.FromString(sourceSubnet.SubnetID)
+		if err != nil {
+			return fmt.Errorf("invalid subnetID in configuration. error: %v", err)
+		}
+
+		client, err := ethclient.Dial(sourceSubnet.GetNodeRPCEndpoint())
```
geoff-vball, line 302: We should defer closing the client here (after checking the error) and at the dial below

config/config.go
cam-schultz: Good call. We should be able to account for this by checking the upgrade config first, then the genesis config if necessary.

geoff-vball: Yeah, we should also account for multiple upgrades having different values, if that's possible. This could also get a little hairy if we're processing historical blocks over an upgrade where this value might change, but that is probably a pretty rare edgecase.

config/config.go
```diff
@@ -223,6 +239,106 @@ func (c *Config) Validate() error {
...
+	upgrades := chainConfig.ToWithUpgradesJSON().UpgradeConfig.PrecompileUpgrades
+	for _, precompile := range upgrades {
+		warpConfig, ok := precompile.Config.(*warp.Config)
+		if ok {
+			return WarpQuorum{
+				QuorumNumerator:   setQuorumNumerator(warpConfig.QuorumNumerator),
+				QuorumDenominator: params.WarpQuorumDenominator,
+			}, nil
+		}
+	}
```
geoff-vball, line 279: Does this iterate from newest to oldest? If so, can you leave a comment here explaining that? We should make sure we get the newest warp config that exists.

cam-schultz: Added a check to keep only the most recent Warp config (if we find one at all in the upgrades config)

config/config.go
geoff-vball: Ah okay great, that helps a lot. Could you leave a comment in here somewhere explaining that?

config/config.go
```diff
@@ -223,6 +239,124 @@ func (c *Config) Validate() error {
...
+		return WarpQuorum{}, fmt.Errorf("failed to fetch chain config for blockchain %s: %v", blockchainID, err)
+	}
+
+	// First, check the list of precompile upgrades to get the most up to date Warp config
+	var warpConfig *warp.Config
+	for _, precompile := range chainConfig.UpgradeConfig.PrecompileUpgrades {
+		cfg, ok := precompile.Config.(*warp.Config)
+		if ok {
+			if warpConfig == nil {
+				cfg = warpConfig
```
geoff-vball, line 276: I think you have this assignment backwards

config/config.go
cam-schultz: Sure thing, added

config/config.go
cam-schultz: D'oh. Fixed

geoff-vball (commented):
This all looks good to me at this point. Just one last question... If there's a network upgrade, will that be included in the block somehow? You mentioned we would never need an older warp config, but if there's a new warp config while the relayer is running, we should have a way of knowing that and grabbing it.

cam-schultz:
> This all looks good to me at this point. Just one last question... If there's a network upgrade, will that be included in the block somehow? You mentioned we would never need an older warp config, but if there's a new warp config while the relayer is running, we should have a way of knowing that and grabbing it.

This case is not explicitly handled (nor am I sure there's a way to determine this on the client side for certain), but in the case of a network upgrade, the API node would shut down, causing the relayer to eventually restart which would refresh the chain configs on re-initialization. The shutdown timer is determined by [maxResubscribeAttempts](https://github.com/ava-labs/awm-relayer/blob/008e978175449f32f24b35b8c6430e11e7365dd0/vms/evm/subscriber.go#L32); we should tune this to shut down on node restart, but to attempt to reconnect for shorter blips.

config/config.go
```diff
@@ -223,6 +239,127 @@ func (c *Config) Validate() error {
...
+// Helper to retrieve the Warp Quorum from the chain config.
+// Differentiates between subnet-evm and coreth RPC internally
+func getWarpQuorum(
+	blockchainID ids.ID,
+	subnetID ids.ID,
+	client ethclient.Client,
+) (WarpQuorum, error) {
+	if subnetID == constants.PrimaryNetworkID {
+		return WarpQuorum{
+			QuorumNumerator:   params.WarpDefaultQuorumNumerator,
```
minghinmatthewlam, line 259: why do we set to default quorum numerator for c-chain? Don't we still use it when aggregating signature

minghinmatthewlam: For a Warp message originating on the c-chain, when we aggregate signatures do we use the quorum numerator of the destination chain? Right now in `message_relayer` it looks like we still use the quorum numerator of the c-chain

cam-schultz:
Closing in favor of https://github.com/ava-labs/awm-relayer/pull/157

config/config.go
cam-schultz: > For a Warp message originating on the c-chain, when we aggregate signatures do we use the quorum numerator of the destination chain

Yes, we should always be using the destination's warp quorum to construct the aggregate signature. I think there were some commits here and in the latest PR https://github.com/ava-labs/awm-relayer/pull/157 in which that wasn't the case. Please let me know if you still see an issue.

PR #128, opened 2024-01-05
Integrate Teleporter message id changes

### LINKED ISSUE #123: Integrate message ID hash
**Context and scope**
https://github.com/ava-labs/teleporter/pull/209 updates the messageID field to consist of a hash of the TeleporterMessenger address, source blockchain ID, destination blockchain ID, and message nonce. We will need to integrate these changes into the relayer.

minghinmatthewlam:
## Why this should be merged
updates relayer to be compatible with Teleporter message id changes, fixes #123 

## How this works
We use the `abi-bindings` pkg from `teleporter` repo and the destination client to create a Teleporter instance. Then when relaying a Teleporter message we use the instance to calculate message id and check message received when deciding whether we should send a message. 

## How this was tested
Updated unit/e2e tests
CI

## How is this documented
Updated code comments


cam-schultz (dismissed):
The changes LGTM. Thanks for cleaning up the Teleporter contract interaction in message manager as well.

messages/teleporter/message_manager.go
```diff
@@ -323,3 +294,40 @@ func (m *messageManager) parseTeleporterMessage(warpMessageID ids.ID, warpPayloa
 	}
 	return teleporterMessage, nil
 }
+
+func (m *messageManager) getTeleporterMessenger(destinationClient vms.DestinationClient) (*teleportermessenger.TeleporterMessenger, error) {
+	client, ok := destinationClient.Client().(ethclient.Client)
```
geoff-vball, line 299: Does this casting work for the c-chain client?

messages/teleporter/message_manager.go
```diff
@@ -323,3 +294,40 @@ func (m *messageManager) parseTeleporterMessage(warpMessageID ids.ID, warpPayloa
...
+	if !ok {
+		m.logger.Error(
+			"Destination client is not an Ethereum client.",
+			zap.String("destinationBlockchainID", destinationClient.DestinationBlockchainID().String()),
+		)
+		return nil, errors.New("destination client is not an Ethereum client")
+	}
+
+	// Get the teleporter messenger contract
+	teleporterMessenger, err := teleportermessenger.NewTeleporterMessenger(common.BytesToAddress(m.protocolAddress[:]), client)
```
geoff-vball, line 309: Kind of unrelated to this PR, but does anyone know why `protocolAddress` is type `common.Hash` instead of `common.Address`? @cam-schultz

cam-schultz: Yeah, this is to support other VMs in the future that may not adhere to EVM 20 bytes addresses.

messages/teleporter/message_manager.go
cam-schultz: It should, since we initialize the C-Chain client using `subnet-evm`'s `ethclient`. This shouldn't be a problem since we're not using the client to fetch blocks, but we should probably rethink this.

geoff-vball: Yeah, for now I think `destinationClient.Client()` should return an `ethclient.Client`. At least this way when we need to add a new client type, we will know we'll need to make a change here.

messages/teleporter/message_manager.go
geoff-vball: Hmm... Seems like that is just adding complexity for now that won't really help us down the road. In order to support other address lengths, we'd need to rearchitect a lot of the repo. Just about everywhere else uses `common.Address` and our golang bindings are specific to the address length as well.

cam-schultz: I definitely agree. As evidenced by this [draft PR](https://github.com/ava-labs/awm-relayer/pull/54), accounting for message protocol changes doesn't fit super nicely into the existing architecture. Whenever we go about supporting the next non-EVM and non-Teleporter use case, we should reevaluate the architecture as well. Until then, I think we should not attempt to design interfaces in anticipation of those future use cases. In other words, let's not change the interfaces until we have a second concrete use case to guide that design.

minghinmatthewlam: Yup agree with yall's points, parts of repo are evm specific, other parts are not. Lets wait for an actual non-evm case to redesign the itnerfaces.

messages/teleporter/message_manager.go
minghinmatthewlam: same as below

messages/teleporter/message_manager.go
geoff-vball, line 309: So the call above to `destinationClient.Client().(ethclient.Client)` seems to me like a pretty big footgun. What I think we should do instead, is on this line we call `NewTeleporterMessengerCaller` instead of `NewTeleporterMessenger`. This gives a read-only version of the abi-binding, which only requires the client to have implemented `CodeAt` and `CallContract`. I propose we add those two functions to the `DestinationClient` interface, and that way we don't need to do any casting.

geoff-vball: > Until then, I think we should not attempt to design interfaces in anticipation of those future use cases. In other words, let's not change the interfaces until we have a second concrete use case to guide that design.

I totally agree with the first sentence here, but the conclusion that leads me to is not to "not change" the interfaces, but to make them no more general than is necessary. 

I think it's kind of dangerous to take a 20-byte address, pass it around as something else, only to convert it back to a 20-byte address later. It's also kind of dangerous to make the desinationClient pass back an interface, if in every case it's using an `ethclient.Client`. I think both of those premature abstractions only lead us to code in strange ways that may be more error-prone as we build around them. I think once the second use-case comes up, that will give the correct indication of how we should go about making the interfaces more general.

cam-schultz: Agreed, feel free to create an issue to track this. I'd just prefer to not expand the scope of this PR.

minghinmatthewlam: since the call `destinationClient.Client().(ethclient.Client)` was already here before, and we have a separate issue https://github.com/ava-labs/awm-relayer/issues/132 to address changing this call, don't think it's worth addressing in this PR, and I will defer it for the new issue.

minghinmatthewlam: New issue created by @geoff-vball  thanks https://github.com/ava-labs/awm-relayer/issues/132

geoff-vball: Yeah sure! Sorry, didn't notice this was already a part of the previous code.

messages/teleporter/message_manager.go
```diff
@@ -323,3 +294,40 @@ func (m *messageManager) parseTeleporterMessage(warpMessageID ids.ID, warpPayloa
...
+	if err != nil {
+		m.logger.Error(
+			"Failed to get teleporter messenger contract",
+			zap.Error(err),
+		)
+		return nil, err
+	}
```
geoff-vball, line 316: I think we should actually just panic in both of these error cases. They're both "impossible" cases. `teleportermessenger.NewTeleporterMessenger` doesn't actually call anything on the client, it just shoves it into a struct that can be used to call functions on the contract.

Right now, these errors will get bubbled up and logged and swallowed, but if either of these two calls ever fail, every call to them will fail every time.

minghinmatthewlam: that makes sense, updated.

geoff-vball: Great! You can also remove the `error` from the return types of the function now that it never returns an error

geoff-vball (approved):

messages/teleporter/message_manager.go
```diff
@@ -125,29 +124,31 @@ func (m *messageManager) ShouldSendMessage(warpMessageInfo *vmtypes.WarpMessageI
 		return false, fmt.Errorf("relayer not configured to deliver to destination. destinationBlockchainID=%s", destinationBlockchainID.String())
 	}
 
+	teleporterMessenger := m.getTeleporterMessenger(destinationClient)
+	teleporterMessageID, err := m.calculateMessageID(teleporterMessenger, warpMessageInfo.WarpUnsignedMessage.SourceChainID, destinationBlockchainID, teleporterMessage.MessageNonce)
```
michaelkaplan13, line 128: It a bit sub-optimal that getting the message ID requires an additional API call here, but given the current interfaces and utilities, it's I agree that it's the easiest option. 

Alternatively, the message ID is emitted as a part of the `SendCrossChainMessage` event and could be used directly from there (though we don't currently subscribe or otherwise get that event), or we could provide a Golang utility for the hash operation, which would be far more efficient than an API call but have a small amount of duplicated logic.

I don't think we need to change it for this PR, but wanted to call out to hear people's thoughts.

messages/teleporter/message_manager.go
```diff
@@ -237,19 +180,31 @@ func (m *messageManager) SendMessage(signedMessage *warp.Message, parsedVmPayloa
...
 	}
 
+	// Get the correct destination client from the global map
+	destinationClient, ok := m.destinationClients[destinationBlockchainID]
+	if !ok {
+		return fmt.Errorf("relayer not configured to deliver to destination. DestinationBlockchainID=%s", destinationBlockchainID)
+	}
+
+	teleporterMessenger := m.getTeleporterMessenger(destinationClient)
+	teleporterMessageID, err := m.calculateMessageID(teleporterMessenger, signedMessage.SourceChainID, destinationBlockchainID, teleporterMessage.MessageNonce)
```
michaelkaplan13, line 190: Making an API call here as well to get the message ID just for the sake of logging feels sub-optimal.

michaelkaplan13 (approved):
The changes LGTM, but would like to follow up about making API calls for `calculateMessageID`. Would be great to remove that in the future by either passing in the proper message ID, or calculating it based on the input parameters directly using keccack256

messages/teleporter/message_manager.go
geoff-vball: Yeah investigated this a bit and unfortunately there is no way to call any methods from the golang ABI without an RPC call, even if you're calling a `pure` function. I totally agree that needing the API here is very undesired.

I think the best solution here is probably to build our own Golang utility that reproduces the functionality. Importantly, we should add a test (or include the sanity check in the e2e network setup) that calls the Teleporter contract to make sure it aligns with our utility.

We can totally defer this to a new ticket, though.

geoff-vball: Also, the utility should be located in the `teleporter` repo.

minghinmatthewlam: created new issue to track: https://github.com/ava-labs/awm-relayer/issues/136

messages/teleporter/message_manager.go
minghinmatthewlam: agreed, will defer for https://github.com/ava-labs/awm-relayer/issues/136

messages/teleporter/message_manager.go
cam-schultz: Agreed on all points. In the future when optimizing end to end relayer latency, minimizing the number of network round trips will be an obvious starting point.

Issue #30, opened 2023-09-06
Implement failure recovery


cam-schultz:
**Context and scope**
Current top-level error handling is inconsistent - in some cases we exit the application, in other cases, we exit only the relayer goroutine for a particular chain. We should identify clear failure modes and appropriate recovery mechanisms for each, whether that be killing the application altogether, or restarting portions of the application internally.

cam-schultz:
We should review the relayer implementation and add automated recovery wherever possible. The current approach when _any_ top-level relayer goroutine fails is to either kill the application directly, or set the health check to fail, which will also effectively kill the application. Instead, we should attempt to recover from failures whenever we can. 

Our failure recovery strategy should broadly follow this order of events
1. Attempt to retry the failed operation
2. Emit errors and warnings, and then continue
3. If the error puts the relayer in a bad state, kill the top-level relayer goroutine

Whenever possible, we should attempt to retry the failed operation, and emit errors if that is not possible. The top-level relayer goroutine should fail only in unexpected scenarios.

For example, when accumulating signatures, we [retry](https://github.com/ava-labs/awm-relayer/blob/79dcc82e1cf2f480a03d67889adab3bf0113d7c9/relayer/message_relayer.go#L203) a set number of times, and if that fails, we emit corresponding errors and then [continue](https://github.com/ava-labs/awm-relayer/blob/b1d0aede4336b13b9425e1968a7318b12a9f6f7b/main/main.go#L252) to next message to relay. We do not intentionally exit the goroutine under any circumstances, however.

As another example, when the subscriber [errors](https://github.com/ava-labs/awm-relayer/blob/b1d0aede4336b13b9425e1968a7318b12a9f6f7b/main/main.go#L254), we attempt to re-establish the subscription, and exit the goroutine if we are unsuccessful.

PR #133, opened 2024-01-10
Exit if no start block is found

cam-schultz:
## Why this should be merged
Originally in https://github.com/ava-labs/awm-relayer/pull/114, we were planning on implementing the catch up mechanism such that if the starting block was not found in the database nor in the `StartBlockHeight` catch up option, we'd emit a warning and process from genesis. [Discussion](https://github.com/ava-labs/awm-relayer/pull/114#discussion_r1443050931).

@minghinmatthewlam [suggested](https://github.com/ava-labs/awm-relayer/pull/114#discussion_r1446611409) we instead emit a warning and exit in this case, leaving it up to the user to explicitly specify that they want to process blocks from genesis.

## How this works
- Updates the `StartBlockHeight` config option to be a signed integer, with a valid range of [-1, +maxint], with -1 representing genesis (i.e. block height 0). We can't use 0 directly because Go JSON unmarshalling does not let us distinguish between a provided value of 0 and the key not being provided at all, since in both cases the Go int is initialized to 0.
- Also updates the relayer goroutine wait logic to exit as soon as a single relayer fails. This is related to https://github.com/ava-labs/awm-relayer/issues/30. In the case of a consistently failing relayer goroutine, the failing source chain should be disabled in the config.

## How this was tested
CI

## How is this documented
N/A


config/config.go
```diff
@@ -216,6 +220,9 @@ func (c *Config) Validate() error {
 			return fmt.Errorf("invalid subnetID in configuration. error: %v", err)
 		}
 		sourceBlockchainIDs = append(sourceBlockchainIDs, blockchainID)
+
+		// Write back to the config
+		c.SourceSubnets[i] = s
```
cam-schultz, line 225: This was an existing bug. `c.SourceSubnets` is a slice of values, not references, so updating `s` (as occurs in `s.Validate`) did not update `c`. We only updated `s` to populate the allowed destinations list, which was not used anywhere in our tests.

config/config.go
```diff
@@ -47,10 +47,14 @@ type SourceSubnet struct {
...
 	SupportedDestinations []string                         `mapstructure:"supported-destinations" json:"supported-destinations"`
-	StartBlockHeight      uint64                           `mapstructure:"start-block-height" json:"start-block-height"`
+	StartBlockHeight      int64                            `mapstructure:"start-block-height" json:"start-block-height"`
 
 	// convenience field to access the supported destinations after initialization
 	supportedDestinations set.Set[ids.ID]
+
+	// convenience field to store the starting block height as a bigint so that we can differentiate
+	// between a 0 value and an unset value
+	StartBlockHeightBigInt *big.Int
```
geoff-vball, line 57: We should probably make `StartBlockHeight` private so that it's not mistakenly used somewhere.

relayer/relayer.go
```diff
@@ -161,34 +159,39 @@ func NewRelayer(
...
 	sub vms.Subscriber,
 	startBlockHeight *big.Int,
 ) error {
-	if startBlockHeight == nil {
-		return fmt.Errorf("startBlockHeight cannot be nil")
-	}
 	// Attempt to get the latest processed block height from the database.
 	// Note that the retrieved latest processed block may have already been partially (or fully) processed by the relayer on a previous run. When
 	// processing a warp message in real time, which is when we update the latest processed block in the database, we have no way of knowing
 	// if that is the last warp message in the block
```
geoff-vball, line 170: I think this comment was supposed to go with the moved line

relayer/relayer.go
```diff
@@ -161,34 +159,39 @@ func NewRelayer(
...
 	//    - In this case, we process from the start block height to the latest block
-	height := startBlockHeight
-	if err == nil {
-		// Use the max of the latest processed block and the start block height
+	var height *big.Int
+	latestProcessedBlockData, err := r.db.Get(r.sourceBlockchainID, []byte(database.LatestProcessedBlockKey))
+	if errors.Is(err, database.ErrChainNotFound) || errors.Is(err, database.ErrKeyNotFound) {
+		// The database does not contain the latest processed block data for the chain, so use the configured StartBlockHeight instead
+		if startBlockHeight == nil {
+			r.logger.Warn("database does not contain latest processed block data and startBlockHeight is nil. Please provide a StartBlockHeight in the configuration.")
```
geoff-vball, line 182: Can we also include the `BlockchainID` of the chain missing the config value?

relayer/relayer.go
```diff
@@ -161,34 +159,39 @@ func NewRelayer(
...
+			return errors.New("database does not contain latest processed block data and startBlockHeight is nil.")
+		}
+		height = startBlockHeight
+	} else if err == nil {
```
geoff-vball, line 186: nit: if we move the else clause on line 209 up here to be `else if err != nil`, and then we can unindent the happy path.

main/main.go
```diff
@@ -159,7 +159,11 @@ func main() {
 	}
 
 	// Create relayers for each of the subnets configured as a source
+	// Each relayer goroutine should persist for the lifetime of the program,
+	// so we exit if any of them fail. In the case of a consistently failing
+	// goroutine, that blockchain should likely be disabled in the configuration.
 	var wg sync.WaitGroup
+	wg.Add(1) // Single counter to exit as soon as any relayer goroutine fails
```
geoff-vball, line 166: Do we just want to get rid of the health check then? It no longer does anything if we exit on the first goroutine failure.

config/config.go
```diff
@@ -278,6 +285,18 @@ func (s *SourceSubnet) Validate(destinationBlockchainIDs *set.Set[string]) error
...
+	// If -1 is provided, set the bigint field to 0 to indicate that the relayer should start from genesis
+	// 0 is treated the same as the field being unset, since there's no way to differentiate between the two when unmarshalling
+	if s.StartBlockHeight < -1 {
+		return fmt.Errorf("invalid start block height in source subnet configuration. Provided height: %d", s.StartBlockHeight)
+	}
+	if s.StartBlockHeight == -1 {
+		s.StartBlockHeightBigInt = big.NewInt(0)
+	} else if s.StartBlockHeight > 0 {
+		s.StartBlockHeightBigInt = big.NewInt(s.StartBlockHeight)
+	}
```
geoff-vball, line 298: Can you leave a separate comment saying that if `StartBlockHeight` is 0, then `StartBlockHeightBigInt` will be nil?

config/config.go
cam-schultz: I believe it needs to be public so that we can unmarshal from JSON into the struct

geoff-vball: Oh hmm seems like you're right.

main/main.go
geoff-vball: Still thinking it through, but I think we should probably keep the original health check. We could panic in certain situations, such as if there's no db entry for a chain and no start height set, which gets us the immediate exit in that scenario but not all

cam-schultz: Wait... that's already implemented lol. I'll just revert the `wg` change altogether

relayer/relayer.go
cam-schultz: Nice spot, made this change

main/main.go
cam-schultz: IMO we should only panic if we encounter something that should otherwise be impossible. Exiting the application should only be done from the top level `main`, or via the health check. Otherwise error handling can become quickly intractible

geoff-vball: I agree with those points for something like `ASSERT` or `os.exit()` in c/c++, but panic in golang is a fair bit different. Idiomatically, panics should be thrown for "should be impossible" situations, but also errors that we know are unrecoverable such as a failed db, misconfiguration, etc. 

IMO, panics actually make error-handling easier, as it will do all the things that bubbling up an error would do - unwinds the stack, executing `defer`'d functions, but also prints a stack-trace which our current error handling doesn't do. It also makes it clear in the code at the earliest possible point that *if this errors, nothing above this can fix it*. The logger we use also contains functions like `StopOnPanic` to catch the panic, log the error, then continue panicing. Repos like `subnet-evm` use panic quite liberally.

All this being said, the failure recovery PR I almost have ready, uses a context instead to cancel the other goroutines, which probably works slightly better because it allows the other goroutines to unwind gracefully.

config/config.go
```diff
@@ -278,6 +285,19 @@ func (s *SourceSubnet) Validate(destinationBlockchainIDs *set.Set[string]) error
...
+	// In this case, StartBlockHeightBigInt is nil
+	if s.StartBlockHeight < -1 {
+		return fmt.Errorf("invalid start block height in source subnet configuration. Provided height: %d", s.StartBlockHeight)
+	}
+	if s.StartBlockHeight == -1 {
+		s.StartBlockHeightBigInt = big.NewInt(0)
+	} else if s.StartBlockHeight > 0 {
+		s.StartBlockHeightBigInt = big.NewInt(s.StartBlockHeight)
+	}
+
```
geoff-vball, line 300: Can you leave a comment above this return that this is the case for s.StartBlockHeight == 0? With the 3 cases above, it's not immediately clear if this is not a coding error.

relayer/relayer.go
```diff
@@ -136,9 +136,7 @@ func NewRelayer(
 	}
 
 	if shouldProcessMissedBlocks {
-		// If StartBlockHeight is not set in the config, then we process from height 0
-		startHeight := big.NewInt(0).SetUint64(sourceSubnetInfo.StartBlockHeight)
-		err = r.processMissedBlocks(sub, startHeight)
+		err = r.processMissedBlocks(sub, sourceSubnetInfo.StartBlockHeightBigInt)
```
geoff-vball, line 139: I'm wondering if we could refactor this part of the code a bit. The issue is that I want to be able to process missed blocks after reconnecting to the subscriber after a subscriber error. The current flow doesn't really work for this. 

Could we create a function `setProcessedBlockHeight()` that takes `shouldProcessMissedBlocks` `sourceSubnetInfo.StartBlockHeightBigInt`, and then does all the logic of checking/setting `latestProcessedBlock` in the db that `processMissedBlocks`, or the call to `SetProcessedBlockHeightToLatest` below are currently doing? 

We could then just call `sub.ProcessFromHeight(<whatever is in the db>)` in every case, and we could also call it after reconnection.

cam-schultz: That's an interesting idea. Slight alteration - I'd replace `sub.ProcessFromHeight(*big.Int)` with `sub.ProcessFromStoredBlock()` so the usage is:
1. Set the starting block in the DB by choosing between the existing value and the config value, according to the established logic
2. call `sub.ProcessFromStoredBlock()`

I'll put this together, and see how it looks.

README.md
```diff
@@ -83,8 +83,12 @@ The relayer is configured via a JSON file, the path to which is passed in via th
   `"supported-destinations": []string` 
   - List of destination subnet IDs that the source subnet supports. If empty, then all destinations are supported.
 
-  `"start-block-height": unsigned integer`
+  `"start-block-height": signed integer`
```
minghinmatthewlam, line 86: why do we want to make this signed

config/config.go
minghinmatthewlam: I agree, was initially confused by this. Would even consider explicitly setting it as `nil` for the case `==0`

relayer/relayer.go
geoff-vball: Yes that sounds great! Thank you!

cam-schultz: Revisiting this code, I don't see any reason why the subscriber needs to access the database at all. This was added before marking a block as processed was outsourced from the subscriber to the relayer.

In this case, your original suggestion makes the most sense. The subscriber will only execute `eth_getLog` queries to add blocks to the process queue, but the relayer will be the entity that processes them and marks them as processed in the db.

Lmk what you think.

geoff-vball: Yes, that sounds great!

geoff-vball (approved):

geoff-vball:
This looks good! I'll probably end up making some small changes to some of it later, but not going to hold up this PR for it.

I noticed we have a race condition while processing missed blocks. Right now we subscribe to the node, which spawns a goroutine which will put new transactions into the logs channel. Then we process historical logs, which will go through all the between whatever we've set and the latest block. The goroutine spawned above could be new blocks to the channel at the same time that `ProcessFromHeight` is, which breaks our assumption of processing blocks in order. 

Now, we can't just wait until we've finished processing historical blocks to subscribe the subscriber, because we might miss blocks between the last historical block we process and when we subscribe. We may just need to call `go s.forwardLogs()` after we've processed the historical blocks.

Additionally, right now `ProcessFromHeight` adds all the found transactions to the logs channel, but nothing ever gets processed until we've completely finished the call to `ProcessFromHeight`. I know we've set the size of the channel to `20000`, which should never normally be exceeded, but in the case that someone actually wants to process historical transactions from genesis in the future, `ProcessFromHeight` will just hang once there are 20000 items in the channel.

michaelkaplan13:
Apologies for a bit of a late comment here, but I think we are likely over-engineering this option using a signed integer with `-1` to support relaying from genesis. I'd recommend that we continue using an unsigned integer, and treat 0 as an invalid option.

The genesis block of an EVM chain is defined by a JSON file of account native token balances, and for a UTXO based chain it is defined by a JSON file with the initial UTXO set, so in either case there are no "transactions" that could send warp messages and need relaying. If there were warp message manually inserted into the genesis, they would need to be relayed explicitly similar to off-chain warp messages either way.

relayer/relayer.go
```diff
@@ -161,55 +165,64 @@ func NewRelayer(
 }
 
 // Helper to process missed blocks. Chooses as the starting block the maximum of the latest processed block and the configured start block height.
-// startBlockHeight cannot be nil
-func (r *Relayer) processMissedBlocks(
+// If the DB does not contain a value, startBlockHeight cannot be nil
+func (r *Relayer) setProcessedBlockHeight(
```
michaelkaplan13, line 169: Wasn't introduced in this PR, but the `error` should be the final return value to keep consistent with conventions.

cam-schultz:
> which breaks our assumption of processing blocks in order

I don't think the race condition you mentioned is an issue, since we actually make no assumption on block processing order. As long as we process all the blocks between the specified start block and current tip, then we are good to go. We can't wait to subscribe until after processing missed blocks for the exact reason you mentioned, so the next best thing is to reverse the order of those operations, and accept the possibility of a message being processed multiple times.

> ProcessFromHeight will just hang once there are 20000 items in the channel.

This is a great catch. I think the fix would be to call `sub.ProcessFromHeight` in a goroutine rather than in a blocking manner, that way we can immediately start processing them [here](https://github.com/ava-labs/awm-relayer/blob/38a08d817a9514df68baa4bdb8248b78ad84ff3c/main/main.go#L246) after the relayer is constructed.

cam-schultz:
> In either case there are no "transactions" that could send warp messages and need relaying

I hadn't realized this, in which case there's no reason for a relayer to ever process the genesis block. I'll make the suggested changes to that config value.

relayer/relayer.go
```diff
@@ -158,63 +164,107 @@ func NewRelayer(
...
+) (*big.Int, error) {
+	var height *big.Int
+	startBlockHeightBigInt := big.NewInt(0).SetUint64(startBlockHeight)
 
-		err = sub.ProcessFromHeight(latestProcessedBlock)
-		if err != nil {
+	// Attempt to get the latest processed block height from the database.
+	// Note that the retrieved latest processed block may have already been partially (or fully) processed by the relayer on a previous run. When
+	// processing a warp message in real time, which is when we update the latest processed block in the database, we have no way of knowing
+	// if that is the last warp message in the block
```
michaelkaplan13, line 183:
```suggestion
	// Note that there may be unrelayed messages in the latest processed block
	// because it is updated as soon as a single message from that block is relayed, 
	// and there may be multiple message in the same block.
```

Nit: Wording, and avoiding using "we"

relayer/relayer.go
```diff
@@ -158,63 +164,107 @@ func NewRelayer(
...
+	if errors.Is(err, database.ErrChainNotFound) || errors.Is(err, database.ErrKeyNotFound) {
+		// The database does not contain the latest processed block data for the chain, so use the configured StartBlockHeight instead
+		if startBlockHeight == 0 {
 			r.logger.Warn(
-				"Encountered an error when processing historical blocks. Continuing to normal relaying operation.",
+				"database does not contain latest processed block data and startBlockHeight is unset. Please provide a non-zero StartBlockHeight in the configuration.",
 				zap.String("blockchainID", r.sourceBlockchainID.String()),
-				zap.Error(err),
 			)
+			return nil, errors.New("database does not contain latest processed block data and startBlockHeight is nil.")
```
michaelkaplan13, line 192:
```suggestion
			return nil, errors.New("database does not contain latest processed block data and startBlockHeight is unset.")
```

relayer/relayer.go
```diff
@@ -158,63 +164,107 @@ func NewRelayer(
...
+	// Note that the retrieved latest processed block may have already been partially (or fully) processed by the relayer on a previous run. When
+	// processing a warp message in real time, which is when we update the latest processed block in the database, we have no way of knowing
+	// if that is the last warp message in the block
+	latestProcessedBlockData, err := r.db.Get(r.sourceBlockchainID, []byte(database.LatestProcessedBlockKey))
+	if errors.Is(err, database.ErrChainNotFound) || errors.Is(err, database.ErrKeyNotFound) {
+		// The database does not contain the latest processed block data for the chain, so use the configured StartBlockHeight instead
+		if startBlockHeight == 0 {
 			r.logger.Warn(
-				"Encountered an error when processing historical blocks. Continuing to normal relaying operation.",
+				"database does not contain latest processed block data and startBlockHeight is unset. Please provide a non-zero StartBlockHeight in the configuration.",
```
michaelkaplan13, line 189:
```suggestion
				"database does not contain latest processed block data and startBlockHeight is unset. Please provide a non-zero startBlockHeight in the configuration.",
```

relayer/relayer.go
```diff
@@ -158,63 +164,107 @@ func NewRelayer(
...
-		// processing a warp message in real time, which is when we update the latest processed block in the database, we have no way of knowing
-		// if that is the last warp message in the block
-		latestProcessedBlock, success := new(big.Int).SetString(string(latestProcessedBlockData), 10)
-		if !success {
-			r.logger.Error("failed to convert latest block to big.Int", zap.Error(err))
-			return err
-		}
+	startBlockHeight uint64,
+) (*big.Int, error) {
+	var height *big.Int
```
michaelkaplan13, line 177: Do we still need to use `big.Int` here, or would it be possible to simplify it to use `uint64`?

michaelkaplan13: Either way, I'd recommend removing the `height` result variable in favor returning the proper value directly in each case down below. Looks like it's only assigned immediately before returning it.

cam-schultz: Simplified to use uint64

geoff-vball:
The race condition does matter if the application crashes while processing historical blocks. If it processes a new block before the historical blocks are done, the new block will be set as the "most recently seen". A crash (or, error thrown by a different relayer goroutine) would mean that on the next startup, the relayer starts processing from newest height, and skips the historical blocks.

geoff-vball (dismissed):
This all looks good to me. I created some tickets for the existing issues I mentioned in the comments.

relayer/relayer.go
```diff
@@ -158,63 +164,102 @@ func NewRelayer(
...
-		// Otherwise, latestProcessedBlock is nil, so we instead store the latest block height.
+
+	// If the database does contain the latest processed block data for the chain,
+	// use the max of the latest processed block and the configured start block height (if it was provided)
+	latestProcessedBlock, success := new(big.Int).SetString(string(latestProcessedBlockData), 10)
+	if !success {
+		r.logger.Error("failed to convert latest block to big.Int", zap.Error(err))
+		return 0, err
+	}
+	if startBlockHeight == 0 || latestProcessedBlock.Uint64() > startBlockHeight {
```
michaelkaplan13, line 209: Sanity checking here, is the `startBlockHeight == 0` condition necessary? I think `latestProcessedBlock.Uint64() > startBlockHeight` is sufficient.

It would also be good to add unit tests for this function.

utils/utils.go
```diff
@@ -94,6 +94,13 @@ func SanitizeHexString(hex string) string {
 	return hex
 }
 
+func MaxBigInt(a, b *big.Int) *big.Int {
```
michaelkaplan13, line 97: This can be removed now I think. If we want to keep it for future use, lets have unit tests for it.

relayer/relayer.go
```diff
@@ -158,63 +164,102 @@ func NewRelayer(
...
+			zap.Error(err),
+		)
+		return 0, err
 	}
-	if errors.Is(err, database.ErrChainNotFound) || errors.Is(err, database.ErrKeyNotFound) {
-		// Otherwise, latestProcessedBlock is nil, so we instead store the latest block height.
+
+	// If the database does contain the latest processed block data for the chain,
+	// use the max of the latest processed block and the configured start block height (if it was provided)
+	latestProcessedBlock, success := new(big.Int).SetString(string(latestProcessedBlockData), 10)
```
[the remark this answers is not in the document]
cam-schultz: Yup, good call

relayer/relayer.go
cam-schultz: That's correct. Simplified this condition and added a unit test

michaelkaplan13 (approved):

geoff-vball (approved):

PR #134, opened 2024-01-11
Failure recovery

geoff-vball:
## Why this should be merged
Makes failure and failure recovery more robust. Closes https://github.com/ava-labs/awm-relayer/issues/30

## How this works
Moves the `subscriber` into the `relayer` struct.

Relayer goroutines will now perpetually try to reconnect to a source-chain node after a disconnection. This only applies after it has successfully connected on startup.

Relayer goroutines trying to reconnect to a source-chain node will set it's health status false until it reconnects. Querying the health-check enpoint will now list the source-chain of any unhealthy relayer goroutines.

Passes a context to every relayer goroutine. The errgroup will cancel the context when any goroutine exits with an error. This will cause every relayer goroutine to gracefully exit if one goroutine exits.

Effectively, the goroutine will no longer exit if it cannot reconnect after a small number of attempts. This will increase the robustness of each goroutine.

A goroutine exitin now indicates a more serious problem, and will cause the other goroutines, and ultimately the program to exit.

TODO We should process missed blocks after a successful reconnection.

## How this was tested
TODO

## How is this documented
Comments


main/main.go
```diff
@@ -251,21 +261,28 @@ func runRelayer(logger logging.Logger,
...
 				logger.Error(
-					"Failed to resubscribe to node. Relayer goroutine exiting.",
+					"Relayer goroutine exiting.",
 					zap.String("originChainID", sourceSubnetInfo.BlockchainID),
 					zap.Error(err),
 				)
-				return
+				return fmt.Errorf("relayer goroutine exiting: %w", err)
 			}
+		case <-ctx.Done():
```
cam-schultz, line 279: Just to confirm my understanding: When a relayer goroutine returns an error in this select loop, the errgroup will cancel the context, and all other relayer goroutines will then be marked unhealthy here. Immediately after, the errgroup will finish waiting and the application will exit anyway.

main/main.go
```diff
@@ -251,21 +261,28 @@ func runRelayer(logger logging.Logger,
...
+		case err := <-relayer.Subscriber.Err():
 			logger.Error(
 				"Received error from subscribed node",
 				zap.String("originChainID", sourceSubnetInfo.BlockchainID),
 				zap.Error(err),
 			)
-			err = subscriber.Subscribe()
+			err = relayer.ReconnectToSubscriber()
 			if err != nil {
 				logger.Error(
```
cam-schultz, line 272: We should mark the relayer as unhealthy in this case too.

relayer/relayer.go
```diff
@@ -26,8 +27,14 @@ import (
 	"go.uber.org/zap"
 )
 
+const (
+	maxSubscribeAttempts   = 10
+	maxResubscribeAttempts = 1000_000
```
cam-schultz, line 32: It'd require a slight rework in `evm/subscriber.go`, but it'd be cleaner to be able to specify in the `Subscribe` argument list that we want to continue attempting to resubscribe forever, rather than passing in a very large number.

relayer/relayer.go
```diff
@@ -217,6 +226,21 @@ func (r *Relayer) processMissedBlocks(
 	return err
 }
 
+// Sets the relayer health status to false while attempting to reconnect.
+func (r *Relayer) ReconnectToSubscriber() error {
+	r.healthStatus.Store(false)
+
+	// Attempt to reconnect the subscription
+	err := r.Subscriber.Subscribe(maxResubscribeAttempts)
```
cam-schultz, line 234: I think we should still limit the number of resubscribe attempts, even if we've successfully subscribed to the node in the past. Resubscription is no more likely to succeed than initial subscription, so we should treat them the same.

cam-schultz: That said, if we drop a connection to a node, we should refresh any values fetched from that node on startup, such as the ChainConfig. I think that would involve a significant refactor though, so let's save it for a future change.

main/main.go
geoff-vball: When a relayer goroutine exits because of an error _anywhere_, the errgroup cancels the context, and all other relayer goroutines will then be marked unhealthy and exit if they're already waiting in this select statement, or once they reach it again.

main/main.go
geoff-vball: The reconnect function marks the relayer as unhealthy/healthy

relayer/relayer.go
geoff-vball: I do think the initial connection and reconnection are distinctly different. If you fail to ever connect, you may have something misconfigured so you will never be able to connect, and exiting the application quickly is probably a good idea. 

If you're trying to reconnect, everything was set up correctly the first time, and something has probably happened to the node. I think we should keep trying to reconnect in case it comes back up, and also not exit the program, because all the other subscribers are also working. We set the unhealthy flag for the reconnecting subscriber, so someone can kill the node if they see it is unhealthy and think that a restart is the correct option.

geoff-vball: Yeah, I agree we should re-fetch values when trying to reconnect. It shouldn't be too bad of a refactor once the open PRs here have landed.

cam-schultz: How about this:
- Let's attempt to resubscribe the same number of times as initial subscription here, and exit the program if we're unable to do so.
- When we implement re-fetching values on successful resubscription, we can relax the resubscription logic to continue to attempt to resubscribe in perpetuity.

My main concern with the proposed change as-is is that when we eventually resubscribe, values might be stale. Quitting early and forcing the program to restart bypasses this.

geoff-vball: Which values in the relayer are stateful? Is it just the Quorum numbers? As-is right now we also won't process blocks missed between disconnection and reconnection, but with the change I asked for in your PR that will be an easy change.

I'm okay to defer both those changes to a future PR if we lower the resubscribe attempts as you suggested here and land this sooner.

main/main.go
geoff-vball: I'll move the set to unhealthy outside of `ReconnectToSubscriber`

relayer/relayer.go
cam-schultz: Yup, for now that's the only state we would need to refresh. But that set could easily grow in the future. I'm envisioning a unified `FetchSubnetValues` method that could be called at startup, on resubscribing, and periodically in the background to keep everything up to date. Then we can easily add in fetched state values in a single place.

relayer/message_relayer.go
```diff
@@ -64,22 +64,19 @@ type messageRelayer struct {
...
 		relayer:                 relayer,
 		warpMessage:             warpMessage,
 		destinationBlockchainID: destinationBlockchainID,
 		messageResponseChan:     messageResponseChan,
-		logger:                  logger,
-		metrics:                 metrics,
-		messageCreator:          messageCreator,
+		logger:                  relayer.logger,
+		metrics:                 relayer.metrics,
+		messageCreator:          relayer.messageCreator,
```
cam-schultz, line 79: We can simplify the `messageRelayer` struct by removing these fields altogether and referencing them via the `relayer` field.

relayer/relayer.go
```diff
@@ -28,8 +30,10 @@ import (
 )
 
 const (
-	maxSubscribeAttempts   = 10
-	maxResubscribeAttempts = 1000_000
+	maxSubscribeAttempts = 10
+	// TODO increase number of resubscribe attempts when we are able to process missed blocks and
+	// refresh the chain config on reconnect.
```
cam-schultz, line 35:
```suggestion
	// TODO attempt to resubscribe in perpetuity once we are able to process missed blocks and
	// refresh the chain config on reconnect.
```

Suggest the following change to reflect the desired function. When we do add that, we should explicitly loop forever, rather than kicking off a long-running loop, even if they are in practice the same.

main/main.go
```diff
@@ -241,50 +243,7 @@ func runRelayer(
 	)
 
 	// Wait for logs from the subscribed node
```
cam-schultz, line 245: Can you update this comment to indicate that the following call will only return on error?

relayer/relayer.go
```diff
@@ -169,6 +179,54 @@ func NewRelayer(
 	return &r, nil
 }
 
+func (r *Relayer) ProcessLogs(ctx context.Context) error {
```
cam-schultz, line 182: Can you add a doc comment to describe this function?

relayer/message_relayer.go
```diff
@@ -64,22 +64,19 @@ type messageRelayer struct {
...
 	warpMessage *warp.UnsignedMessage,
 	destinationBlockchainID ids.ID,
 	messageResponseChan chan message.InboundMessage,
-	messageCreator message.Creator,
 ) *messageRelayer {
 	return &messageRelayer{
 		relayer:                 relayer,
 		warpMessage:             warpMessage,
 		destinationBlockchainID: destinationBlockchainID,
 		messageResponseChan:     messageResponseChan,
```
cam-schultz, line 76: We can also reference `relayer.responseChan` directly instead of passing it as a separate argument.

cam-schultz (approved):
LGTM!

main/main.go
```diff
@@ -105,16 +104,23 @@ func main() {
 		return
 	}
 
-	// Create a health check server that polls a single atomic bool, settable by any relayer goroutine on failure
-	healthy := atomic.Bool{}
-	healthy.Store(true)
+	// Each goroutine will have an atomic bool that it can set to false if it ever disconnects from its subscription.
+	relayerHealth := make(map[ids.ID]*atomic.Bool)
```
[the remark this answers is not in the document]
michaelkaplan13: Oh never mind, I think I misunderstood how it's used. The values in the map (pointers) are only set once by the main goroutine. The other goroutines do not update (or have reference to) the map at all, only the pointer to their respective atomic bool. 👍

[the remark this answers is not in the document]
geoff-vball: We never delete items from the map and adding items to the map only happens at initialization. I can move the addition of keys above the creation of the health endpoint so that the health endpoint always sees the same set of keys.

As for setting the bools, they already need to happen in a linear codepath, or else one path could set the status to healthy after both codepaths set it to unhealthy (aka we need an independent bool for any reason the relayer could be unhealthy). Synchronization isn't going to help this.

The sync package map doesn't give the functionality we need here (there is a range funciton, but is more of a map-reduce) so it's not easy to swap it out.

We could make our own struct with a mutex and the desired functionality, but I really think it's unnecessary

michaelkaplan13 (commented):
Just one small suggestion but otherwise LGTM!

main/main.go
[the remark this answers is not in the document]
michaelkaplan13: Yeah, I misunderstood and thought multiple routines would be updating the value in the map, but that's not the case.

michaelkaplan13 (approved):

Issue #139, opened 2024-01-16
Process Logs while processing historical blocks


geoff-vball:
**Context and scope**
We currently process historical blocks synchronously, which only adds any relevant transactions to a channel. The channel is only processed once we have finished looking at all the historical blocks. We should be consuming the channel at the same time that we're adding to it, or we could end up in a deadlock.

**Discussion and alternatives**
A deadlock here is unlikely in the near-term, since the channel capacity is 20000, and we would need to fill that for a deadlock, but currently if the application crashes or errors while we're adding all the historical transactions to the channel, all progress will be lost. If we consume these transactions at the same time, we will be able to pick up where we left off after a restart.

cam-schultz:
Fixed by https://github.com/ava-labs/awm-relayer/pull/140

Issue #138, opened 2024-01-16
Relayer should process historical blocks first.


geoff-vball:
**Context and scope**
Currently the relayer goroutines will call `forwardLogs()` when first setting up the subscription, causing new logs to be sent to `logsChan` before any historical logs are processed. If the relayer crashes, or the context is cancelled by another relayer coroutine, we may process a new log and write that log's block height to the database. On restart, we will then process from the blockheight in the db, skipping any historical logs that were missed.

**Discussion and alternatives**
Any solution here should also take into account reconnections. Currently, the `forwardLogs()` goroutine will persist even when the subscriber disconnects, so it could lead to new blocks being processed before missed blocks after a reconnection. We should try our best to process logs in block-order so that there is no chance of a database discrepancy.

cam-schultz:
Fixed by https://github.com/ava-labs/awm-relayer/pull/140

PR #140, opened 2024-01-16
Do not store processed height during startup

cam-schultz:
## Why this should be merged
This PR proposes fixes to https://github.com/ava-labs/awm-relayer/issues/139 and https://github.com/ava-labs/awm-relayer/issues/138, which were initially identified in the conversations [here](https://github.com/ava-labs/awm-relayer/pull/133#issuecomment-1889920704) and [here](https://github.com/ava-labs/awm-relayer/pull/133#issuecomment-1890248380), respectively.

## How this works
- Calls `sub.ProcessHistoricalBlocks` in a goroutine to prevent `maxClientSubscriptionBuffer` from filling up before the main relayer processing loop begins relaying messages. Currently, this would lead to deadlock. 
- Adds an argument `storeProcessedHeight` to `Relayer.RelayMessage` to indicate whether or not to write the processed block height to the DB. On startup, historical blocks and live blocks are processed concurrently, so it's not safe to write the processed height to the DB until we are sure blocks are processed in order. One workaround would be to always process historical blocks first so that blocks are always processed in order, but that is a non-trivial change that comes with its own set of edge cases. A simpler approach is to not write to the database while we are catching up. Obviously, if the relayer halts during the catch up process, then on restart the relayer would attempt to re-process blocks, but the existing checks that a message is not yet delivered should prevent the relayer from repeating the most expensive part of processing each block (that being signature aggregation and transaction sending). 

## How this was tested
CI

## How is this documented
N/A


main/main.go
```diff
@@ -212,9 +212,17 @@ func runRelayer(logger logging.Logger,
...
+		zap.String("originBlockchainID", sourceSubnetInfo.BlockchainID),
 	)
 
+	// Marks when the relayer has finished the catch-up process on startup.
+	// Until that time, we do not know the order in which messages are processed,
+	// since the catch-up process occurs concurrently with normal message processing
+	// via the subscriber's Subscribe method. As a result, we cannot safely write the
+	// latest processed block to the database without risking missing a block in a fault
+	// scenario.
+	doneCatchingUp := make(chan bool, 1)
```
geoff-vball, line 224: I think using a channel here is a little overengineered. We can make `doneCatchingUp` an atomic bool that is a field in the `Relayer` object that is initialized to `false`. 

We can then pass a pointer to that atomic bool to `sub.ProcessFromHeight`, which it can set once it finishes. Then we don't need the extra case in the select statement below.

`relayer.RelayMessage` can then just check the bool itself instead of it being passed in.

relayer/relayer.go
```diff
@@ -343,12 +348,32 @@ func (r *Relayer) RelayMessage(warpLogInfo *vmtypes.WarpLogInfo, metrics *Messag
...
-		r.logger.Error(
-			fmt.Sprintf("failed to put %s into database", database.LatestProcessedBlockKey),
-			zap.Error(err),
-		)
+	if storeProcessedHeight {
+		// First, check that the stored height is less than the current block height
+		// This is necessary because the relayer may be processing blocks out of order on startup
+		latestProcessedBlockData, err := r.db.Get(r.sourceBlockchainID, []byte(database.LatestProcessedBlockKey))
+		if err != nil && !errors.Is(err, database.ErrChainNotFound) && !errors.Is(err, database.ErrKeyNotFound) {
+			r.logger.Warn(
```
geoff-vball, line 356: If we're returning an error here, which I believe will cause the application to exit, we should log an error.

relayer/relayer.go
```diff
@@ -343,12 +348,32 @@ func (r *Relayer) RelayMessage(warpLogInfo *vmtypes.WarpLogInfo, metrics *Messag
...
+				"Encountered an unknown error while getting latest processed block from database",
+				zap.String("blockchainID", r.sourceBlockchainID.String()),
+				zap.Error(err),
+			)
+			return err
+		}
+		latestProcessedBlock, success := new(big.Int).SetString(string(latestProcessedBlockData), 10)
+		if !success {
+			r.logger.Error("failed to convert latest block to big.Int", zap.Error(err))
```
geoff-vball, line 365: This should maybe panic. If we can't convert the string to a big int, then the same thing will happen on the next startup, and we'll enter a crashloop until the database is fixed.

relayer/relayer.go
```diff
@@ -343,12 +348,32 @@ func (r *Relayer) RelayMessage(warpLogInfo *vmtypes.WarpLogInfo, metrics *Messag
...
+			return err
+		}
+		if warpLogInfo.BlockNumber > latestProcessedBlock.Uint64() {
+			err = r.db.Put(r.sourceBlockchainID, []byte(database.LatestProcessedBlockKey), []byte(strconv.FormatUint(warpLogInfo.BlockNumber, 10)))
+			if err != nil {
+				r.logger.Error(
+					fmt.Sprintf("failed to put %s into database", database.LatestProcessedBlockKey),
+					zap.Error(err),
+				)
+			}
```
geoff-vball, line 375: If we don't want to return an error here, we should explicity state so. Not sure if it was on purpose or not, but we could definitely get away with swallowing the error here, since worst case is that we reprocess some number of blocks.

relayer/relayer.go
```diff
@@ -149,7 +150,10 @@ func NewRelayer(
 			)
 			return nil, nil, err
 		}
-		sub.ProcessFromHeight(big.NewInt(0).SetUint64(height))
+		// Process historical blocks asynchronously so that the main processing loop can
+		// start processing new blocks as soon as possible. Otherwise, it's possible for
+		// ProcessFromHeight to overload the message queue and cause a deadlock.
+		go sub.ProcessFromHeight(big.NewInt(0).SetUint64(height), doneProcessingMissedBlocks)
```
geoff-vball, line 156: So we have an issue here where all errors from this function now get swallowed. If you leave a TODO here, I'll plumb through the context that I implemented in the failure recovery PR, which will allow us to both cancel the context from inside the goroutine, and listen for cancellations of the other goroutines.

relayer/relayer.go
```diff
@@ -343,12 +348,32 @@ func (r *Relayer) RelayMessage(warpLogInfo *vmtypes.WarpLogInfo, metrics *Messag
 	r.currentRequestID++
 
 	// Update the database with the latest processed block height
-	err = r.db.Put(r.sourceBlockchainID, []byte(database.LatestProcessedBlockKey), []byte(strconv.FormatUint(warpLogInfo.BlockNumber, 10)))
-	if err != nil {
-		r.logger.Error(
-			fmt.Sprintf("failed to put %s into database", database.LatestProcessedBlockKey),
-			zap.Error(err),
-		)
+	if storeProcessedHeight {
```
michaelkaplan13, line 351: To reduce multiple indentation levels, I recommend inverting this if statement such that we return early if not storing the processed height, and otherwise continue on. 

May also want to move the logic of storing the processed height out to helper for readability.

relayer/relayer.go
```diff
@@ -343,12 +348,32 @@ func (r *Relayer) RelayMessage(warpLogInfo *vmtypes.WarpLogInfo, metrics *Messag
...
+		// First, check that the stored height is less than the current block height
+		// This is necessary because the relayer may be processing blocks out of order on startup
+		latestProcessedBlockData, err := r.db.Get(r.sourceBlockchainID, []byte(database.LatestProcessedBlockKey))
```
michaelkaplan13, line 354: I'd recommend a helper for `getLatestProcessedBlock` also, that handles the DB read and conversion to a `big.Int`

main/main.go
cam-schultz: So I repurposed the channel so that it could be used to pass an error to select statement in `ProcessLogs`. We'd probably want a channel to propagate errors anyway.

relayer/relayer.go
cam-schultz: Nice catch, fixed

relayer/relayer.go
cam-schultz: As mentioned [here](https://github.com/ava-labs/awm-relayer/pull/140#discussion_r1456613267), I repurposed the bool channel to also pass an error to the select loop in `ProcessLogs`. This could be done with an atomic bool and the context, but using the channel is more consistent with how we're treating async errors elsewhere, such as the subscriber.

relayer/relayer.go
cam-schultz: Made both of these changes 👍

relayer/relayer.go
cam-schultz: Added

main/main.go
geoff-vball: Sorry this is a novel... We can chat offline if any of this is confusing. 

I think we have a few problems currently: 

1. Errors in the catchup goroutine don't actually get propogated up. We log the error in the goroutine, and then the main thread propogates a new error, instead of one error all the way up.
2. We need to worry about the catchup channel, instead of using idiomatic error handling.
3. Cancellation can only happen in one direction. An error in the catchup block will cause the main goroutine to cancel, but an error in the main goroutine, or context cancellation will leak the catchup goroutine. (We should also fix this with the `forwardLogs()` goroutine)
4. We should update the database when processing the historical blocks, otherwise if we crash/exit before completing, we lose all progress.
5. The current implementation will make it hard to run the catchup goroutine again after a disconnection.

I think we can solve these fairly easily with something approximating the following steps:

- Move the call to ` go s.forwardLogs()` out of `dialAndSubscribe`
- Spawn a goroutine that takes a **_child_** context and calls (if required) `ProcessFromHeight(getLatestProcessedBlockHeight())` and then once that's finished, it can call `forwardLogs`. Both of these functions should check if the context is still live periodically.
  - The child context is to that `ProcessLogs` can cancel it in the case of a disconnection. This way we can create a new instance of the goroutine to process any missed blocks while we were disconnected.
  - Alternatively, we could split the `ProcessFromHeight` and `forwardLogs` into separate goroutines contained by an errGroup, but `ProcessLogs` needs a way of knowing it's allowed to write to the db for logs coming from `ProcessFromHeight`. This means we could process the historical and new blocks at the same time, but at the cost of adding complexity.
- Now when we hit `case err := <-r.Subscriber.Err():`, we should cancel the child context, resubscribe, and then spawn a new goroutine with a new child context.
- Any errors from the goroutine containing `ProcessFromHeight` should probably not propogate all the way up. We should probably swallow and log them in `ProcessLogs` and then try to resubscribe.

relayer/relayer.go
```diff
@@ -174,24 +184,21 @@ func NewRelayer(
...
 		}
-		err = sub.ProcessFromHeight(big.NewInt(0).SetUint64(height))
-		if err != nil {
-			logger.Error(
-				"Failed to process blocks from height on startup",
-				zap.Error(err),
-			)
-			return nil, err
-		}
+		// Process historical blocks asynchronously so that the main processing loop can
```
michaelkaplan13, line 187:
```suggestion
		// Process historical blocks in a separate goroutine so that the main processing loop can
```

relayer/relayer.go
```diff
@@ -438,13 +448,32 @@ func (r *Relayer) RelayMessage(warpLogInfo *vmtypes.WarpLogInfo) error {
...
+	if !storeProcessedHeight {
+		return nil
+	}
+
 	// Update the database with the latest processed block height
-	err = r.db.Put(r.sourceBlockchainID, []byte(database.LatestProcessedBlockKey), []byte(strconv.FormatUint(warpLogInfo.BlockNumber, 10)))
-	if err != nil {
+	// First, check that the stored height is less than the current block height
+	// This is necessary because the relayer may be processing blocks out of order on startup
+	latestProcessedBlock, err := r.getLatestProcessedBlockHeight()
```
michaelkaplan13, line 458: Maybe not for this PR, but could you create an issue to track the latest processed block height in memory so that we can get rid of the unnecessary disk reads here?

michaelkaplan13 (commented):
One follow up, but LGTM otherwise.

main/main.go
cam-schultz: I want to avoid overengineering a non-critical feature. Zooming out to focus on the functional issues, I think the main feature the current implementation is missing is the fact that if the relayer fails during catch up, we lose all progress. As I mentioned in another comment, we don't lose all progress, since the relayer checks if a message has already been delivered before aggregating signatures, but we can certainly do better.

A simple extension to the existing implementation would be to add additional data to the `WarpMessageInfo` to denote if it's a catch-up message. If it is, then we'd always write it to the DB. Otherwise, we only write to the DB if the catch up process is finished. That way, we can process catch up messages and live messages concurrently on startup, and also checkpoint our progress.

I'm open to the refactors you've described, but would rather implement them in a standalone PR (if we decide they're necessary) rather than expanding the scope of this PR.

cam-schultz: > 1. Errors in the catchup goroutine don't actually get propogated up. We log the error in the goroutine, and then the main thread propogates a new error, instead of one error all the way up.

While true, the main goroutine isn't distinguishing between error conditions from the catch up goroutine, so the boolean suffices.

> 2. We need to worry about the catchup channel, instead of using idiomatic error handling.

An error channel is idiomatic for passing errors from child goroutine to parent. We could have separate error/result goroutines, but for the reasons mentioned in my reply to 1., we can get away with just a single channel for now.

> 3. Cancellation can only happen in one direction. An error in the catchup block will cause the main goroutine to cancel, but an error in the main goroutine, or context cancellation will leak the catchup goroutine. (We should also fix this with the forwardLogs() goroutine)

If the main goroutine exits, all child goroutines will exit, so leaky goroutines aren't an issue in practice.

> 4. We should update the database when processing the historical blocks, otherwise if we crash/exit before completing, we lose all progress.

Agreed, it would be better if this were not the case

> 5. The current implementation will make it hard to run the catchup goroutine again after a disconnection.

The catch-up routine gets logs via RPC not websocket, so we don't need a persistent connection for correctness.

michaelkaplan13 (approved):

main/main.go
geoff-vball: > While true, the main goroutine isn't distinguishing between error conditions from the catch up goroutine, so the boolean suffices.

What I'm getting at here is that the error logs become uncorrelated and might be difficult to find the root cause of an error in the logs. If we at least put the error into the channel instead of a bool, we can pass that error up.

> An error channel is idiomatic for passing errors from child goroutine to parent. We could have separate error/result goroutines, but for the reasons mentioned in my reply to 1., we can get away with just a single channel for now.

Not sure what you're suggesting here with two channels, but that's definitely not what I want. We should at least keep the error handling in `ProcessFromHeight` as it was before, returning an error, and wrap the function that puts `true/false` or `err/nil` into the channel in one spot.

> If the main goroutine exits, all child goroutines will exit, so leaky goroutines aren't an issue in practice.

If the main application exits, all goroutines will exit, but there is no child/parent relationship between goroutines.

> The catch-up routine gets logs via RPC not websocket, so we don't need a persistent connection for correctness.

I'm not worried about the connection. If we want to be able to reconnect, then to do so properly we need to be able to catch up on blocks that were missing while we were disconnected. Currently if we disconnect while still processing historical blocks, we can't start another goroutine because then we either have to have to create a new channel, leaking the old channel, or we have two different goroutines potentially writing conflicting things to the channel.

All this being said, I don't want to block this PR with what I've noticed or the suggestions. I'll create a quick followon PR to this as an example, and we can merge it if it seems worthwhile.

geoff-vball (approved):

michaelkaplan13 (dismissed):

relayer/relayer.go
```diff
@@ -201,18 +208,32 @@ func NewRelayer(
...
 				"Handling Teleporter submit message log.",
 				zap.String("txId", hex.EncodeToString(txLog.SourceTxID)),
 				zap.String("originChainId", r.sourceBlockchainID.String()),
 				zap.String("sourceAddress", txLog.SourceAddress.String()),
 			)
 
-			// Relay the message to the destination chain. Continue on failure.
-			err := r.RelayMessage(&txLog)
+			// Messages are either catch-up messages, or live incoming messages.
+			// For live messages, we only write to the database if we're done catching up.
```
minghinmatthewlam, line 235: nit: can you add a comment for why we can store processed height for catch up messages

relayer/relayer.go
```diff
@@ -438,20 +450,67 @@ func (r *Relayer) RelayMessage(warpLogInfo *vmtypes.WarpLogInfo) error {
...
 
 // Returns whether destinationBlockchainID is a supported destination.
 // If supportedDestinations is empty, then all destination chain IDs are supported.
 func (r *Relayer) CheckSupportedDestination(destinationBlockchainID ids.ID) bool {
 	return len(r.supportedDestinations) == 0 || r.supportedDestinations.Contains(destinationBlockchainID)
 }
+
+// Get the latest processed block height from the database.
+// Note that there may be unrelayed messages in the latest processed block
+// because it is updated as soon as a single message from that block is relayed,
```
minghinmatthewlam, line 469: are there any plans to address this in the future? Think we'd need to somehow batch the Warp logs from the same block together somehow in this case.

geoff-vball: Nope. Lots of headache for very little benefit. Currently the worst-case scenario is that we reprocess one block containing transactions, which short-circuits if those messages have already been delivered.

cam-schultz: It's unfortunately a difficult problem to fully solve. The relayer processes Warp messages on a log-by-log basis, not a block-by-block basis, meaning that when the relayer sees a Warp message, it has no way of knowing if that's the last Warp message in that particular block.

So our choices are to either write the block to the DB as soon as we see it, or write it to the DB as soon as we stop seeing that block in new Warp messages. Since the time between Warp messages is unbounded, the former choice is the safest option. Therefore, when reading a block from the database on startup, all we know is that at least one Warp message from that block has been processed, but there may be more. The only workarounds I can think of would be to either process entire blocks atomically, or write each processed Warp message ID to the database, both of which are significant changes. OTOH, the only downside of the current approach is that messages may be processed more than once, but 1) that's limited to a single block and 2) there are safeguards in place to minimize the amount of work the relayer actually does in those cases.

relayer/relayer.go
cam-schultz: Sure thing, expanded this comment

geoff-vball (approved):

minghinmatthewlam (approved):

Issue #131, opened 2024-01-10
messageManager protocolAddress should use common.Address


geoff-vball:
**Context and scope**
The `protocolAddress` field of the `messageManager` should use `common.Address` instead of `common.Hash`.

**Discussion and alternatives**
We had originally used `common.Hash` in order to support future address types, but it has just introduced extra complexity to the system. We currently take addresses, format them into hashes, only to reformat them back into addresses. Supporting any new address scheme will require a thoughtful rearchitecture of many parts of the system. Until then, we can remove this complexity.

PR #143, opened 2024-01-17
Use common.Address instead of common.Hash

geoff-vball:
## Why this should be merged
Closes https://github.com/ava-labs/awm-relayer/issues/131

## How this works
The only place that gives back a `common.Hash` is the warp logs. We convert this into a `common.Address` instead of the other way around.

## How this was tested
Existing tests

### DIFF
--- a/vms/evm/subscriber.go
+++ b/vms/evm/subscriber.go
@@ -105,6 +105,7 @@ func (s *subscriber) NewWarpLogInfo(log types.Log) (*vmtypes.WarpLogInfo, error)
 	}
 
 	return &vmtypes.WarpLogInfo{
+		// BytesToAddress takes the last 20 bytes of the byte array if it is longer than 20 bytes
 		SourceAddress:    common.BytesToAddress(log.Topics[1][:]),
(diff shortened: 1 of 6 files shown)


vms/evm/subscriber.go
```diff
@@ -105,7 +105,7 @@ func (s *subscriber) NewWarpLogInfo(log types.Log) (*vmtypes.WarpLogInfo, error)
 	}
 
 	return &vmtypes.WarpLogInfo{
-		SourceAddress:    log.Topics[1],
+		SourceAddress:    common.BytesToAddress(log.Topics[1][:]),
```
michaelkaplan13, line 108: Could you add a comment here that the data gets (properly) cropped from the left? I had to double check the documentation to myself because I thought this may have used the first 20 bytes of the hash rather than the last 20 bytes.

michaelkaplan13 (commented):
Request 1 doc comment but LGTM otherwise.

michaelkaplan13 (approved):

cam-schultz (approved):

PR #144, opened 2024-01-17
Use crypto library for private keys

geoff-vball:
## Why this should be merged
We reimplemented populating the private key struct before we knew that the crypto library would do it for us.

## How this works
Uses crypto library to create a private key from a hex string.

## How this was tested
existing tests


michaelkaplan13 (approved):
Nice catch 🙏

minghinmatthewlam (approved):
LGTM

Issue #126, opened 2024-01-05
Add config option to process specific Warp messages


cam-schultz:
**Context and scope**
The catch up mechanism is intended to automatically process any blocks that were missed while the relayer was down. https://github.com/ava-labs/awm-relayer/pull/114 adds an option to explicitly set a block height from which to back-process on startup - the relayer will choose between the maximum of this value and the value stored in the DB to determine where to start from.

There's no easy way to process a message that occured in a block _before_ the one stored in the DB. (As a workaround, the DB entry may be manually deleted, and the `StartBlockHeight` option provided). We should give users a way to do this without having to manually alter the DB. In these cases, we should specify Warp messages to process, rather than a block height at which to process from, since this scenario is most likely to occur when the user wishes to relay a specific message that was not processed in due course.

**Discussion and alternatives**
We'll need to add a separate startup routine to process messages by ID, rather than by block.

**Open questions**
How can we map Warp message ID to Teleporter message ID (or any other message protocol)? Would it make more sense to have this be specific to Teleporter? If so, we should implement this such that it can be extended to other future protocols.

michaelkaplan13:
I think the [off-chain warp message support issue](https://github.com/ava-labs/awm-relayer/issues/12) can likely be combined with this one. 

A relayer should be able to specify the off-chain message IDs it wants to relay using this config option pretty easily.

Issue #12, opened 2023-08-28
Support manual "out-of-band" Warp messages


minghinmatthewlam:
**Context and scope**
Out-of-band warp messages:

1. Use the warp precompile directly (not Teleporter)
2. Are not triggered by any on-chain event

We should extend the relayer to allow for manually kicking off the relaying of an out-of-band warp message, identified by it's message hash.

PR #153, opened 2024-01-24
Manual Warp message config option

cam-schultz:
## Why this should be merged
Fixes https://github.com/ava-labs/awm-relayer/issues/126 and unblocks https://github.com/ava-labs/awm-relayer/issues/12

## How this works
Adds a configuration option `ManualWarpMessages` that fully specifies a list of Warp messages to send on startup, independent of the catch-up mechanism or normal operation. The `ManualWarpMessage` type makes as few assumptions about the Warp payload as possible, and so requires the source and destination blockchain IDs and addresses.

Note: once https://github.com/ava-labs/subnet-evm/pull/1062 and https://github.com/ava-labs/coreth/pull/465 are included in releases, we can change the config to only require a Warp message ID, rather than the entire unsigned Warp message bytes.

## How this was tested
Added manual_message.go E2E test.

## How is this documented
Updated README


tests/basic_relay.go
```diff
@@ -298,29 +217,17 @@ func sendBasicTeleporterMessage(
 	destination interfaces.SubnetTestInfo,
 	fundedKey *ecdsa.PrivateKey,
 	fundedAddress common.Address,
-) (teleportermessenger.TeleporterMessage, ids.ID) {
-	log.Info("Packing Teleporter message")
-	teleporterMessage := teleportermessenger.TeleporterMessage{
-		MessageNonce:            big.NewInt(1),
```
cam-schultz, line 304: This was an existing bug. We now get the Teleporter message from the event logs, which contain the correct message nonce.

utils/utils_test.go
```diff
@@ -85,11 +85,6 @@ func TestSanitizeHexString(t *testing.T) {
 			hash:           "1234",
 			expectedResult: "1234",
 		},
-		{
```
cam-schultz, line 88: No longer needed since `strings.TrimPrefix` has no such requirements. This behavior was always a little odd, since `SanitizeHexString` isn't intended to be a verification function.

vms/evm/contract_message.go
```diff
@@ -23,13 +24,22 @@ func NewContractMessage(logger logging.Logger, subnetInfo config.SourceSubnet) *
 }
 
 func (m *contractMessage) UnpackWarpMessage(unsignedMsgBytes []byte) (*vmtypes.WarpMessageInfo, error) {
+	// This function may be called with raw UnsignedMessage bytes or with ABI encoded bytes as emitted by the Warp precompile
+	// The latter case is the steady state behavior, so check that first. The former only occurs on startup.
```
cam-schultz, line 28: Definitely open to feedback on this... When taken from an event log's data, `unsignedMsgBytes` is prefixed by some additional metadata (that I believe comes from ABI encoding the unsigned message bytes). 

I'm sure we could trim this data if it exists in all cases, but that would slightly increase the amount of computation in the normal operation case. Here, the execution time for the normal operation case is unaffected.

vms/evm/contract_message_test.go
```diff
@@ -72,6 +72,7 @@ func TestUnpack(t *testing.T) {
 			input, err := hex.DecodeString(testCase.input)
 			require.NoError(t, err)
 
+			mockLogger.EXPECT().Warn(gomock.Any(), gomock.Any()).Times(testCase.errorLogTimes)
```
cam-schultz, line 75: This had to be added to account for the changes discussed [here](https://github.com/ava-labs/awm-relayer/pull/153/files#r1464206597)

cam-schultz:
https://github.com/ava-labs/coreth/pull/465 is merged, but we should wait to include it here until it's been included in an Avalanchego release. Otherwise, the requests will fail when querying nodes.

tests/manual_message.go
```diff
@@ -0,0 +1,151 @@
...
+
+	delivered2, err := subnetBInfo.TeleporterMessenger.MessageReceived(
+		&bind.CallOpts{}, id2,
+	)
+	Expect(err).Should(BeNil())
+	Expect(delivered2).Should(BeFalse())
+
+	// Cancel the command and stop the relayer
+	relayerCancel()
+	_ = relayerCmd.Wait()
```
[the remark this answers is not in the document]
cam-schultz: It's not necessarily the case that we always want to kill the relayer at the end of the test case. For example in basic_relay.go, we kill the relayer in the middle of the test so that we can rewrite the config. I'm not sure how defer works if the variable containing the function is overwritten in between the defer statement and its execution, but I think explicitly calling it like we do here avoids such confusion, even if it's slightly more error prone.

In any case, if any test errors, the whole suite stops so there's no risk of a failed test impacting subsequent tests.

config/config.go
```diff
@@ -31,6 +31,21 @@ type MessageProtocolConfig struct {
...
+	SourceAddress           string `mapstructure:"source-address" json:"source-address"`
+	DestinationAddress      string `mapstructure:"destination-address" json:"destination-address"`
+
+	// convenience fields to access the values after initialization
+	unsignedMessageBytes    []byte
+	sourceBlockchainID      ids.ID
+	destinationBlockchainID ids.ID
+	sourceAddress           common.Address
+	destinationAddress      common.Address
+}
```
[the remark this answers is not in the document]
cam-schultz: I think that would make `Config` more difficult to use, since the programmer would need to know to _not_ use the `cfg.ManualWarpMessages` and instead use `cfg.ManualWarpMessagesParsed`. We could instead set the JSON field to be `cfg.ManualWarpMessagesUnparsed` and have the parsed field be `cfg.ManualWarpMessages`, but then the `Config` struct becomes more difficult to read.

IMO there's no obviously cleanest solution, so I'm going to keep it as is. The benefit here is that there's only a single struct to worry about, and it's never incorrect to reference the unparsed fields directly, you'd just need to parse them again at the callsite.

tests/utils/utils.go
```diff
@@ -56,3 +69,100 @@ func ReadHexTextFile(filename string) string {
 	Expect(err).Should(BeNil())
 	return strings.TrimRight(string(fileData), "\n")
 }
+
+func CreateDefaultRelayerConfig(
+	subnetAInfo interfaces.SubnetTestInfo,
+	subnetBInfo interfaces.SubnetTestInfo,
```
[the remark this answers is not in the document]
cam-schultz: Good call, updated

tests/manual_message.go
```diff
@@ -0,0 +1,151 @@
...
+
+	fundAmount := big.NewInt(0).Mul(big.NewInt(1e18), big.NewInt(10)) // 10eth
+	fundRelayerTxA := utils.CreateNativeTransferTransaction(
+		ctx, subnetAInfo, fundedKey, relayerAddress, fundAmount,
+	)
+	utils.SendTransactionAndWaitForSuccess(ctx, subnetAInfo, fundRelayerTxA)
+	fundRelayerTxB := utils.CreateNativeTransferTransaction(
+		ctx, subnetBInfo, fundedKey, relayerAddress, fundAmount,
+	)
+	utils.SendTransactionAndWaitForSuccess(ctx, subnetBInfo, fundRelayerTxB)
```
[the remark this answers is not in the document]
cam-schultz: Good call, updated

vms/evm/contract_message.go
cam-schultz: Yeah, it's more of a stylistic thing than anything - keeping the normal case on the shortest path is good practice.

I updated the error logging so that we only emit an error if both cases fail. I kept the Warn log, but bumped it down to Debug so that we have a way to measure one failling but not the other. I also elected to keep this check inline since it's only used in one place, and the individual cases are already outsourced to utilities.

tests/manual_message.go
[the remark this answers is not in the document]
geoff-vball: > In any case, if any test errors, the whole suite stops so there's no risk of a failed test impacting subsequent tests.

I'm pretty sure that's not true. AFAIK this only happens if you pass `--fail-fast` to ginkgo.

> I'm not sure how defer works if the variable containing the function is overwritten

As long as you don't pass the cancelfunc as a parameter to the deferred function, it'll just grab the value from the scope when it executes. This means you'll get the most recent `relayerCancel`. Cancelfuncs are also idempotent so it's fine to defer as many of them as we want.

[Explanation of defer closures here ](https://stackoverflow.com/questions/16010694/how-golangs-defer-capture-closures-parameter)

I do really think we should defer the cancelfunc, but the `relayerCmd.Wait()` should be fine to call at the end.

config/config.go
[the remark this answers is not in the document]
geoff-vball: Yeah, I'm not picky about this one. Was just hoping there was an cleaner way but I think you're right.

tests/utils/utils.go
```diff
@@ -70,92 +73,73 @@ func ReadHexTextFile(filename string) string {
...
-		"blockChainA", subnetAInfo.BlockchainID.String(),
-		"hostB", hostB,
-		"portB", portB,
-		"blockChainB", subnetBInfo.BlockchainID.String(),
-		"subnetA", subnetAInfo.SubnetID.String(),
-		"subnetB", subnetBInfo.SubnetID.String(),
+		"hosts", hosts,
+		"port", ports,
+		"blockchainIDs", blockchainIDs,
+		"subnetIDs", subnetIDs,
```
geoff-vball, line 131: If we move the log inside the foreach loop, we then don't need the slices for hosts, ports, blockchainIDs, subnetIDs

tests/utils/utils.go
```diff
@@ -166,3 +150,20 @@ func RelayerStorageLocation() string {
 func ClearRelayerStorage() error {
 	return os.RemoveAll(storageLocation)
 }
+
+func FundRelayers(
```
geoff-vball, line 154: Love this 🙏

tests/manual_message.go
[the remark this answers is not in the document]
cam-schultz: > I'm pretty sure that's not true. AFAIK this only happens if you pass --fail-fast to ginkgo.

Oh, yeah you're right. We should probably use `--fail-fast` since our tests share network state. Try as we might to isolate them, we're bound to make mistakes.

As for the cancelfunc, it still seems odd to me to defer it in some instances and call it directly in others. In the case of a mid-test relayer restart, we'd call the cancelfunc along with relayer.wait, but at the end of the test we'd only call relayer.wait. My preference is to keep the usage consistent in each test, so in that spirit I'll defer the call in manual_message.go but leave it as is in basic_relay.go

tests/manual_message.go
```diff
@@ -0,0 +1,151 @@
...
+	//
+
+	// Subscribe to the destination chain
+	newHeadsB := make(chan *types.Header, 10)
+	sub, err := subnetBInfo.WSClient.SubscribeNewHead(ctx, newHeadsB)
+	Expect(err).Should(BeNil())
+	defer sub.Unsubscribe()
+
+	log.Info("Starting the relayer")
+	relayerCmd, relayerCancel := testUtils.RunRelayerExecutable(ctx, relayerConfigPath)
```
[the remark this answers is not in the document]
cam-schultz: I want to be wary of adding too many helpers/utilities, as excessive use can obfuscate what the test is doing, making it less readable. I think the balance point here would be to combine the build/exec steps into a common helper, which I've done.

tests/manual_message.go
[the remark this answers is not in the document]
geoff-vball: Oh yeah we wouldn't be able to defer just the cancelfunc. I'm still going to suggest we defer both then, since that's idiomatically how to acheive RAII in go. Cancelfunc is idempotent, and relayerCmd.Wait() just returns an error if it's already been called, but we're not checking the error anyway.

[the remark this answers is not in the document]
geoff-vball: As far as --fail-fast, I'm easy either way, at least for now, but I don't think we should lean on it as an excuse not to clean up the state if we can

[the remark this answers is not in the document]
cam-schultz: Addressed in https://github.com/ava-labs/awm-relayer/pull/153/commits/853222c12a5f02f2edb26c35b656e982217743f8. Thanks for authoring that!

tests/utils/utils.go
```diff
@@ -48,11 +70,107 @@ func RunRelayerExecutable(ctx context.Context, relayerConfigPath string) (*exec.
...
+		"Setting up relayer config",
+	)
+	// Construct the config values for each subnet
+	hosts := make([]string, len(subnetsInfo))
+	ports := make([]uint32, len(subnetsInfo))
+	sources := make([]config.SourceSubnet, len(subnetsInfo))
+	destinations := make([]config.DestinationSubnet, len(subnetsInfo))
+	for i, subnetInfo := range subnetsInfo {
+		var err error
+		hosts[i], ports[i], err = teleporterTestUtils.GetURIHostAndPort(subnetInfo.NodeURIs[0])
```
geoff-vball, line 102:
```suggestion
		host, port, err := teleporterTestUtils.GetURIHostAndPort(subnetInfo.NodeURIs[0])
```

tests/utils/utils.go
```diff
@@ -48,11 +70,107 @@ func RunRelayerExecutable(ctx context.Context, relayerConfigPath string) (*exec.
...
+	teleporterContractAddress common.Address,
+	fundedAddress common.Address,
+	relayerKey *ecdsa.PrivateKey,
+) config.Config {
+	log.Info(
+		"Setting up relayer config",
+	)
+	// Construct the config values for each subnet
+	hosts := make([]string, len(subnetsInfo))
+	ports := make([]uint32, len(subnetsInfo))
```
geoff-vball, line 97:
```suggestion
```

tests/utils/utils.go
```diff
@@ -48,11 +70,107 @@ func RunRelayerExecutable(ctx context.Context, relayerConfigPath string) (*exec.
...
+		hosts[i], ports[i], err = teleporterTestUtils.GetURIHostAndPort(subnetInfo.NodeURIs[0])
+		Expect(err).Should(BeNil())
+
+		sources[i] = config.SourceSubnet{
+			SubnetID:          subnetInfo.SubnetID.String(),
+			BlockchainID:      subnetInfo.BlockchainID.String(),
+			VM:                config.EVM.String(),
+			EncryptConnection: false,
+			APINodeHost:       hosts[i],
+			APINodePort:       ports[i],
```
geoff-vball, line 111:
```suggestion
			APINodeHost:       host,
			APINodePort:       port,
```

tests/utils/utils.go
```diff
@@ -48,11 +70,107 @@ func RunRelayerExecutable(ctx context.Context, relayerConfigPath string) (*exec.
...
+			},
+		}
+
+		destinations[i] = config.DestinationSubnet{
+			SubnetID:          subnetInfo.SubnetID.String(),
+			BlockchainID:      subnetInfo.BlockchainID.String(),
+			VM:                config.EVM.String(),
+			EncryptConnection: false,
+			APINodeHost:       hosts[i],
+			APINodePort:       ports[i],
```
geoff-vball, line 128:
```suggestion
			APINodeHost:       host,
			APINodePort:       port,
```

tests/utils/utils.go
```diff
@@ -48,11 +70,107 @@ func RunRelayerExecutable(ctx context.Context, relayerConfigPath string) (*exec.
...
+			AccountPrivateKey: hex.EncodeToString(relayerKey.D.Bytes()),
+		}
+
+		log.Info(
+			"Creating relayer config for subnet",
+			"subnetID", subnetInfo.SubnetID.String(),
+			"blockchainID", subnetInfo.BlockchainID.String(),
+			"host", hosts[i],
+			"port", ports[i],
```
geoff-vball, line 137:
```suggestion
			"host", host,
			"port", port,
```

geoff-vball (commented):
Just one small refactor suggestion but otherwise lgtm

cam-schultz:
> Just one small refactor suggestion

Thanks for those suggestions. Included in a unified commit

geoff-vball (dismissed):
👍

geoff-vball (approved):

minghinmatthewlam:
double checking here now that `GetMessage` is merged for `subnet-evm` can the config be changed to take message ID, does the same need to be merged for `coreth` first?

cam-schultz:
> double checking here now that `GetMessage` is merged for `subnet-evm` can the config be changed to take message ID, does the same need to be merged for `coreth` first?

Yeah, we need to wait until the coreth changes are included in an avalanchego release first.

config/config.go
```diff
@@ -31,6 +31,21 @@ type MessageProtocolConfig struct {
 	Settings      map[string]interface{} `mapstructure:"settings" json:"settings"`
 }
 
+type ManualWarpMessage struct {
+	UnsignedMessageBytes    string `mapstructure:"unsigned-message-bytes" json:"unsigned-message-bytes"`
```
minghinmatthewlam, line 35: Follow up on the PR description, could we use Warp message ID in the config and use `GetMessage()`?

main/main.go
```diff
@@ -164,6 +166,17 @@ func main() {
 		return
 	}
 
+	manualWarpMessages := make(map[ids.ID][]*vmtypes.WarpLogInfo)
+	for _, msg := range cfg.ManualWarpMessages {
+		sourceBlockchainID := msg.GetSourceBlockchainID()
+
+		warpMessageInfo := vmtypes.WarpLogInfo{
```
minghinmatthewlam, line 173: was there any specific reason we didn't directly save the parsed `warp.UnsignedMessage` in `WarpLogInfo`? I think we should do so for manual and regular warp info, the former when we set up config, and latter when we listen for new warp log events.

config/config.go
```diff
@@ -219,6 +238,59 @@ func (c *Config) Validate() error {
...
+	}
+	destinationBlockchainID, err := ids.FromString(m.DestinationBlockchainID)
+	if err != nil {
+		return err
+	}
+	destinationAddress, err := hex.DecodeString(utils.SanitizeHexString(m.DestinationAddress))
+	if err != nil {
+		return err
+	}
+	m.unsignedMessageBytes = unsignedMsg
```
minghinmatthewlam, line 289: similar to below I think we should parsee as `UnsignedMessage` here

config/config.go
```diff
@@ -219,6 +238,59 @@ func (c *Config) Validate() error {
...
+	return m.sourceAddress
+}
+func (m *ManualWarpMessage) GetDestinationBlockchainID() ids.ID {
+	return m.destinationBlockchainID
+}
+func (m *ManualWarpMessage) GetDestinationAddress() common.Address {
+	return m.destinationAddress
+}
+
+func (m *ManualWarpMessage) Validate() error {
```
michaelkaplan13, line 268: We may want to consider adding a comment documenting that `Validate` initializes the private fields of the struct.

main/main.go
```diff
@@ -243,7 +258,30 @@ func runRelayer(
...
+	for _, warpMessage := range manualWarpMessages {
+		logger.Info(
+			"Relaying manual Warp message",
+			zap.String("blockchainID", sourceSubnetInfo.BlockchainID),
+			zap.String("warpMessageBytes", hex.EncodeToString(warpMessage.UnsignedMsgBytes)),
+		)
+		err := relayer.RelayMessage(warpMessage, false)
+		if err != nil {
+			logger.Error(
+				"Failed to relay manual Warp message. Skipping.",
```
michaelkaplan13, line 275:
```suggestion
				"Failed to relay manual Warp message. Continuing.",
```

vms/evm/contract_message.go
```diff
@@ -23,13 +26,24 @@ func NewContractMessage(logger logging.Logger, subnetInfo config.SourceSubnet) *
 }
 
 func (m *contractMessage) UnpackWarpMessage(unsignedMsgBytes []byte) (*vmtypes.WarpMessageInfo, error) {
+	// This function may be called with raw UnsignedMessage bytes or with ABI encoded bytes as emitted by the Warp precompile
```
michaelkaplan13, line 29: Nit: Could we include a comment above the function definition here, documenting how it can be used? We can leave the comments explaining the reasoning for details of how it's implemented within the body.

tests/utils/utils.go
```diff
@@ -6,15 +6,37 @@ package tests
...
+	teleporterTestUtils "github.com/ava-labs/teleporter/tests/utils"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/log"
 	. "github.com/onsi/gomega"
 )
 
-func RunRelayerExecutable(ctx context.Context, relayerConfigPath string) (*exec.Cmd, context.CancelFunc) {
+var (
+	storageLocation = fmt.Sprintf("%s/.awm-relayer-storage", os.TempDir())
```
michaelkaplan13, line 30: Could you add a comment stating that it's okay to use /tmp for tests. I saw this at first and was confused why we were using it for storage.

michaelkaplan13 (commented):
Few small nits, but mostly LGTM

config/config.go
cam-schultz: Looks like this endpoint is available in avalanchego v1.11.0-fuji, so we should be good to update this PR to match. I'm also considering merging this PR as is and following up with another PR that makes this change while also updating to the latest avalanchego.

main/main.go
cam-schultz: We do save the parsed Warp message in the `WarpMessageInfo` type, but it's worth noting that https://github.com/ava-labs/awm-relayer/pull/156, which has this branch as its target, removes `WarpMessageInfo` altogether and does the parsing internally in `MessageManager`.

geoff-vball (approved):

main/main.go
minghinmatthewlam: I mean instead of saving the message's bytes `[]byte` we can directly save the `warp.UnsignedMessage`, and not need to do multiple parses

config/config.go
minghinmatthewlam: I'm good with that since other PRs are dependent on this one as base, can you create a new issue for it and link it here?

main/main.go
cam-schultz: Yup, in the PR I linked above most of the interfaces are changed to accept a `*warp.UnsignedMessage` directly. We only need to parse once in the message manager.

config/config.go
cam-schultz: Sure thing: https://github.com/ava-labs/awm-relayer/issues/171

minghinmatthewlam (approved):
LGTM!

Issue #119, opened 2023-12-19
Add config option to use Warp API to gather signatures


cam-schultz:
**Context and scope**
We directly make AppRequests to validators for their BLS signatures, and then aggregate the signatures internally. Both `subnet-evm` and `coreth` provide a Warp API to do this instead. We should add a configuration option to use these APIs.

**Discussion and alternatives**
The main advantage of directly making AppRequests from the relayer is that this should work for any VM out of the box, whereas in order to use the API, the VM must provide that API (which is not guaranteed to have a standard interface with respect to other VMS) and the relayer must integrate it. The API approach would require more VM-specific code within the relayer, but would greatly simplify the signature aggregation complexity by offloading it to the VMs. We should consider what long term approach we want to take; likely supporting both will not be necessary nor desirable.

cam-schultz:
We will need to provide the Warp API node endpoint as part of the configuration.

iansuvak:
Should we close this? It looks like it was fixed with #345 ?

PR #156, opened 2024-01-25
Off chain warp messages

cam-schultz:
## Why this should be merged
Adds support for off-chain Warp messages that upgrade the `TeleporterMessenger` version registered in `TeleporterRegistry` on a given chain. Does **not** provide support for generic off-chain messages that invoke arbitrary methods of arbitrary contracts. This design can be extended to support that in the future.

Fixes https://github.com/ava-labs/awm-relayer/issues/12

## How this works
Functional changes:
- Adds package `offchainregistry` that implements `MessageManager`. This type specifically implements off-chain Warp messages that call the `addProtocolVersion` method of `TeleporterRegistry`
- Adds a configuration for this message protocol consisting of the `TeleporterRegistry` contract address. Each chain that is configured to use this message protocol must provide their `TeleporterRegistry` address.

Additional changes:
- Updates `subnet-evm` to `v0.5.11`
- Updates `avalanchego` to `v1.10.18`
  - This version of `avalanchego` broke the relayer's integration with the App Request test network. Rather than debugging something we're likely to deprecate anyway (see https://github.com/ava-labs/awm-relayer/issues/119), I elected to switch to calling `createSignedMessage` instead of `createSignedMessageAppRequest` in this PR. We should follow up and remove all of the App Request code from the relayer entirely.
- Simplifies the MessageManager` interface
- Removes unneeded `WarpMessageInfo` type

## How this was tested
- Added `teleporter_registry.go` E2E test
- CI

## How is this documented
N/A


go.mod
```diff
@@ -3,10 +3,10 @@ module github.com/ava-labs/awm-relayer
...
 
 require (
-	github.com/ava-labs/avalanchego v1.10.17
-	github.com/ava-labs/coreth v0.12.9-rc.9
-	github.com/ava-labs/subnet-evm v0.5.10
-	github.com/ava-labs/teleporter v0.0.0-20240108172200-f03f526e5312
+	github.com/ava-labs/avalanchego v1.10.18
+	github.com/ava-labs/coreth v0.12.10-rc.5
+	github.com/ava-labs/subnet-evm v0.5.11
+	github.com/ava-labs/teleporter v0.1.1-0.20240125232816-d92ef8e89b7a
```
geoff-vball, line 9: I have so many questions about this tag... How did we get `v0.1.1-0`? I don't see that tag in the teleporter repo. Also why is there a dash? Was this supposed to be an rc?

messages/off-chain-registry/message_manager.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/ava-labs/avalanchego/vms/platformvm/warp"
+	"github.com/ava-labs/awm-relayer/config"
+	"github.com/ava-labs/awm-relayer/vms"
+	teleporterregistry "github.com/ava-labs/teleporter/abi-bindings/go/Teleporter/upgrades/TeleporterRegistry"
+	"github.com/ethereum/go-ethereum/common"
+	"go.uber.org/zap"
+)
+
+var (
+	OffChainRegistrySourceAddress = common.HexToAddress("0x0000000000000000000000000000000000000000")
```
geoff-vball, line 18: nit: could be `common.Address{}`

messages/off-chain-registry/message_manager.go
```diff
@@ -0,0 +1,105 @@
...
+
+type messageManager struct {
+	logger             logging.Logger
+	destinationClients map[ids.ID]vms.DestinationClient
+	registryAddress    common.Address
+}
+
+func NewMessageManager(
+	logger logging.Logger,
+	messageProtocolAddress common.Address,
```
geoff-vball, line 33: This is unused. I don't think we need to keep the function signature for `NewMessageManager` the same, so I'd just remove it. If we do need to keep it for some reason, just rename it to `_`

messages/off-chain-registry/message_manager.go
```diff
@@ -0,0 +1,105 @@
...
+
+func (m *messageManager) SendMessage(signedMessage *warp.Message, destinationBlockchainID ids.ID) error {
+	// Get the correct destination client from the global map
+	destinationClient, ok := m.destinationClients[destinationBlockchainID]
+	if !ok {
+		return fmt.Errorf("relayer not configured to deliver to destination. DestinationBlockchainID=%s", destinationBlockchainID)
+	}
+
+	// Construct the transaction call data to call the TeleporterRegistry contract.
+	callData, err := teleporterregistry.PackAddProtocolVersion(0)
```
geoff-vball, line 75: Can you add a comment saying we only include one warp message at a time, so the warp message index is always 0?

messages/off-chain-registry/message_manager.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		m.logger.Error(
+			"Failed packing receiveCrossChainMessage call data",
+			zap.String("destinationBlockchainID", destinationBlockchainID.String()),
+			zap.String("warpMessageID", signedMessage.ID().String()),
+		)
+		return err
+	}
+
+	err = destinationClient.SendTx(signedMessage, m.registryAddress.Hex(), addProtocolVersionGasLimit, callData)
```
geoff-vball, line 85: I think we should probably test the transaction before we actually send it. Right now if the relayer restarts, it'll try to send the message again, which will revert, right?

go.mod
cam-schultz: This includes the changes in https://github.com/ava-labs/teleporter/pull/256. `d92ef8e89b7a` is the first commit of that PR - I'll go back and fetch the latest commit once that PR is merged.

From what I've observed, when fetching a non-tagged commit using `go get`, go.mod will take the latest semver release and automatically increment the bugfix version before appending the timestamp and commit hash.

The changes to go.mod here were made by running `go get github.com/ava-labs/teleporter@d92ef8e89b7a`

messages/off-chain-registry/message_manager.go
cam-schultz: Yes, but I think it's a bit clearer that this is the zero address and not an uninitalized value. If we could make this a const, then I'd be on board but unfortunately common.Address is not a const-compatible type.

messages/off-chain-registry/message_manager.go
cam-schultz: Since the tx bytes will be different, we'll actually successfully send the tx each time. It'd then be up to the TeleporterRegistry implementation to revert. That very well may be the case, but the relayer shouldn't depend on contract implementation details to function correctly.

We should definitely be checking if the address/version has already been registered, and abort relaying if so. I'll implement `ShouldSendMessage` to check for this. Thanks for pointing this out.

messages/off-chain-registry/message_manager.go
cam-schultz: Added

messages/off-chain-registry/message_manager.go
cam-schultz: Good call, removed.

go.mod
geoff-vball: Okay thank you!

michaelkaplan13:
Synced briefly offline, but commenting here for visibility that I'll defer reviewing this PR until we have further design discussions regarding whether off-chain warp message support in the relayer should be made contract-agnostic by having the configuration specify the raw input data of the transaction to be sent to deliver the message. 

This would allow for specifying the destination contract, function, and parameter values in the configuration, and would remove the need for any contract specific code in the relayer. The downside @cam-schultz called out is that the relayer then would not know how to check whether or not the message has been properly delivered or not, creating a risk that it could try to deliver the message multiple times if stuck in a restart loop with a off-chain message specified in its configuraiton.

cam-schultz:
> Synced briefly offline, but commenting here for visibility that I'll defer reviewing this PR until we have further design discussions regarding whether off-chain warp message support in the relayer should be made contract-agnostic by having the configuration specify the raw input data of the transaction to be sent to deliver the message.
> 
> This would allow for specifying the destination contract, function, and parameter values in the configuration, and would remove the need for any contract specific code in the relayer. The downside @cam-schultz called out is that the relayer then would not know how to check whether or not the message has been properly delivered or not, creating a risk that it could try to deliver the message multiple times if stuck in a restart loop with a off-chain message specified in its configuration.

We came to the decision that this should be implemented as a separate `MessageManager` implementation. The main downside is that supporting generic messages limits how fine-grained `ShouldSendMessage` is able to be. Thus, generic message support and more "strongly typed" message support represent different use cases. I created https://github.com/ava-labs/awm-relayer/issues/165 to track the generic off-chain message use case

geoff-vball:
Can you fix the merge conflicts here? Then I'll take a look

cam-schultz:
> Can you fix the merge conflicts here? Then I'll take a look

Should be all set now

messages/off-chain-registry/message_manager.go
```diff
@@ -0,0 +1,199 @@
...
+		return false, nil
+	}
+
+	version, err := registry.GetVersionFromAddress(&bind.CallOpts{}, entry.ProtocolAddress)
+	if err != nil {
+		// Check if execution reverted due to the address not being found
+		if strings.Contains(err.Error(), revertAddressNotFoundString) {
+			return true, nil
+		}
+		m.logger.Info(
```
geoff-vball, line 142: Should this be an error log?

messages/off-chain-registry/message_manager.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	// Check if the version/protocol address tuple is already registered in the TeleporterRegistry contract.
+	registry, err := teleporterregistry.NewTeleporterRegistryCaller(m.registryAddress, client)
+	if err != nil {
+		m.logger.Error(
+			"Failed to create TeleporterRegistry caller",
+			zap.Error(err),
+		)
+		return false, err
```
geoff-vball, line 112: Would it make sense for all the errors in this function to wrap the error we return with the error message, and then log the error from the calling function?

cam-schultz: We don't really have a formal logging strategy in this repo, but in general we err on the side of over rather than under logging. What that translates to is that most of the data structures are composed of `Logger`s, and log internally before returning an error, which may lead to the same error being logged again.

As for wrapping, I think that would be a good idea, but I also think it would bloat the source code significantly when combined with the above logging strategy. We'd have to log the error wrapping text and then also wrap the error itself with the same text before returning.

Thinking about how to unify this, we could implement our own `Logger`, adding a new method `WrapError` that 1) logs the wrapped error and 2) returns the wrapped error so that it can easily be propagated. Something like:

```
func (l *LoggerWrapper) WrapError(wrapText string, err error) error {
    wrappedErr := fmt.Errorf("%s: %w", wrapText, err)
    l.Error(
        wrapText,
        zap.Error(err),
    )
    return wrappedErr
}
```

Regardless, I think the current implementation in this PR is generally consistent with how we're handling error logging in the rest of the repo, so I'd prefer to defer to a separate PR in which we unify our error logging across the entire repo at once.

messages/off-chain-registry/message_manager.go
cam-schultz: Yep! Fixed.

messages/off-chain-registry/message_manager.go
geoff-vball: Yeah no problem for this PR. I've just been thinking about this a bit and would love a way to get the code size down a bit if possible.

geoff-vball (dismissed):

messages/off-chain-registry/message_manager.go
```diff
@@ -0,0 +1,198 @@
...
+		if strings.Contains(err.Error(), revertVersionNotFoundString) {
+			return true, nil
+		}
+		m.logger.Error(
+			"Failed to get address from version",
+			zap.Error(err),
+		)
+		return false, err
+	}
+	if address == entry.ProtocolAddress {
```
michaelkaplan13, line 126: If there is any address registered for this versions already, should we not send the message?

messages/off-chain-registry/message_manager.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+	return &messageManager{
+		logger:             logger,
+		destinationClients: destinationClients,
+		registryAddress:    common.HexToAddress(messageConfig.TeleporterRegistryAddress),
+	}, nil
+}
+
+// ShouldSendMessage returns false if the version/protocol address tuple is already registered in the TeleporterRegistry contract.
+func (m *messageManager) ShouldSendMessage(unsignedMessage *warp.UnsignedMessage, destinationBlockchainID ids.ID) (bool, error) {
```
michaelkaplan13, line 69: Could you add some unit tests for the new message manager implementation (particularly `ShouldSendMessage`)

messages/off-chain-registry/message_manager.go
cam-schultz: Yup, that's indeed the case. We only send the message if the version is not found. In that case, we don't care if the address is already registered, since the same address can map to multiple versions. We can therefor eliminate the call to GetVersionFromAddress altogether.

messages/off-chain-registry/message_manager.go
cam-schultz: Sure thing, added unit tests in for `ShouldSendMessage`. I omitted them for `SendMessage`, since that's basically a pass through to `DestinationClient.SendTx`, which already has unit tests.

messages/off-chain-registry/message_manager.go
cam-schultz: Yup, that's indeed the case. We only send the message if the version is not found. In that case, we don't care if the address is already registered, since the same address can map to multiple versions. We can therefor eliminate the call to GetVersionFromAddress altogether.

geoff-vball (approved):

michaelkaplan13 (approved):
LGTM!

Lets loop back to the longer term conversation of where it makes most sense to relay off-chain messages from soon-ish.

PR #157, opened 2024-01-26
Fetch warp numerator (rebased off of `off-chain-warp-messages`)

cam-schultz:
## Why this should be merged
Fixes https://github.com/ava-labs/awm-relayer/issues/8

## How this works
Fetches the Warp Quorum from the Warp Config on startup. Hardcodes the quorum value for the primary network, since that will never change.

## How this was tested
Added unit tests

## How is this documented
N/A


relayer/message_relayer.go
```diff
@@ -53,17 +54,20 @@ type messageRelayer struct {
...
 	warpMessage             *avalancheWarp.UnsignedMessage
 	destinationBlockchainID ids.ID
+	warpQuorum              config.WarpQuorum
 }
 
 func newMessageRelayer(
 	relayer *Relayer,
 	warpMessage *avalancheWarp.UnsignedMessage,
 	destinationBlockchainID ids.ID,
+	cfg config.Config,
```
geoff-vball, line 64: Let's just make the config a field in the `Relayer` struct so we don't have to pass it around like this

main/main.go
```diff
@@ -49,6 +49,12 @@ func main() {
 		fmt.Printf("couldn't build config: %s\n", err)
 		os.Exit(1)
 	}
+	// Initialize the Warp Quorum values by fetching via RPC
+	// We do this here so that BuildConfig doesn't need to make RPC calls
+	if err = cfg.InitializeWarpQuorum(); err != nil {
+		fmt.Printf("couldn't initialize warp quorum: %s\n", err)
+		os.Exit(1)
```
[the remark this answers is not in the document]
geoff-vball: I don't think there's any reason to prefer and exit of a panic here. We currently basically get the same behaviour between panic and exit, but if any of this code becomes encapsulated, or any defer statements are made before any of these exit points, it'll cause problems.

A panic causing the program to exit will print `exit status 2`, which is go's exit status, but it actually returns `1` as the exit code to the system, so we get the same behaviour there.

I think it's a good idea to always prefer using the `%w` verb to wrap errors. Again, it doesn't really matter for this _specific_ code, but it's a good practice so that the type of the error is preserved for `errors.Is` and `errors.As`.

main/main.go
```diff
@@ -86,11 +92,11 @@ func main() {
...
 
 	// Initialize metrics gathered through prometheus
-	gatherer, registerer, err := initMetrics()
+	gatherer, registerer, err := initializeMetrics()
 	if err != nil {
-		logger.Fatal("failed to set up prometheus metrics",
+		logger.Fatal("Failed to set up prometheus metrics",
 			zap.Error(err))
-		panic(err)
+		return
```
geoff-vball, line 99: `return` after a `logger.Fatal` is very strange. I think we should either mimic the behaviour of the most common `logger.Fatal` - `panic`, or make this an error log.

cam-schultz: Yup, agreed. I went ahead and replaced all the "if error then log and return" with panics, with the exception of the final "Relayer exiting" log since it's the last statment of the `main` function.

main/main.go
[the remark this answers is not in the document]
cam-schultz: Great points! I've gone ahead and changed all the returns with panics in the `main` function.

relayer/message_relayer.go
cam-schultz: Good call, made this change.

config/config.go
```diff
@@ -294,6 +305,127 @@ func (m *ManualWarpMessage) Validate() error {
...
+	}
+	return WarpQuorum{}, fmt.Errorf("failed to find warp config for blockchain %s", blockchainID)
+}
+
+func (c *Config) InitializeWarpQuorum() error {
+	c.warpQuorum = make(map[ids.ID]WarpQuorum)
+
+	// Fetch the Warp quorum values for each source subnet
+	for _, sourceSubnet := range c.SourceSubnets {
+		blockchainID, err := ids.FromString(sourceSubnet.BlockchainID)
```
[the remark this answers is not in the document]
cam-schultz: Yup, blockchains are the distinct unit in the relayer config, so multiple blockchains in the same subnet would need their own distinct entries in the config.

config/config.go
```diff
@@ -294,6 +305,127 @@ func (m *ManualWarpMessage) Validate() error {
...
+	blockchainID ids.ID,
+	subnetID ids.ID,
+	client ethclient.Client,
+) (WarpQuorum, error) {
+	if subnetID == constants.PrimaryNetworkID {
+		return WarpQuorum{
+			QuorumNumerator:   warp.WarpDefaultQuorumNumerator,
+			QuorumDenominator: warp.WarpQuorumDenominator,
+		}, nil
+	}
```
geoff-vball, line 328: Warp can't send messages to the P and X chains, can it? Do we know what happens if someone tried to? Is there a way we can grab the blockchainID for the c-chain?

config/config.go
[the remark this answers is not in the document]
geoff-vball: So probably not something for this PR, but I think that means the slices in the config should be `SourceBlockchains` and `DestinationBlockchains`. I think we're fine for now, but I can definitely see this causing confusion for anyone else who comes along in the future (and anyone consuming/forking this repo)

geoff-vball (commented):
Just a couple more small questions about subnetID/blockchainID, but otherwise LGTM

config/config.go
```diff
@@ -294,6 +305,127 @@ func (m *ManualWarpMessage) Validate() error {
 	return nil
 }
 
+// If the numerator in the Warp config is 0, use the default value
+func setQuorumNumerator(cfgNumerator uint64) uint64 {
```
michaelkaplan13, line 309: Can we rename this function since it's not actually setting anything itself?

config/config.go
```diff
@@ -294,6 +305,127 @@ func (m *ManualWarpMessage) Validate() error {
...
+		return warp.WarpDefaultQuorumNumerator
+	}
+	return cfgNumerator
+}
+
+// Helper to retrieve the Warp Quorum from the chain config.
+// Differentiates between subnet-evm and coreth RPC internally
+func getWarpQuorum(
+	blockchainID ids.ID,
+	subnetID ids.ID,
```
michaelkaplan13, line 320: Nit: Can we put `subnetID` as the first param here since it's technically a broader scope than `blockchainID`? 

Every blockchain ID belongs to exactly one subnet ID, but agree it's not worth doing the lookup of the subnet ID based on the blockchain ID here since it's easier to be passed both.

config/config.go
michaelkaplan13: > Warp can't send messages to the P and X chains, can it? Do we know what happens if someone tried to?
That's correct for now, but in the future I would expect warp support on all chains, including X and P. There technically isn't way to try to send a message to the X or P chain since their VMs do not support it at all.

I'm neither here or there on using the C-chain blockchain ID as the condition here rather than the primary network subnet ID though.

config/config.go
```diff
@@ -294,6 +305,127 @@ func (m *ManualWarpMessage) Validate() error {
...
+		cfg, ok := precompile.Config.(*warp.Config)
+		if ok {
+			if warpConfig == nil {
+				warpConfig = cfg
+				continue
+			}
+			if *cfg.Timestamp() > *warpConfig.Timestamp() {
+				warpConfig = cfg
+			}
+		}
```
michaelkaplan13, line 351:
```suggestion
                if !ok {
                        continue
                } 
	        if warpConfig == nil {
			warpConfig = cfg
			continue
		}
		if *cfg.Timestamp() > *warpConfig.Timestamp() {
			warpConfig = cfg
		}
```

Nit: I'm all about minimizing nesting whenever possible

config/config.go
```diff
@@ -294,6 +305,127 @@ func (m *ManualWarpMessage) Validate() error {
...
+	}
+	if warpConfig != nil {
+		return WarpQuorum{
+			QuorumNumerator:   setQuorumNumerator(warpConfig.QuorumNumerator),
+			QuorumDenominator: warp.WarpQuorumDenominator,
+		}, nil
+	}
+
+	// If we didn't find the Warp config in the upgrade precompile list, check the genesis config
+	warpConfig, ok := chainConfig.GenesisPrecompiles["warpConfig"].(*warp.Config)
```
michaelkaplan13, line 361: Can we move "warpConfig" to a const?

config/config.go
cam-schultz: The Warp protocol is compatible with the P and X chains, though support has not been implemented (nor am I aware of any plans to do so).

However, since neither of them are EVM chains, the `evm` VM type in the relayer config will fail to work properly for those chains. I'm not exactly sure what would happen, but it wouldn't result in a tx posted to either of those chains.

> Is there a way we can grab the blockchainID for the c-chain?

There are endpoints to map subnet ID <-> blockchain ID, but the current config requires both to be specified by the user. I think this is the most general approach setup, since no subnet is guaranteed to have a fixed number of blockchains in perpetuity. I also think we should keep the primary network special casing as minimal as possible.

config/config.go
```diff
@@ -88,6 +96,9 @@ type Config struct {
 	// convenience fields to access the source subnet and chain IDs after initialization
 	sourceSubnetIDs     []ids.ID
 	sourceBlockchainIDs []ids.ID
+
+	// convenience field to store the Warp quorum figures for each subnet
+	warpQuorum map[ids.ID]WarpQuorum
```
michaelkaplan13, line 101: I think we only need the quorum values for the destination chains right? If so, I'd recommend putting this within `DestinationSubnet` rather than an extra map in `Config`

config/config.go
```diff
@@ -294,6 +305,127 @@ func (m *ManualWarpMessage) Validate() error {
...
+			QuorumDenominator: warp.WarpQuorumDenominator,
+		}, nil
+	}
+	return WarpQuorum{}, fmt.Errorf("failed to find warp config for blockchain %s", blockchainID)
+}
+
+func (c *Config) InitializeWarpQuorum() error {
+	c.warpQuorum = make(map[ids.ID]WarpQuorum)
+
+	// Fetch the Warp quorum values for each source subnet
```
michaelkaplan13, line 374: I don't think we should need the quorum values for source subnet since it is only used to determine if a signature is valid upon delivery.

config/config.go
geoff-vball: Thanks!

cam-schultz: > I'm neither here or there on using the C-chain blockchain ID as the condition here rather than the primary network subnet ID though.

Using the subnetID would be simpler because the primary network subnet ID is always the constant `ids.Empty`, whereas the C-Chain blockchain ID would have to be fetched from the network.

config/config.go
geoff-vball: Golang doesn't support runtime consts - only compile-time consts. If the value can't be determined by the compiler, it can't be a const - so basically only primitive types.

michaelkaplan13: Ah, I meant the string literal `"warpConfig"` that's used in this line haha

config/config.go
cam-schultz: Made this change as suggested. One thing to note is that `Config.GetWarpQuorum` now iterates over the list of destination subnets rather than indexing into a map. I thought about adding a `map[ids.ID]DestinationSubnet` to `Config`, but since the length of that list is effectively constant, iterating is fine.

config/config.go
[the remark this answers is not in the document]
cam-schultz: Ticket here: https://github.com/ava-labs/awm-relayer/issues/170

config/config.go
```diff
@@ -294,6 +306,83 @@ func (m *ManualWarpMessage) Validate() error {
...
+		return WarpQuorum{
+			QuorumNumerator:   warp.WarpDefaultQuorumNumerator,
+			QuorumDenominator: warp.WarpQuorumDenominator,
+		}, nil
+	}
+
+	// Fetch the subnet's chain config
+	chainConfig, err := client.ChainConfig(context.Background())
+	if err != nil {
+		return WarpQuorum{}, fmt.Errorf("failed to fetch chain config for blockchain %s: %v", blockchainID, err)
```
geoff-vball, line 334: nit: we should prefer `%w` over `%v` for wrapping errors https://stackoverflow.com/questions/61283248/format-errors-in-go-s-v-or-w

config/config.go
```diff
@@ -294,6 +306,83 @@ func (m *ManualWarpMessage) Validate() error {
...
+}
+
+func (c *Config) InitializeWarpQuorums() error {
+	// Fetch the Warp quorum values for each destination subnet.
+	for i, destinationSubnet := range c.DestinationSubnets {
+		err := destinationSubnet.initializeWarpQuorum()
+		if err != nil {
+			return fmt.Errorf("failed to initialize Warp quorum for destination subnet %s: %v", destinationSubnet.SubnetID, err)
+		}
+		c.DestinationSubnets[i] = destinationSubnet
```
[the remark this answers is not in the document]
cam-schultz: We do have to rewrite to `c` since `destinationSubnet` is a copy of the slice element, and not a reference to the element itself. AFAIK Go has no way to iterate over a slice by reference (compare to `auto& [x, y]` in C++ for example).

That said, I did not realize that Go's JSON package could unmarshal into pointers. I've changed `Config` to be composed of  slices of pointers so that we can iterate over the slice by reference and skip having to write back to the `Config` object.

config/config.go
cam-schultz: Nice catch, I've made this change in all the `fmt.Errorf`'s in this package, but will defer applying this to the rest of the repo for later.

geoff-vball (dismissed):

michaelkaplan13 (dismissed):

geoff-vball (approved):

michaelkaplan13 (approved):

relayer/relayer.go
```diff
@@ -427,7 +429,7 @@ func (r *Relayer) RelayMessage(warpLogInfo *vmtypes.WarpLogInfo, storeProcessedH
 	}
 
 	// Create and run the message relayer to attempt to deliver the message to the destination chain
-	messageRelayer := newMessageRelayer(r, unsignedMessage, destinationBlockchainID)
+	messageRelayer, err := newMessageRelayer(r, unsignedMessage, destinationBlockchainID)
```
michaelkaplan13, line 432: Looks like this `err` check below was unnecessary previously. Glad it's corrected now 👍

PR #169, opened 2024-02-02
Improve tests error handling

minghinmatthewlam:
## Why this should be merged
The error printed would indicate the wrong test case failing. We want to log the actual test case that fails.

Also add additional check and test case for `ProcessFromHeight`

## How this works
We need to create the go mock controllers with `gomock.NewController(t)` while we iterate through the test cases to get the correct test case that fails printed in error.

## How this was tested
If you make one of `TestShouldSendMessage` or `TestSendTx` fail on `main` it will print the error indicating the wrong test case failing. Compared to this branch, it'll indicate the correct test case.

For example, I had `TestShouldSendMessage/not_allowed` fail on `main`, but the error prints 
```
    --- FAIL: TestShouldSendMessage/message_already_delivered (0.00s)
        testing.go:1490: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
```
With these changes will print the actual failing error 
```
    --- FAIL: TestShouldSendMessage/not_allowed (0.00s)
        controller.go:98: missing call(s) to *mocks.MockDestinationClient.Client() /Users/matthewlam/go/src/awm-relayer/messages/teleporter/message_manager_test.go:184
```
CI
## How is this documented
n/a

### DIFF
--- a/vms/evm/subscriber_test.go
+++ b/vms/evm/subscriber_test.go
@@ -2,7 +2,6 @@ package evm
 
 import (
 	"math/big"
-	"os"
 	"testing"
 
 	"github.com/ava-labs/avalanchego/utils/logging"
@@ -11,8 +10,7 @@ import (
 	"github.com/ava-labs/subnet-evm/core/types"
 	"github.com/ava-labs/subnet-evm/ethclient"
 	"github.com/ava-labs/subnet-evm/interfaces"
-	"github.com/ava-labs/subnet-evm/precompile/contracts/warp"
-	"github.com/ethereum/go-ethereum/common"
+	"github.com/stretchr/testify/require"
 	"go.uber.org/mock/gomock"
 )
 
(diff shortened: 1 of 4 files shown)


geoff-vball (approved):
Thanks for fixing this 🙏

vms/evm/subscriber_test.go
```diff
@@ -27,14 +25,7 @@ func makeSubscriberWithMockEthClient(t *testing.T) (*subscriber, *mock_ethclient
...
 
-	logger := logging.NewLogger(
-		"awm-relayer-test",
-		logging.NewWrappedCore(
-			logging.Info,
-			os.Stdout,
-			logging.JSON.ConsoleEncoder(),
-		),
-	)
+	logger := logging.NoLog{}
```
cam-schultz, line 28: Ooh great find

cam-schultz (approved):
I don't really understand why the previous implementation was emitting the wrong test, but glad it's fixed.

PR #180, opened 2024-02-09
Release Job

### LINKED ISSUE #176: Add build jobs
**Context and scope**
We should package releases with pre built binaries for supported systems. We'll want to add CI jobs to do so when a release is created.

We should build for linux/amd64 and darwin/arm64.

cam-schultz:
## Why this should be merged
Adds a unified release workflow that creates releases, builds binaries, and builds docker images for the supported platforms. Releases are triggered on every tag push (but will fail unless the tag follows [semver](https://goreleaser.com/limitations/semver/)), resulting in the following:
- Release notes generated
- Binaries for supported platforms uploaded as artifacts
- Docker images for supported platforms pushed to Dockerhub

A similar setup is used by [avalanche-cli](https://github.com/ava-labs/avalanche-cli), and was adapted to put together this change.

Fixes #176 

## How this works
Uses [Goreleaser](https://goreleaser.com/) to manage releases. 

## How this was tested
- Ran the release job against a dummy tag (see [here](https://github.com/ava-labs/awm-relayer/actions/runs/7848967125) for a successful run). I've since deleted the resulting releases to not muddy the repo.
- Verified the published docker images work with the Teleporter docker setup on both linux/amd64 and darwin/arm64 platforms.

## How is this documented
N/A


.github/workflows/release.yml
```diff
@@ -0,0 +1,67 @@
...
+  push:
+    tags:
+      - "*"
+
+jobs:
+  release:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Git checkout
+        uses: actions/checkout@v3
```
geoff-vball, line 16:
```suggestion
        uses: actions/checkout@v4
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,67 @@
...
+        with:
+          fetch-depth: 0
+          path: awm-relayer
+      - name: Set Go version
+        run: |
+          source ./awm-relayer/scripts/versions.sh
+          echo GO_VERSION=$GO_VERSION >> $GITHUB_ENV
+      - name: Set up Go
+        uses: actions/setup-go@v3
```
geoff-vball, line 25:
```suggestion
        uses: actions/setup-go@v5
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,67 @@
...
+          cd osxcross
+          wget https://github.com/joseluisq/macosx-sdks/releases/download/12.3/$MACOS_SDK_FNAME -O tarballs/$MACOS_SDK_FNAME
+          echo $MACOS_SDK_CHECKSUM tarballs/$MACOS_SDK_FNAME | sha256sum -c -
+          UNATTENDED=1 ./build.sh
+          echo $PWD/target/bin >> $GITHUB_PATH
+        env:
+          MACOS_SDK_FNAME: MacOSX12.3.sdk.tar.xz
+          MACOS_SDK_CHECKSUM: 3abd261ceb483c44295a6623fdffe5d44fc4ac2c872526576ec5ab5ad0f6e26c
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v2
```
geoff-vball, line 58:
```suggestion
        uses: goreleaser/goreleaser-action@v5
```

Dockerfile.release
```diff
@@ -0,0 +1,5 @@
+FROM debian:11-slim
```
michaelkaplan13, line 1: Any reason to use `debian:11-slim` specifically here? Should be fine, I'm just not aware of anywhere else we use it.

Dockerfile.release
```diff
@@ -0,0 +1,5 @@
...
+RUN apt-get update && \
+    apt-get install -y git
```
michaelkaplan13, line 3: What do we need `git` for?

.github/workflows/release.yml
```diff
@@ -0,0 +1,67 @@
...
+jobs:
+  release:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Git checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          path: awm-relayer
+      - name: Set Go version
```
michaelkaplan13, line 20: Nit: Could you put an extra newline in between the steps here so that it's slightly easier to read?

.github/workflows/release.yml
```diff
@@ -0,0 +1,67 @@
...
+          UNATTENDED=1 ./build.sh
+          echo $PWD/target/bin >> $GITHUB_PATH
+        env:
+          MACOS_SDK_FNAME: MacOSX12.3.sdk.tar.xz
+          MACOS_SDK_CHECKSUM: 3abd261ceb483c44295a6623fdffe5d44fc4ac2c872526576ec5ab5ad0f6e26c
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v5
+        with:
+          distribution: goreleaser
+          version: latest
```
michaelkaplan13, line 61: Will this still push a dockerhub image with the specific version tag, or just `latest`?

Dockerfile.release
cam-schultz: That's the image layer `avalanche-cli` uses. I was running into some compatability issues with the golang bullseye layer, so I elected to just go with the battle tested option.

.github/workflows/release.yml
cam-schultz: Yes, it still pushes a tagged image

geoff-vball (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,76 @@
...
+      - "*"
+
+jobs:
+  release:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Git checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
minghinmatthewlam, line 18: why do we need to specify 0 fetch depth here?

cam-schultz: It's probably not strictly needed, but it doesn't hurt to fetch the full commit history in the job.

PR #182, opened 2024-02-13
Check cfg for supported destination

cam-schultz:
## Why this should be merged
Checks the configured list of destination subnets in the call to `CheckSupportedDestination`. Previously, we only checked the source subnet's list of supported destinations.

## How this works
Self explanatoru

## How this was tested
CI

## How is this documented
N/A


relayer/relayer_test.go
```diff
@@ -72,6 +88,13 @@ func TestCheckSupportedDestination(t *testing.T) {
 				supportedDestinations: set.Set[ids.ID]{
 					id1: {},
 				},
+				globalConfig: config.Config{
+					DestinationSubnets: []*config.DestinationSubnet{
+						{
+							BlockchainID: id1.String(),
+						},
+					},
+				},
```
geoff-vball, line 97: Would it be too much trouble to structure the `globalConfig` as we do in the real config, with `SupportedDestinations` inside a `SourceSubnet` which we then validate/parse?

cam-schultz: This has been reverted

config/config_test.go
```diff
@@ -542,3 +543,76 @@ func TestGetWarpQuorum(t *testing.T) {
...
+				ids.Add(id)
+			}
+
+			sourceSubnet := testCase.sourceSubnet()
+			res := sourceSubnet.Validate(&ids)
+			if testCase.expectError {
+				require.Error(t, res)
+			} else {
+				require.NoError(t, res)
+			}
```
michaelkaplan13, line 615: I think we're missing a check on the `expectedSupportedDestinations` here

geoff-vball (dismissed):

config/config_test.go
cam-schultz: Added

michaelkaplan13 (approved):

geoff-vball (dismissed):

relayer/relayer.go
```diff
@@ -99,22 +99,11 @@ func NewRelayer(
...
-	var filteredDestinationClients map[ids.ID]vms.DestinationClient
-	supportedDestinationsBlockchainIDs := sourceSubnetInfo.GetSupportedDestinations()
-	if len(supportedDestinationsBlockchainIDs) > 0 {
-		filteredDestinationClients := make(map[ids.ID]vms.DestinationClient)
-		for id := range supportedDestinationsBlockchainIDs {
-			filteredDestinationClients[id] = destinationClients[id]
-		}
-	} else {
-		filteredDestinationClients = destinationClients
-	}
```
cam-schultz, line 111: I'm not entirely sure why, but this was resulting in an empty `filteredDestinationClients` with the rest of the changes in this PR. I didn't look too deeply into it because this filtering is done in config.go anyway.

michaelkaplan13: I think it's because we were creating a new local `filteredDestinationClients` variable in the `if` clause by using ":=" instead of "="

cam-schultz: Gah that's definitely it. I'm surprised the linter didn't pick up `filteredDestinationClients` as unused, but maybe that's because it's being passed mutably to a function.

michaelkaplan13 (approved):

config/config.go
```diff
@@ -432,6 +432,18 @@ func (s *SourceSubnet) Validate(destinationBlockchainIDs *set.Set[string]) error
...
+	// If the list of supported destinations is empty, populate with all of the configured destinations
+	if len(s.SupportedDestinations) == 0 {
+		for _, blockchainIDStr := range destinationBlockchainIDs.List() {
+			blockchainID, err := ids.FromString(blockchainIDStr)
+			if err != nil {
+				return fmt.Errorf("invalid blockchainID in configuration. error: %w", err)
+			}
+			s.supportedDestinations.Add(blockchainID)
+		}
+	}
```
minghinmatthewlam, line 445: nit: can we return `nil` here to make it clear the below code doesn't apply

geoff-vball (approved):

relayer/message_relayer.go
```diff
@@ -64,7 +64,7 @@ func newMessageRelayer(
 	quorum, err := relayer.globalConfig.GetWarpQuorum(destinationBlockchainID)
 	if err != nil {
 		relayer.logger.Error(
-			"Failed to get warp quorum",
+			"Failed to get warp quorum from config. Relayer may not be configured to deliver to the destination chain.",
```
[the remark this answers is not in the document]
minghinmatthewlam: can you change the ordering between zap.Error and zap.String though?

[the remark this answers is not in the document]
cam-schultz: > nit I don't think we need the extra error log changes, since this shouldn't happen now due to destination chain not being configured.

I'm going to keep the exapanded error log, since it's possible that would change in the future.

config/config.go
cam-schultz: Sure

geoff-vball (approved):

config/config.go
michaelkaplan13: Not to nit pick a nit, but want to note that it's actually more error prone to return early here because it's likely for more validation (possibly not related to the supported destinations) to be added in the future, and if we didn't remove the early return, future validation would not be hit.

I think generally, it's good practice for these types of methods to only have a single "successful" return case at the end, and all early returns be known error cases.

michaelkaplan13 (dismissed):

config/config.go
geoff-vball: Yeah, I'm not too fussed with however we implement this, but I think my ideal solution would be to wrap the for loop starting on line 448 in an `else` clause to this `if`. I normally hate unnecessary nesting, but right now it's not clear from a first pass that these two loops are related.

cam-schultz: Ah that's a good point. I've reverted this change.

geoff-vball (approved):

minghinmatthewlam (approved):

PR #200, opened 2024-02-26
Add devcontainer config and sample relayer config

martineckardt:
## Why this should be merged
Make the awm-relayer easily accessible to a wider range of developers independent of their operating system and system setup. 

## How this works
By leveraging [devcontainers](https://code.visualstudio.com/docs/devcontainers/containers) the developers can open the repository either in a [Codespace](https://github.com/features/codespaces) or locally in a docker container.

## How this was tested
Manually

## How is this documented
Added a note to the readme

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -101,6 +101,7 @@ type Config struct {
 func SetDefaultConfigValues(v *viper.Viper) {
 	v.SetDefault(LogLevelKey, logging.Info.String())
 	v.SetDefault(StorageLocationKey, "./.awm-relayer-storage")
+	v.SetDefault(ProcessMissedBlocksKey, false)
 }
 
 // BuildConfig constructs the relayer config using Viper.
--- a/config/config.go
+++ b/config/config.go
@@ -101,7 +101,7 @@ type Config struct {
 func SetDefaultConfigValues(v *viper.Viper) {
 	v.SetDefault(LogLevelKey, logging.Info.String())
 	v.SetDefault(StorageLocationKey, "./.awm-relayer-storage")
-	v.SetDefault(ProcessMissedBlocksKey, false)
+	v.SetDefault(ProcessMissedBlocksKey, true)
 }
 
 // BuildConfig constructs the relayer config using Viper.
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,13 +3,16 @@
 {
 	"name": "Avalanche-DevContainer",
 
-	"build": {
+	/*"build": {
 		"dockerfile": "../Dockerfile",
 		"args": { "GO_VERSION": "latest" }
-	},
+	},*/
+
+	"image": "avaplatform/awm-relayer:latest",
 
 	"runArgs": ["--network=host"],
 
 	// Uncomment to connect as root instead. More info: https://aka.ms/dev-containers-non-root.
 	"remoteUser": "root"
+
 }
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,6 +1,6 @@
 // For format details, see https://aka.ms/devcontainer.json.
 {
-	"name": "Avalanche Warp Messaging Relayer DevContainer",
+	"name": "AWM Relayer DevContainer",
 
 	"image": "golang:latest",
 	"runArgs": ["--network=host"],
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -2,8 +2,10 @@
 {
 	"name": "Avalanche Warp Messaging Relayer DevContainer",
 
-	"image": "avaplatform/awm-relayer:latest",
+	"image": "golang:latest",
 	"runArgs": ["--network=host"],
 	
-	"remoteUser": "root"
+	"remoteUser": "root",
+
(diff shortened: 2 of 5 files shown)


geoff-vball:
@martineckardt I just changed the base of this PR to be `cleanup-cfg` to make it easier to review, since that branch has a PR [here](https://github.com/ava-labs/awm-relayer/pull/206/commits)

geoff-vball (approved):

cam-schultz (approved):
LGTM! Thanks for putting this together.

michaelkaplan13 (commented):
LGTM, the dev container will be awesome to have!

My only question is if it matters than the .json files here aren't valid JSON due to the comment at the top and the "...", as a value.

martineckardt:
> My only question is if it matters than the .json files here aren't valid JSON due to the comment at the top and the "...", as a value.

@michaelkaplan13 Had a convo about this topic with @cam-schultz yesterday. I created the template so wherever possible we have a valid example value. The only field where this is not possible and the user needs to change the values are `reward-address`, `account-private-key` and `start-block-height`.

For `reward-address` and `account-private-key` I was able to use valid JSON values that would create a clear error message when the config is used to run a relayer (something with malformed address / private key).

With start-block-height it is a bit more tricky. I saw the following options:
- Use 0 as the example value: Results in an [error](https://github.com/ava-labs/awm-relayer/blob/e4cab3998b1a77ba991774602443a6276a12c195/relayer/relayer.go#L272-L273)
- Use some valid number (e.g. 30471060): Results in the relayer processing many blocks at start up, since the delta between the head and the example value grows over time
- Use "..." to make clear that the user has to replace it with a valid value. I added instructions for this to the [academy course](https://academy.avax.network/path-player?courseid=teleporter&unit=65e642df775772eb460509a8Unit) 

I understand you guys had this [conversation](https://github.com/ava-labs/awm-relayer/pull/114#discussion_r1442257670) in January. I would advocate for some kind of way to start with the head to simplify this and avoid having the people to go to the explorer and fetch the latest block height manually either by making the value optional or including a value to indicate that it should start at the chain head (maybe -1, analogous to the JS array spec where array[-1] returns the last element).

martineckardt:
> My only question is if it matters than the .json files here aren't valid JSON due to the comment at the top and the "...", as a value.

@michaelkaplan13 Had a convo about this topic with @cam-schultz yesterday. I created the template so wherever possible we have a valid example value. The only field where this is not possible and the user needs to change the values are `reward-address`, `account-private-key` and `start-block-height`.

For `reward-address` and `account-private-key` I was able to use valid JSON values that would create a clear error message when the config is used to run a relayer (something with malformed address / private key).

With start-block-height it is a bit more tricky. I saw the following options:
- Use 0 as the example value: Results in an [error](https://github.com/ava-labs/awm-relayer/blob/e4cab3998b1a77ba991774602443a6276a12c195/relayer/relayer.go#L272-L273)
- Use some valid number (e.g. 30471060): Results in the relayer processing many blocks at start up, since the delta between the head and the example value grows over time
- Use "..." to make clear that the user has to replace it with a valid value. I added instructions for this to the [academy course](https://academy.avax.network/path-player?courseid=teleporter&unit=65e642df775772eb460509a8Unit) 

I understand you guys had this [conversation](https://github.com/ava-labs/awm-relayer/pull/114#discussion_r1442257670) in January. That being said, I would advocate for some kind of way to start with the head to simplify this and avoid having the people to go to the explorer and fetch the latest block height manually either by making the value optional or including a value to indicate that it should start at the chain head (maybe -1, analogous to the JS array spec where array[-1] returns the last element).

geoff-vball:
@martineckardt I think the discussion you linked isn't exactly what we finally implemented. We only error if `start-block-height` is `true` AND `process-missed-blocks` in the top-level config is `true`. This is to make it clear to the user that if they set `process-missed-blocks` to `true`, that they should set a starting block because we know they probably don't want to process from genesis. If `process-missed-blocks` is `false`, then we will start with the head, as long as we haven't set the head before.

Looking at this now, we may want to rename `process-missed-blocks` to `process-historical-blocks`, which would behave the same as now, but I believe that we may always want to process "missed" blocks, ie blocks missed between a restart. Right now we have no way for someone to start at the chain head, and then also process any missed blocks during a shutdown without modifying their config. @cam-schultz @michaelkaplan13

config/config.go
```diff
@@ -103,7 +101,6 @@ type Config struct {
 func SetDefaultConfigValues(v *viper.Viper) {
 	v.SetDefault(LogLevelKey, logging.Info.String())
 	v.SetDefault(StorageLocationKey, "./.awm-relayer-storage")
-	v.SetDefault(ProcessMissedBlocksKey, true)
```
cam-schultz, line 106: Please revert this.

martineckardt: Sorry, must have happened with the merge. Added back with the default value changed to false

geoff-vball: We're actually going to keep these key to default to true. I'm about to open a PR to fix your issue a slightly different way.

martineckardt:
@michaelkaplan13 Merged the config changes and removed the invalid JSON from the sample config

cam-schultz (approved):

michaelkaplan13 (approved):
🚀

PR #206, opened 2024-03-01
Cleanup config

cam-schultz:
## Why this should be merged
Simplifies the relayer configuration by making the following changes:
- Removes the `network-id` field, instead fetching it from the info API.
- Add new `info-api-url` field instead of reusing `p-chain-api-url`. This removes the requirement that same node has both APIs enabled.
- Removes endpoint construction entirely, instead requiring the RPC and WS endpoints to be fully specified in the configuration. The current approach of constructing endpoints from the provided host, port, and encryption connection was error prone and didn't give the user with any additional functionality that couldn't be attained by providing a fully specified endpoint.

## How this works
- Removes Config.NetworkID
- Adds Config.InfoAPIURL
- Removes Config.EncryptConnection
- Removes SourceSubnet.EncryptConnection, SourceSubnet.APINodeHost, and SourceSubnet.APINodePort
- Removes DestinationSubnet.EncryptConnection, DestinationSubnet.APINodeHost, and DestinationSubnet.APINodePort

## How this was tested
CI

## How is this documented
Updated README


geoff-vball (dismissed):

README.md
```diff
@@ -90,25 +90,22 @@ The relayer binary accepts a path to a JSON configuration file as the sole argum
...
 `"p-chain-api-url": string`
 
 - The URL of the Avalanche P-Chain API node to which the relayer will connect. This API node needs to have the following methods enabled:
-  - info.peers
   - platform.getHeight
   - platform.validatedBy
   - platform.getValidatorsAt
 
-`"encrypt-connection": boolean`
+`"info-api-url": string`
```
minghinmatthewlam, line 104: i'm not too familiar with the info api nodes, is it common for the p chain api node to have info api enabled as well?

minghinmatthewlam: thinking if worth have more documentation about how to find their info api node, or whether to add a comment about testing with the p chain api node

cam-schultz: It's entirely up to the node operator which APIs to support. The P-Chain and Info APIs are independent, so we shouldn't assume that they'll both be enabled on the same node here.

michaelkaplan13: The public APIs support both the info endpoints and most of the P-chain endpoints (with the exception of `getValidatorsAt`. 

```
$ curl 'https://api.avax-test.network/ext/info' \
> --header 'Content-Type: application/json' \
> --data '{
>     "jsonrpc":"2.0",
>     "id"     :1,
>     "method" :"info.getNetworkID"
> }'
{"jsonrpc":"2.0","result":{"networkID":"5"},"id":1}
```

I probably would have gone the route of having a single `avalancheAPIURL` configuration option and adding the path extensions as needed in the application, but not opposed to having these split up in the event that the node used for the P-chain API (possibly private node) doesn't have the Info API enabled (all public APIs have the Info API available).

michaelkaplan13 (dismissed):

README.md
cam-schultz: Yeah, in the case of the public API, both the info and P-chain API share the same base URL. That's not true in general though, so I think it's better to make as few assumptions here as we can.

michaelkaplan13: Yeah, totally agree. Thanks for elaborating.

minghinmatthewlam (dismissed):

README.md
minghinmatthewlam: is there any additional comment/documentation we can add for finding the info api, or is this relatively understood/straightforward for node operators?

geoff-vball (approved):

peers/app_request_network.go
```diff
@@ -57,9 +56,20 @@ func NewNetwork(
...
+			"Failed to get network ID",
+			zap.Error(err),
+		)
+		return nil, nil, err
+	}
+
 	if networkID != constants.MainnetID &&
 		networkID != constants.FujiID &&
-		len(APINodeURL) == 0 {
+		InfoAPINodeURL == "" {
```
michaelkaplan13, line 72: I think `InfoAPINodeURL` being empty would cause an error above when it's used to get the network ID, so this case isn't possible. Do you know why it was required originally?

scripts/build_local_image.sh
```diff
@@ -19,8 +19,13 @@ source "$RELAYER_PATH"/scripts/versions.sh
 full_commit_hash="$(git --git-dir="$RELAYER_PATH/.git" rev-parse HEAD)"
 commit_hash="${full_commit_hash::8}"
 
+./scripts/build.sh
```
michaelkaplan13, line 22: Was this broken previously? Seems unrelated to the rest of this PR

peers/app_request_network.go
cam-schultz: Good catch, we will require this url be non-empty in all cases since it's used to fetch the network ID.

Looking at it again, this check is no longer valid. In an older release, the Avalanchego test network had a set of hard-coded node IDs for Fuji and mainnet, so we didn't need to explicitly connect to them here. That's no longer the case, so we need the info API to fetch the peers in all cases.

scripts/build_local_image.sh
cam-schultz: Sorry meant to add a comment here explaining this. We removed the build stage from the Dockerfile [here](https://github.com/ava-labs/awm-relayer/pull/188/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557), in favor of goreleaser's approach which builds the binary outside of the image. This change should have been included in that PR.

geoff-vball (approved):

relayer/message_relayer.go
```diff
@@ -145,7 +145,9 @@ func (r *messageRelayer) relayMessage(unsignedMessage *avalancheWarp.UnsignedMes
 // will need to be accounted for here.
 func (r *messageRelayer) createSignedMessage() (*avalancheWarp.Message, error) {
 	r.relayer.logger.Info("Fetching aggregate signature from the source chain validators via API")
-	warpClient, err := warpBackend.NewClient(r.relayer.apiNodeURI, r.relayer.sourceBlockchainID.String())
+	// TODO: To properly support this, we should provide a dedicated Warp API endpoint in the config
```
michaelkaplan13, line 148: Can you create an issue for this if it doesn't already exist?

cam-schultz: Yup, this is already captured here: https://github.com/ava-labs/awm-relayer/issues/119

michaelkaplan13 (approved):

geoff-vball (approved):

README.md
minghinmatthewlam: makes sense, I agree

minghinmatthewlam (approved):

PR #211, opened 2024-03-04
add api node to network if it is a validator

felipemadero:
## Why this should be merged
When the so called API node is inside the validator set for the source blockchain, it is not added itself as it is not a peer of himself (only peers to API node are added). In situations were that validator is needed to fully sign a msg, teleporter can't proceed beyond relayer step. 

## How this works
The API node adds himself is belonging to the validators set

## How this was tested
It was tested on teleporter deploys on fuji for subnets of only one validator.

## How is this documented

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,9 @@ See the [Building](#building) section for instructions on how to build the relay
     - platform.getHeight
     - platform.validatedBy
     - platform.getValidatorsAt OR platform.getCurrentValidators
+  - If the P-Chain API node is also a subnet validator, it must have enabled:
+    - info.getNodeID
+    - info.getNodeIP
 
 ## Usage
 
--- a/peers/app_request_network.go
+++ b/peers/app_request_network.go
@@ -254,7 +254,7 @@ func (n *AppRequestNetwork) ConnectPeers(nodeIDs set.Set[ids.NodeID]) (set.Set[i
 					zap.String("nodeIP", apiNodeIP),
 					zap.Error(err),
 				)
-				retErr = fmt.Errorf("failed to parse api Node IP: %v", err)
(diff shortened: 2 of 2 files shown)


cam-schultz (commented):
Thanks for bringing this up, we should definitely account for this case.

However, this is incompatible with the public API, as info.GetNodeID and info.GetNodeIP are disabled. Instead, I think we should move your changes below the existing peer connection loop like so:

```
get the list of peers
foreach peer {
    attempt to connect to peer
}

// new code starts here
if len(connectedPeers) < len(nodeIDs) {
    attempt manually connect to API node
}

```

If the relayer is configured to use the public API, then the peer connection loop should succeed, and we will not call info.GetNodeID / info.GetNodeIP. Otherwise, if the info API node is also a validator, it's required for those two methods to be enabled.

Can you please make this change, as well as document the updated API requirements in the README? Thanks.

felipemadero:
> Thanks for bringing this up, we should definitely account for this case.
> 
> However, this is incompatible with the public API, as info.GetNodeID and info.GetNodeIP are disabled. Instead, I think we should move your changes below the existing peer connection loop like so:
> 
> ```
> get the list of peers
> foreach peer {
>     attempt to connect to peer
> }
> 
> // new code starts here
> if len(connectedPeers) < len(nodeIDs) {
>     attempt manually connect to API node
> }
> ```
> 
> If the relayer is configured to use the public API, then the peer connection loop should succeed, and we will not call info.GetNodeID / info.GetNodeIP. Otherwise, if the info API node is also a validator, it's required for those two methods to be enabled.
> 
> Can you please make this change, as well as document the updated API requirements in the README? Thanks.

done

cam-schultz (approved):
LGTM

geoff-vball (approved):

PR #212, opened 2024-03-04
Rename subnet -> blockchain in Config

### LINKED ISSUE #170: Key the configuration by blockchain, not subnet
**Context and scope**
The two primary types that compose `Config` are `DestinationSubnet` and `SourceSubnet`, each of which must specify a unique blockchain ID, but the subnet ID may be non-unique within each list. This is because subnets can support multiple blockchains, but a blockchain can only ever belong to a single subnet.

To reflect this, we should change these types to `DestinationBlockchain` and `SourceBlockchain` to better reflect what each entry represents.

cam-schultz:
## Why this should be merged
Fixes #170 

The application requires unique Blockchain IDs in the configuration, but the naming suggests that Subnets also need to be unique. Subnets can contain multiple blockchains, in which case there would need to be multiple entries in `SourceSubnets` and `DestinationSubnets` with the same subnet ID.

## How this works
Renames `SourceSubnet` to `SourceBlockchain` and `DestinationSubnet` to `DestinationBlockchain` in the configuration.

## How this was tested
CI

## How is this documented
Updated README

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,9 +139,9 @@ The relayer is configured via a JSON file, the path to which is passed in via th
 
   - The address of the destination account that will receive the Warp message.
 
-`"source-subnets": []SourceSubnets`
+`"source-blockchains": []SourceBlockchains`
 
-- The list of source subnets to support. Each `SourceSubnet` has the following configuration:
+- The list of source subnets to support. Each `SourceBlockchain` has the following configuration:
 
   `"subnet-id": string`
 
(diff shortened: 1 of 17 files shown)


geoff-vball:
Thanks for doing this! Just want to link the ticket we had for this issue here https://github.com/ava-labs/awm-relayer/issues/170

michaelkaplan13 (approved):

geoff-vball (approved):

minghinmatthewlam (approved):

PR #213, opened 2024-03-05
Check connected peer weight

cam-schultz:
## Why this should be merged
Currently, we abandon message relaying if we fail to connect to any of the subnet validators. However, the aggregate signature only needs a quorum of stake weight to be valid, so this constraint is overly restrictive.

This change relaxes that constraint by attempting to connect to all peers, then comparing the total connected stake against the quorum value of the destination blockchain.

## How this works
`AppRequestNetwork.ConnectPeers` returns the set of connected peers, and no error. It is up to the caller to determine if the returned value constitutes an error. The message relayer uses the connected stake to determine whether or not to proceed with relaying.

## How this was tested
CI

## How is this documented
N/A

### DIFF
--- a/peers/app_request_network.go
+++ b/peers/app_request_network.go
@@ -183,21 +183,15 @@ func NewNetwork(
 }
 
 // ConnectPeers connects the network to peers with the given nodeIDs.
-// On success, returns the provided set of nodeIDs and a nil error.
-// On failure, returns the set of nodeIDs that successfully connected and an error.
-func (n *AppRequestNetwork) ConnectPeers(nodeIDs set.Set[ids.NodeID]) (set.Set[ids.NodeID], error) {
+// Returns the set of nodeIDs that were successfully connected to.
+func (n *AppRequestNetwork) ConnectPeers(nodeIDs set.Set[ids.NodeID]) set.Set[ids.NodeID] {
 	n.lock.Lock()
 	defer n.lock.Unlock()
-
-	var (
-		retErr       error
-		trackedNodes set.Set[ids.NodeID]
-	)
 
 	// First, check if we are already connected to all the peers
 	connectedPeers := n.Network.PeerInfo(nodeIDs.List())
 	if len(connectedPeers) == nodeIDs.Len() {
-		return nodeIDs, nil
+		return nodeIDs
 	}
 
 	// If we are not connected to all the peers already, then we have to iterate
@@ -212,10 +206,11 @@ func (n *AppRequestNetwork) ConnectPeers(nodeIDs set.Set[ids.NodeID]) (set.Set[i
 			"Failed to get peers",
 			zap.Error(err),
 		)
-		return nil, err
+		return nil
 	}
 
 	// Attempt to connect to each peer
+	var trackedNodes set.Set[ids.NodeID]
 	for _, peer := range peers {
 		if nodeIDs.Contains(peer.ID) {
 			ipPort, err := ips.ToIPPort(peer.PublicIP)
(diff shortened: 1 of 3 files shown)


geoff-vball (dismissed):

relayer/message_relayer.go
```diff
@@ -247,13 +247,19 @@ func (r *messageRelayer) createSignedMessageAppRequest(requestID uint32) (*avala
 	for node := range nodeValidatorIndexMap {
 		nodeIDs.Add(node)
 	}
+	connectedNodes := r.relayer.network.ConnectPeers(nodeIDs)
 
-	// TODO: We may still be able to proceed with signature aggregation even if we fail to connect to some peers.
-	// 		 We should check if the connected set represents sufficient stake, and continue if so.
-	_, err = r.relayer.network.ConnectPeers(nodeIDs)
-	if err != nil {
+	// Check if we've connected to a stake threshold of nodes
```
michaelkaplan13, line 252: May want to move this to a `checkSufficientStakeConnected` helper.

peers/app_request_network.go
```diff
@@ -225,43 +220,39 @@ func (n *AppRequestNetwork) ConnectPeers(nodeIDs set.Set[ids.NodeID]) (set.Set[i
...
 			n.Network.ManuallyTrack(peer.ID, ipPort)
 			if len(trackedNodes) == nodeIDs.Len() {
-				return trackedNodes, retErr
+				return trackedNodes
 			}
 		}
 	}
 
-	// attempt adding api node in case it is a validator
+	// If we haven't yet connected to all peers, attempt to connect to the API node in case it is a validator
```
michaelkaplan13, line 233: If the API node is a validator, wouldn't it's node ID already have been included in `nodeIDs`? I find it a little odd that the caller here gets to specify a set of node IDs to connect to but the implementation may add to that set.

cam-schultz: If we reach this point, the only thing we'd want to do is manually track the Info API node as a potential validator. The passed in list of `nodeIDs` is the full set of subnet validators as fetched from the P-Chain, so we should definitely not mutate that list. I think what's missing here are validation checks that the fetched Info API node ID is in fact in the validator set. I'll add that check in.

cam-schultz: @michaelkaplan13 let me know if this approach makes sense to you. We're not mutating the input list at all.

cam-schultz: Rewrote the comment to make it clear that we're checking for the case in which the API node is in `nodeIDs`

geoff-vball (approved):

peers/app_request_network.go
michaelkaplan13: Ah, yeah that makes sense, thanks.

relayer/message_relayer.go
cam-schultz: Modified this to use the existing `utils.CheckStakeWeightExceedsThreshold` helper. I don't think it's worth adding another helper just to calculate the connected stake though.

michaelkaplan13 (dismissed):

relayer/message_relayer.go
michaelkaplan13: Sounds good

relayer/message_relayer.go
```diff
@@ -247,13 +247,24 @@ func (r *messageRelayer) createSignedMessageAppRequest(requestID uint32) (*avala
...
+		totalValidatorWeight,
+		r.warpQuorum.QuorumNumerator,
+		r.warpQuorum.QuorumDenominator,
+	) {
 		r.relayer.logger.Error(
-			"Failed to connect to peers",
+			"Failed to connect to a threshold of stake",
+			zap.Uint64("connectedWeight", connectedWeight),
+			zap.Uint64("totalValidatorWeight", totalValidatorWeight),
+			zap.Any("warpQuorum", r.warpQuorum),
```
geoff-vball, line 267: `err` referenced on lines 268 and 270 is no longer referencing a relevant error.

cam-schultz: Fixed

geoff-vball (approved):

michaelkaplan13 (approved):

PR #218, opened 2024-03-05
Simplify options for process-historical-blocks

geoff-vball:
## Why this should be merged
Previously, we were using `process-missed-blocks` to determine whether to process historical AND missed blocks. This PR splits those out into two different config options.

## How this works
`process-historical-blocks` defaults to `false` will only work if `start-block-height` is set, and will start processing blocks from the greater of `start-block-height` and the last processed block height (if it exists in the db).

`process-missed-blocks` defaults to `true`, and will start processing blocks from the last processed block height (if it exists in the db). If there is no latest block in the db, or `process-missed-blocks` is `false`, we will start processing from the chain head.

## How is this documented
README

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -54,14 +54,14 @@ type ManualWarpMessage struct {
 }
 
 type SourceBlockchain struct {
-	SubnetID              string                           `mapstructure:"subnet-id" json:"subnet-id"`
-	BlockchainID          string                           `mapstructure:"blockchain-id" json:"blockchain-id"`
-	VM                    string                           `mapstructure:"vm" json:"vm"`
-	RPCEndpoint           string                           `mapstructure:"rpc-endpoint" json:"rpc-endpoint"`
-	WSEndpoint            string                           `mapstructure:"ws-endpoint" json:"ws-endpoint"`
-	MessageContracts      map[string]MessageProtocolConfig `mapstructure:"message-contracts" json:"message-contracts"`
-	SupportedDestinations []string                         `mapstructure:"supported-destinations" json:"supported-destinations"`
-	StartBlockHeight      uint64                           `mapstructure:"process-historical-blocks-from-height" json:"process-historical-blocks-from-height"`
+	SubnetID                          string                           `mapstructure:"subnet-id" json:"subnet-id"`
+	BlockchainID                      string                           `mapstructure:"blockchain-id" json:"blockchain-id"`
+	VM                                string                           `mapstructure:"vm" json:"vm"`
+	RPCEndpoint                       string                           `mapstructure:"rpc-endpoint" json:"rpc-endpoint"`
+	WSEndpoint                        string                           `mapstructure:"ws-endpoint" json:"ws-endpoint"`
+	MessageContracts                  map[string]MessageProtocolConfig `mapstructure:"message-contracts" json:"message-contracts"`
+	SupportedDestinations             []string                         `mapstructure:"supported-destinations" json:"supported-destinations"`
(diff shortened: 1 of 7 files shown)


martineckardt:
Idea: Why don't we remove the option `ProcessHistoricalBlocks` from the global config and process historical blocks if the `StartBlockHeight` optional value is set in the sourceBlockchain config?

If I understand correctly, if the `ProcessHistoricalBlocks` option is set to `true` then every sourceBlockchain needs a `StartBlockHeight`. This means the relayer operator cannot decide this on a blockchain-by-blockchain basis. 

With omitting `ProcessHistoricalBlocks` from the global config we additionally eliminate the scenario where a relayer operator set's the `StartBlockHeight`, but forgets to also set `ProcessHistoricalBlocks` option is set to `true`. 

To make it more intuitive, I would rename the `"start-block-height"` option to something like `"process-historical-blocks-from-height"`.

geoff-vball:
@martineckardt I did actually just have that thought as I pushed the last update here. I'll make that change now.

config/config.go
```diff
@@ -63,7 +61,7 @@ type SourceBlockchain struct {
 	WSEndpoint            string                           `mapstructure:"ws-endpoint" json:"ws-endpoint"`
 	MessageContracts      map[string]MessageProtocolConfig `mapstructure:"message-contracts" json:"message-contracts"`
 	SupportedDestinations []string                         `mapstructure:"supported-destinations" json:"supported-destinations"`
-	StartBlockHeight      uint64                           `mapstructure:"start-block-height" json:"start-block-height"`
+	StartBlockHeight      uint64                           `mapstructure:"process-historical-blocks-from-height" json:"process-historical-blocks-from-height"`
```
cam-schultz, line 64: Let's update this variable name

relayer/relayer.go
```diff
@@ -174,7 +169,7 @@ func NewRelayer(
 		// ProcessFromHeight to overload the message queue and cause a deadlock.
 		go sub.ProcessFromHeight(big.NewInt(0).SetUint64(height), r.catchUpResultChan)
 	} else {
-		err = r.setProcessedBlockHeightToLatest()
+		_, err = r.setProcessedBlockHeightToLatest()
```
cam-schultz, line 172: Can we add a log here that we're initializing the relayer from the latest height?

relayer/relayer.go
```diff
@@ -306,15 +303,15 @@ func (r *Relayer) calculateStartingBlockHeight(startBlockHeight uint64) (uint64,
 	return startBlockHeight, nil
 }
 
-func (r *Relayer) setProcessedBlockHeightToLatest() error {
+func (r *Relayer) setProcessedBlockHeightToLatest() (uint64, error) {
```
cam-schultz, line 306: Can you add a doccomment to this function explaining the return values?

relayer/relayer.go
```diff
@@ -174,7 +169,11 @@ func NewRelayer(
 		// ProcessFromHeight to overload the message queue and cause a deadlock.
 		go sub.ProcessFromHeight(big.NewInt(0).SetUint64(height), r.catchUpResultChan)
 	} else {
-		err = r.setProcessedBlockHeightToLatest()
+		r.logger.Info(
+			"processed-missed-blocks set to false, starting process from chain head",
```
cam-schultz, line 173:
```suggestion
			"processed-missed-blocks set to false, starting processing from chain head",
```

cam-schultz (approved):
LGTM

michaelkaplan13 (approved):
Makes sense to me

Issue #193, opened 2024-02-22
Validate peer version


cam-schultz:
**Context and scope**
If there is a peer mismatch between the relayer and subnet validators, then app requests to fetch signatures will fail. We should explicitly check peer compatibility on startup to more clearly signal a mismatch.

**Discussion and alternatives**
The check should take into consideration the subnet's Warp stake threshold. For example, if a subnet's Warp quorum numerator is 67, then we only need 67% of validator peers to be reachable.

**Open questions**
There's no RPC to fetch the peer version directly, so we'll need to decide the best way to detect a mismatch. Some ideas:
- Count P2P connection errors
- Fetch the node version from the info API, and compare against the minor version run by the relayer (bug fix versions should all be compatible)

felipemadero:
In relation with CLI, it may be also desirable to give CLI or other automating tools a way to decide on the compatibility of a given relayer with a given avalanchego. Subnet-evm has something like this here https://github.com/ava-labs/subnet-evm/blob/master/compatibility.json which we can use together with https://github.com/ava-labs/avalanchego/blob/master/version/compatibility.json to decide which avago to use in relation with a given subnet.
In the same direction, we may want to know which relayer to use in relation with a given avago version, so maybe you can add a compability file to your repo with the map between relayer and avago versions.

cam-schultz:
> In relation with CLI, it may be also desirable to give CLI or other automation tools a way to decide on the compatibility of a given relayer with a given avalanchego. Subnet-evm has something like this here https://github.com/ava-labs/subnet-evm/blob/master/compatibility.json which we can use together with https://github.com/ava-labs/avalanchego/blob/master/version/compatibility.json to decide which avago to use in relation with a given vm. In the same direction, we may want to know which relayer to use in relation with a given avago version, so maybe you can add a compability file to your repo with the map between relayer and avago versions.

I agree it should be possible to do this filtering ahead of runtime. Rather than a compatibility manifest, we should be able to use the avalanchego dependency version in go.mod to determine compatibility. The relayer's P2P networking layer is built on top of tooling hosted in avalanchego, so the problem is the same as determining if a given avalanchego node is compatible at the P2P layer with its peers.

I'd like to maintain a single source of truth regarding dependencies as much as possible.

felipemadero:
[...]

PR #231, opened 2024-03-13
Validate peers on startup

cam-schultz:
## Why this should be merged
Fixes https://github.com/ava-labs/awm-relayer/issues/193

## How this works
Manually connects to subnet validators on startup, rather than at message relay time. Each subnet must be connected to enough stake that any of the configured destinations would be able to validate Warp messages sent by the subnet, otherwise the relayer exits. We still need to attempt to connect to peers at message relay time, since the validator set may have changed. This change is more intended to surface any issues earlier, rather than optimizing performance.

In addition to moving this check to startup, this change also refactors some of the peer connection and signature management logic into common helpers.

## How this was tested
CI

## How is this documented
N/A


relayer/message_relayer.go
```diff
@@ -483,55 +465,6 @@ func (r *messageRelayer) createSignedMessageAppRequest(requestID uint32) (*avala
 	return nil, errNotEnoughSignatures
 }
 
-func (r *messageRelayer) getCurrentCanonicalValidatorSet() ([]*avalancheWarp.Validator, uint64, error) {
```
cam-schultz, line 486: Signing subnet check moved to `messageRelayer` constructor; the rest of this function moved to `CanonicalValidatorClient.GetCurrentCanonicalValidatorSet`

relayer/message_relayer.go
```diff
@@ -211,60 +218,35 @@ func (r *messageRelayer) createSignedMessage() (*avalancheWarp.Message, error) {
...
 		r.relayer.logger.Error(
-			"Failed to get the canonical subnet validator set",
-			zap.String("subnetID", r.relayer.sourceSubnetID.String()),
+			"Failed to connect to canonical validators",
 			zap.Error(err),
 		)
 		return nil, err
 	}
-
-	// We make queries to node IDs, not unique validators as represented by a BLS pubkey, so we need this map to track
```
cam-schultz, line 234: This logic was moved to `AppRequestNetwork.ConnectToCanonicalValidators`

relayer/message_relayer.go
```diff
@@ -156,18 +175,6 @@ func (r *messageRelayer) createSignedMessage() (*avalancheWarp.Message, error) {
 		)
 		return nil, err
 	}
-	signingSubnetID := r.relayer.sourceSubnetID
```
cam-schultz, line 159: This was moved to the `messageRelayer` constructor

config/config.go
```diff
@@ -97,10 +103,6 @@ type Config struct {
 	DestinationBlockchains []*DestinationBlockchain `mapstructure:"destination-blockchains" json:"destination-blockchains"`
 	ProcessMissedBlocks    bool                     `mapstructure:"process-missed-blocks" json:"process-missed-blocks"`
 	ManualWarpMessages     []*ManualWarpMessage     `mapstructure:"manual-warp-messages" json:"manual-warp-messages"`
-
-	// convenience fields to access the source subnet and chain IDs after initialization
-	sourceSubnetIDs     []ids.ID
-	sourceBlockchainIDs []ids.ID
```
cam-schultz, line 103: These were only used to pass to the app request network constructor. With this change, it makes more sense to pass the configuration and iterate over the `SourceBlockchain` and `DestinationBlockchain` slices directly, so the parsed IDs were moved there.

peers/app_request_network.go
```diff
@@ -103,83 +102,99 @@ func NewNetwork(
 		return nil, nil, err
 	}
 
-	// We need to initially connect to some nodes in the network before peer
```
cam-schultz, line 106: There's no need to connect to these beacon nodes, since we explicitly connect to every node that we send `AppRequest` messages to now.

config/config.go
```diff
@@ -417,6 +401,18 @@ func (s *SourceBlockchain) Validate(destinationBlockchainIDs *set.Set[string]) e
 		}
 	}
 
+	// Validate and store the subnet and blockchain IDs for future use
```
michaelkaplan13, line 404: Not for this PR, but we should consider splitting `config.go` up into multiple files at some point soon. I think `SourceBlockchain` and `DestinationBlockchain` could each be put in their own file still in the `config` package and could help improve navigation/readability.

database/json_file_storage.go
```diff
@@ -34,30 +34,30 @@ type JSONFileStorage struct {
 }
 
 // NewJSONFileStorage creates a new JSONFileStorage instance
-func NewJSONFileStorage(logger logging.Logger, dir string, networks []ids.ID) (*JSONFileStorage, error) {
+func NewJSONFileStorage(logger logging.Logger, dir string, chainIDs []ids.ID) (*JSONFileStorage, error) {
```
michaelkaplan13, line 37: Nit: Can we name it `blockchainIDs` instead of `chainIDs` to avoid possible confusion?

main/main.go
```diff
@@ -167,7 +174,12 @@ func main() {
 	}
 
 	// Initialize the database
-	db, err := database.NewJSONFileStorage(logger, cfg.StorageLocation, sourceBlockchainIDs)
+	sourceChainIDs := make([]ids.ID, len(cfg.SourceBlockchains))
```
michaelkaplan13, line 177: Same request here (`sourceBlockchainIDs`)

michaelkaplan13: Actually, could we just pass `cfg.SourceBlockchains` directly to `NewJSONFileStorage`? I don't think it's necessary to do an iteration to extract the blockchain ID values, just to then iterate over the new slice.

peers/app_request_network.go
```diff
@@ -103,83 +100,99 @@ func NewNetwork(
...
-		if len(peers[index].PublicIP) == 0 {
-			continue
-		}
-		beaconIPs = append(beaconIPs, peers[index].PublicIP)
-		beaconIDs = append(beaconIDs, peers[index].ID.String())
-		if len(beaconIDs) == numInitialTestPeers {
-			break
+	// Manually connect to the validators of each of the source subnets.
+	// We return an error if we are unable to connect to sufficient stake on any of the subnets.
+	// Sufficient stake is determined by the Warp quora of the configured supported destinations,
```
michaelkaplan13, line 113: TIL that "quora" is a word. 😅

peers/app_request_network.go
```diff
@@ -103,83 +100,99 @@ func NewNetwork(
...
+	// or if the subnet supports all destinations, by the quora of all configured destinations.
+	for _, sourceBlockchain := range cfg.SourceBlockchains {
```
michaelkaplan13, line 115: This loop body is pretty large. May want to consider moving parts out to helper functions.

peers/app_request_network.go
```diff
@@ -103,83 +100,99 @@ func NewNetwork(
...
+			quorum, err := cfg.GetWarpQuorum(destinationBlockchainID)
+			if err != nil {
+				logger.Error(
+					"Failed to get warp quorum from config",
+					zap.Error(err),
+				)
+				return nil, nil, err
+			}
+			if !utils.CheckStakeWeightExceedsThreshold(
+				big.NewInt(0).SetUint64(connectedValidators.ConnectedWeight),
```
michaelkaplan13, line 175: In the case that the source network is the C-Chain (or generally a chain in the primary network), we assigned `connectedValidators` multiple times in a nested for loop above, but I think only check the quorom here against the final assignment value.

peers/app_request_network.go
michaelkaplan13: Would also be great to add unit tests for this since the logic is relatively complex. I'm not sure how difficult it is to mock the various clients and `arNetwork` though.

config/config.go
minghinmatthewlam: agreed and some of the similarities between the source/destination can be abstracted

peers/app_request_network.go
```diff
@@ -103,83 +100,99 @@ func NewNetwork(
...
-		beaconIPs = append(beaconIPs, peers[index].PublicIP)
-		beaconIDs = append(beaconIDs, peers[index].ID.String())
-		if len(beaconIDs) == numInitialTestPeers {
-			break
+	// Manually connect to the validators of each of the source subnets.
+	// We return an error if we are unable to connect to sufficient stake on any of the subnets.
+	// Sufficient stake is determined by the Warp quora of the configured supported destinations,
+	// or if the subnet supports all destinations, by the quora of all configured destinations.
+	for _, sourceBlockchain := range cfg.SourceBlockchains {
+		// Get the list of destination chains that this source may relay to
```
minghinmatthewlam, line 116: do we want to differentiate `chains` and clarify as blockchain

peers/app_request_network.go
```diff
@@ -103,83 +100,99 @@ func NewNetwork(
...
+		var destinationBlockchainIDs []ids.ID
+		if supportedDsts := sourceBlockchain.GetSupportedDestinations(); supportedDsts.Len() > 0 {
+			destinationBlockchainIDs = supportedDsts.List()
+		} else {
+			for _, dst := range cfg.DestinationBlockchains {
```
minghinmatthewlam, line 121: can you add a comment about if `supported-destination` config si not set, takes all destinations

peers/app_request_network.go
```diff
@@ -103,83 +100,99 @@ func NewNetwork(
...
+			if !utils.CheckStakeWeightExceedsThreshold(
+				big.NewInt(0).SetUint64(connectedValidators.ConnectedWeight),
+				connectedValidators.TotalValidatorWeight,
+				quorum.QuorumNumerator,
+				quorum.QuorumDenominator,
+			) {
+				logger.Error(
+					"Failed to connect to a threshold of stake",
+					zap.Uint64("connectedWeight", connectedValidators.ConnectedWeight),
+					zap.Uint64("totalValidatorWeight", connectedValidators.TotalValidatorWeight),
```
minghinmatthewlam, line 183: logging the blockchains

peers/app_request_network.go
```diff
@@ -103,83 +100,99 @@ func NewNetwork(
...
+					zap.Any("warpQuorum", quorum),
+				)
+				return nil, nil, errors.New("failed to connect to a threshold of stake")
+			}
```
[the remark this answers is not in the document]
minghinmatthewlam: when verifying on destination you compare against the sender's stake weight, but you require a threshold of the dsetination's warp quorum

peers/app_request_network.go
```diff
@@ -103,83 +100,99 @@ func NewNetwork(
...
+				subnetID, err := pChainClient.ValidatedBy(context.Background(), destinationBlockchainID)
+				if err != nil {
+					logger.Error(
+						"Failed to get subnet ID",
+						zap.Error(err),
+					)
+					return nil, nil, err
+				}
+				connectedValidators, err = arNetwork.ConnectToCanonicalValidators(
+					validators.NewCanonicalValidatorClient(logger, pChainClient),
```
minghinmatthewlam, line 139: Does the `CanonicalValidatorClient` need to be newly created for each loop? Seems like we can move this out of the for loop

minghinmatthewlam: or actually we can save `pChainClient` in `AppRequestNetwork` struct and don't need to pass that value in for `ConnectToCanonicalValidators`

peers/app_request_network.go
```diff
@@ -257,3 +270,62 @@ func (n *AppRequestNetwork) ConnectPeers(nodeIDs set.Set[ids.NodeID]) set.Set[id
 
 	return trackedNodes
 }
+
+// Helper struct to hold connected validator information
+// Warp Validators sharing the same BLS key may consist of multiple nodes,
+// so we need to track the node ID to validator index mapping
+type ConnectedCanonicalValidators struct {
```
minghinmatthewlam, line 277: I think we should move this struct to a separate file, and ideally have mockign tests

config/config.go
cam-schultz: > some of the similarities between the source/destination can be abstracted

Viper doesn't play nicely with this unfortunately. There may be a workaround, but I haven't looked to deeply into it.

relayer/message_relayer.go
```diff
@@ -71,10 +73,27 @@ func newMessageRelayer(
 		)
 		return nil, err
 	}
+	var signingSubnet ids.ID
+	if relayer.sourceSubnetID == constants.PrimaryNetworkID {
+		// If the message originates from the primary subnet, then we instead "self sign" the message using the validators of the destination subnet.
+		signingSubnet, err = relayer.pChainClient.ValidatedBy(context.Background(), destinationBlockchainID)
```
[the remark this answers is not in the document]
cam-schultz: Message relayers are coupled to the source blockchain config, not the destination. We get the destination blockchain from the Teleporter message itself, but then have to look up the corresponding destination subnet.

peers/app_request_network.go
cam-schultz: Good calls. Fixed.

peers/app_request_network.go
cam-schultz: Sure I've updated this comment

main/main.go
```diff
@@ -167,7 +174,12 @@ func main() {
...
+	for i, sourceBlockchain := range cfg.SourceBlockchains {
+		sourceChainIDs[i] = sourceBlockchain.GetBlockchainID()
+	}
+
+	db, err := database.NewJSONFileStorage(logger, cfg.StorageLocation, sourceChainIDs)
```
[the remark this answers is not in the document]
cam-schultz: No longer an issue since we now pass cfg.SourceBlockchains to the ctr

peers/app_request_network.go
```diff
@@ -103,83 +100,99 @@ func NewNetwork(
...
+			for _, destinationBlockchainID := range destinationBlockchainIDs {
+				subnetID, err := pChainClient.ValidatedBy(context.Background(), destinationBlockchainID)
+				if err != nil {
+					logger.Error(
+						"Failed to get subnet ID",
+						zap.Error(err),
+					)
+					return nil, nil, err
+				}
+				connectedValidators, err = arNetwork.ConnectToCanonicalValidators(
```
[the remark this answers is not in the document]
cam-schultz: This is fixed now.

peers/app_request_network.go
cam-schultz: This is fixed now.

relayer/message_relayer.go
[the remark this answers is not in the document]
geoff-vball: Yeah, I'm just saying we could look up the destination subnet because all the pairings are set in the config to avoid the API call. If it seems like more trouble than it's worth that's okay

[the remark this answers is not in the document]
cam-schultz: Oh, gotcha. I misunderstood your original message. 

That makes a lot of sense, and reducing the number of RPC calls is an easy optimization. I'll include that in this PR.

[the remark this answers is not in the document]
cam-schultz: Added

peers/app_request_network.go
```diff
@@ -103,83 +101,105 @@ func NewNetwork(
...
-		beaconIDs = append(beaconIDs, peers[index].ID.String())
-		if len(beaconIDs) == numInitialTestPeers {
-			break
+	// Manually connect to the validators of each of the source subnets.
+	// We return an error if we are unable to connect to sufficient stake on any of the subnets.
+	// Sufficient stake is determined by the Warp quora of the configured supported destinations,
+	// or if the subnet supports all destinations, by the quora of all configured destinations.
+	for _, sourceBlockchain := range cfg.SourceBlockchains {
+		// Get the list of destination blockchains that this source may relay to.
+		// If supported destinations are not specified, use the configured destinations.
```
michaelkaplan13, line 121: I think this case is handled by the `config` package already, so just calling `GetSupportedDestinations` is sufficient here in both cases.

peers/app_request_network.go
cam-schultz: Eh, I don't think it's worth all that special treatment since this type is POD. It's really just there to simplify calls to `ConnectToCanonicalValidators` and keep associated data packaged together.

peers/app_request_network.go
```diff
@@ -103,83 +100,99 @@ func NewNetwork(
...
+		if sourceBlockchain.GetBlockchainID() == constants.PrimaryNetworkID {
+			// For the primary network, connect to the validators of the destination chains
+			for _, destinationBlockchainID := range destinationBlockchainIDs {
+				subnetID, err := pChainClient.ValidatedBy(context.Background(), destinationBlockchainID)
+				if err != nil {
+					logger.Error(
+						"Failed to get subnet ID",
+						zap.Error(err),
+					)
+					return nil, nil, err
```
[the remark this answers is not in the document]
cam-schultz: Agreed that we can have a more unified logging / error handling strategy. I also agree that we should go through and unify the repo once we decide on that strategy. In the meantime, however, I'd like to err on the side of over-verbosity and log where we can, even if some logs are redundant.

peers/app_request_network.go
```diff
@@ -103,83 +101,105 @@ func NewNetwork(
...
-				"Failed to parse beaconID",
-				zap.String("beaconID", beaconIDStr),
-				zap.Error(err),
-			)
-			return nil, nil, err
+		var connectedValidators *ConnectedCanonicalValidators
+		// If the source blockchain is the primary network, connect to the validators of
+		// the destination blockchains in the following loop.
+		// Otherwise, connect to the validators of the source blockchain
+		if sourceBlockchain.GetSubnetID() != constants.PrimaryNetworkID {
```
michaelkaplan13, line 135: Could we simplify this logic by moving the case for the primary network and the case for non-primary networks out to helper functions?

For non-primary sources, we:
- connect to source
- iterate over each destination to check the quorum value

For primary sources we:
- iterate over each destination and create connections
- check the connection reaches a quorum for that destination

peers/app_request_network.go
cam-schultz: Great spot. Fixed this now

peers/app_request_network.go
michaelkaplan13: We could also add in a `checkSufficientStakeForDestination` helper used by both of those that takes a `*ConnectedCanonicalValidators` and a single destination chain ID.

cam-schultz: Sure thing. Added these helpers.

peers/app_request_network.go
```diff
@@ -257,3 +270,62 @@ func (n *AppRequestNetwork) ConnectPeers(nodeIDs set.Set[ids.NodeID]) set.Set[id
 
 	return trackedNodes
 }
+
+// Helper struct to hold connected validator information
+// Warp Validators sharing the same BLS key may consist of multiple nodes,
+// so we need to track the node ID to validator index mapping
+type ConnectedCanonicalValidators struct {
+	ConnectedWeight       uint64
+	TotalValidatorWeight  uint64
```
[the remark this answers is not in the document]
cam-schultz: I think those ideas make sense, but I'm going to defer that for later. I think we'll likely want to add an async routine to update the validator set, rather than fetching it in every message relay, at which point we'll want to carefully consider how the global state is managed. For now though, this type is only used in the message relayer, so the risk of it being misused is fairly low.

michaelkaplan13 (approved):

geoff-vball (approved):

Issue #234, opened 2024-03-14
Add periodic database write routine


cam-schultz:
**Context and scope**
Currently, we write the checkpointed block to the database after every processed Warp message. This approach has unbounded write frequency in highly concurrent loads, especially when considering future changes such as https://github.com/ava-labs/awm-relayer/issues/225, which may result in many relayer instances sharing a single database. 

**Discussion and alternatives**
In reality, writing on every processed Warp message is overkill, since the catch-up mechanism is intelligent enough to not attempt to re-broadcast already delivered Warp messages (at least for Teleporter; future message protocol integrations should also take this approach - see https://github.com/ava-labs/awm-relayer/issues/233). Instead, we can get away with caching the checkpoint in memory, and writing to the database periodically. This period should be settable from the configuration, so that deployments can scale the figure with the number of relayer instances to prevent a write queue.

Issue #31, opened 2023-09-06
Process messages from the same chain concurrently


cam-schultz:
**Context and scope**
When adding bootstrapping logic and state in https://github.com/ava-labs/awm-relayer/pull/20, we identified issues with processing messages from the same source chain concurrently. The workaround was to process such messages in serial in order to provide stronger guarantees of block ordering within the application and database.

**Discussion and alternatives**
We should consider adding back in this concurrent processing of messages from the same source chain in order to potentially improve performance.

**Open questions**
Under what conditions will adding this concurrency improve performance? In any case, relayer message processing is capped by the rate at which messages are included in the source chain, so we should make sure this would yield performance improvements before pursuing, especially since the synchronization logic is likely to be complex.

PR #243, opened 2024-03-19
Support shared DB access

### LINKED ISSUE #240: Support shared database
**Context and scope**
An individual relayer instance's data footprint is relatively small. Rather than requiring a standalone database for each instance (especially when using an [external database](https://github.com/ava-labs/awm-relayer/issues/225)), we should support shared database access across a fleet of relayer instances. In order to do so, we need to consider the functional role that each individual instance fills, and key the database accordingly.

Currently, the DB key consists solely of the source blockchain ID. This becomes a problem in a fleet deployment scenario as soon as there are multiple relayer instances listening to the same blockchain, but configured to relay different groups of messages. For instance, relayers are able to be configured with different allowed destination blockchains, and are planned to be able to filter by [sender](https://github.com/ava-labs/awm-relayer/issues/80) and [destination](https://github.com/ava-labs/awm-relayer/issues/80) addresses. With the current DB key, relayers would not be guaranteed to be fetching data pertaining to their functional spec.

_Key choice_
To get around the functional limitation on the DB key described above, we should choose the key such that it uniquely identifies a relayer instance at the functional level. The proposed relayer key is `hash(sourceBlockchainID, destinationBlockchainID, sourceAddress, destinationAddress)`. That would provide relayer data uniqueness when considering all of these possible filters. We don't need to realistically worry about filtering by Teleporter version or sender EOA, since relayers are expected to be configured at the dApp level.

_Missed blocks_
On initialization, the relayer queries the database to fetch the block from which to start indexing. If no such entry is found, then the relayer uses the `process-historical-blocks-from-height` configuration parameter. If that is missing, the relayer skips catching up and begins processing from the current head. If both values are present, the relayer selects the maximum of the two.

How should we determine the starting block for new relayer keys? A conservative approach would be to take the minimum of all other keys on that source blockchain and the value of `process-historical-blocks-from-height` as the starting block. In that case, we would only start from the head if`process-historical-blocks-from-height` is missing, as well as _all_ values for keys pertaining to the particular source blockchain. The only cost here is that existing keys would re-process already delivered Warp messages (this is not a problem for Teleporter, since we exit early if a message has already been delivered. https://github.com/ava-labs/awm-relayer/issues/233 would mitigate the effect of this for other message protocol types)

For example, if a relayer currently is listening to blockchainA, and has checkpointed block 1000 in the database at a key1, adding a new key to the configuration on blockchainA and specifying `process-historical-blocks-from-height=500` would cause key1 to reprocess blocks 500-1000 on startup. If `process-historical-blocks-from-height` is omitted in this case, the new key will start processing from block 1000.

An alternative would be to perform catch up on a per-key basis, rather than on a per-source blockchain basis, as is currently done. This would involve more significant architectural changes, for an unclear benefit when compared to the first option.

_Application Relayers_

Currently, the relayer application is composed of one persistent listener (of type `Relayer`) per configured source blockchain, each of which spins up a new ephemeral `messageRelayer` to process an individual Warp message. With the proposed key changes, `Relayer` instantiations may result in writes to multiple keys, but in the current design, the actual key is not determined until the ephemeral `messageRelayer` is already processing the message.

This is not such a problem now, but could become an issue when we add [concurrent message handling](https://github.com/ava-labs/awm-relayer/issues/31), as we'd need to coordinate database writes across ephemeral `messageRelayer` instances.

A better design is to uniquely associate `messageRelayer` instantiations with a specific database key, and persist those instances to simplify coordination. Rather than handling a single message, the `messageRelayer` type would instead handle all message relays from a particular (source chain, destination chain, source address, destination address) tuple. The name "Application Relayer" better captures the function of this type.

**Open questions**
Questions that are still being discussed.

cam-schultz:
## Why this should be merged
Fixes #240 

## How this works
- `messageRelayer` instances are persisted, and now pertain to a unique (source chain, destination chain, source address, destination address) tuple
- `Relayer` instances are composed of a set of `messageRelayers`, rather than `messageRelayer` being composed of the parent `Relayer`
- Migrates database interfacing methods from `Relayer` to `messageRelayer`
- Database keys changed from `ids.ID` to `common.Hash`
- Adds `DestinationAddress` and `SourceAddress` getters to `MessageManager` to facilitate key construction
- Adds database key utilities

## How this was tested
- CI, updated tests
- Added multiple relayers, single DB e2e test
- TODO: New unit tests

## How is this documented
Updated README

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -16,7 +16,6 @@ import (
 	"github.com/ava-labs/avalanchego/utils/constants"
 	"github.com/ava-labs/avalanchego/utils/logging"
 	"github.com/ava-labs/avalanchego/utils/set"
-	"github.com/ava-labs/awm-relayer/database"
 	"github.com/ava-labs/awm-relayer/utils"
 
 	"github.com/ava-labs/subnet-evm/ethclient"
@@ -245,14 +244,6 @@ func (c *Config) GetSubnetID(blockchainID ids.ID) ids.ID {
 	return c.blockchainIDToSubnetID[blockchainID]
 }
 
-func (c *Config) GetAllRelayerKeys() []database.RelayerKey {
-	var keys []database.RelayerKey
-	for _, s := range c.SourceBlockchains {
-		keys = append(keys, s.GetRelayerKeys()...)
-	}
-	return keys
-}
-
 func (m *ManualWarpMessage) GetUnsignedMessageBytes() []byte {
 	return m.unsignedMessageBytes
 }
@@ -470,19 +461,6 @@ func (s *SourceBlockchain) GetSubnetID() ids.ID {
 
 func (s *SourceBlockchain) GetBlockchainID() ids.ID {
 	return s.blockchainID
-}
-
-func (s *SourceBlockchain) GetRelayerKeys() []database.RelayerKey {
-	var keys []database.RelayerKey
-	for _, dst := range s.GetSupportedDestinations().List() {
-		keys = append(keys, database.RelayerKey{
-			SourceBlockchainID:      s.GetBlockchainID(),
-			DestinationBlockchainID: dst,
-			OriginSenderAddress:     common.Address{}, // TODO: populate with allowed sender/receiver addresses
-			DestinationAddress:      common.Address{},
-		})
-	}
-	return keys
 }
 
 // Validatees the destination subnet configuration
--- a/config/config.go
+++ b/config/config.go
@@ -95,10 +95,13 @@ type WarpQuorum struct {
 }
 
 type Config struct {
-	LogLevel               string                   `mapstructure:"log-level" json:"log-level"`
-	PChainAPIURL           string                   `mapstructure:"p-chain-api-url" json:"p-chain-api-url"`
-	InfoAPIURL             string                   `mapstructure:"info-api-url" json:"info-api-url"`
-	StorageLocation        string                   `mapstructure:"storage-location" json:"storage-location"`
+	LogLevel        string `mapstructure:"log-level" json:"log-level"`
+	PChainAPIURL    string `mapstructure:"p-chain-api-url" json:"p-chain-api-url"`
+	InfoAPIURL      string `mapstructure:"info-api-url" json:"info-api-url"`
+	StorageLocation string `mapstructure:"storage-location" json:"storage-location"`
+	APIPort         uint64 `mapstructure:"api-port" json:"api-port"`
+	MetricsPort     uint64 `mapstructure:"metrics-port" json:"metrics-port"`
+
 	SourceBlockchains      []*SourceBlockchain      `mapstructure:"source-blockchains" json:"source-blockchains"`
 	DestinationBlockchains []*DestinationBlockchain `mapstructure:"destination-blockchains" json:"destination-blockchains"`
 	ProcessMissedBlocks    bool                     `mapstructure:"process-missed-blocks" json:"process-missed-blocks"`
@@ -112,6 +115,8 @@ func SetDefaultConfigValues(v *viper.Viper) {
 	v.SetDefault(LogLevelKey, logging.Info.String())
 	v.SetDefault(StorageLocationKey, "./.awm-relayer-storage")
 	v.SetDefault(ProcessMissedBlocksKey, true)
+	v.SetDefault(APIPortKey, 8080)
+	v.SetDefault(MetricsPortKey, 9090)
 }
 
 // BuildConfig constructs the relayer config using Viper.
@@ -141,6 +146,8 @@ func BuildConfig(v *viper.Viper) (Config, bool, error) {
 	cfg.InfoAPIURL = v.GetString(InfoAPIURLKey)
 	cfg.StorageLocation = v.GetString(StorageLocationKey)
 	cfg.ProcessMissedBlocks = v.GetBool(ProcessMissedBlocksKey)
+	cfg.APIPort = v.GetUint64(APIPortKey)
+	cfg.MetricsPort = v.GetUint64(MetricsPortKey)
 	if err := v.UnmarshalKey(ManualWarpMessagesKey, &cfg.ManualWarpMessages); err != nil {
 		return Config{}, false, fmt.Errorf("failed to unmarshal manual warp messages: %w", err)
 	}
--- a/config/config.go
+++ b/config/config.go
@@ -99,8 +99,8 @@ type Config struct {
 	PChainAPIURL    string `mapstructure:"p-chain-api-url" json:"p-chain-api-url"`
 	InfoAPIURL      string `mapstructure:"info-api-url" json:"info-api-url"`
 	StorageLocation string `mapstructure:"storage-location" json:"storage-location"`
-	APIPort         uint64 `mapstructure:"api-port" json:"api-port"`
-	MetricsPort     uint64 `mapstructure:"metrics-port" json:"metrics-port"`
+	APIPort         uint16 `mapstructure:"api-port" json:"api-port"`
+	MetricsPort     uint16 `mapstructure:"metrics-port" json:"metrics-port"`
 
 	SourceBlockchains      []*SourceBlockchain      `mapstructure:"source-blockchains" json:"source-blockchains"`
 	DestinationBlockchains []*DestinationBlockchain `mapstructure:"destination-blockchains" json:"destination-blockchains"`
@@ -146,8 +146,8 @@ func BuildConfig(v *viper.Viper) (Config, bool, error) {
 	cfg.InfoAPIURL = v.GetString(InfoAPIURLKey)
 	cfg.StorageLocation = v.GetString(StorageLocationKey)
 	cfg.ProcessMissedBlocks = v.GetBool(ProcessMissedBlocksKey)
-	cfg.APIPort = v.GetUint64(APIPortKey)
-	cfg.MetricsPort = v.GetUint64(MetricsPortKey)
+	cfg.APIPort = v.GetUint16(APIPortKey)
+	cfg.MetricsPort = v.GetUint16(MetricsPortKey)
 	if err := v.UnmarshalKey(ManualWarpMessagesKey, &cfg.ManualWarpMessages); err != nil {
 		return Config{}, false, fmt.Errorf("failed to unmarshal manual warp messages: %w", err)
 	}
--- a/config/config.go
+++ b/config/config.go
@@ -16,6 +16,7 @@ import (
 	"github.com/ava-labs/avalanchego/utils/constants"
 	"github.com/ava-labs/avalanchego/utils/logging"
 	"github.com/ava-labs/avalanchego/utils/set"
+	"github.com/ava-labs/awm-relayer/database"
 	"github.com/ava-labs/awm-relayer/utils"
 
 	"github.com/ava-labs/subnet-evm/ethclient"
@@ -244,25 +245,8 @@ func (c *Config) GetSubnetID(blockchainID ids.ID) ids.ID {
 	return c.blockchainIDToSubnetID[blockchainID]
 }
 
-// TODONOW: Put this somewhere else
-type RelayerKey struct {
-	SourceBlockchainID      ids.ID
-	DestinationBlockchainID ids.ID
-	OriginSenderAddress     common.Address
-	DestinationAddress      common.Address
-}
-
-func (k RelayerKey) CalculateRelayerKey() common.Hash {
-	return utils.CalculateRelayerKey(
-		k.SourceBlockchainID,
-		k.DestinationBlockchainID,
-		k.OriginSenderAddress,
-		k.DestinationAddress,
-	)
-}
-
-func (c *Config) GetAllRelayerKeys() []RelayerKey {
-	var keys []RelayerKey
+func (c *Config) GetAllRelayerKeys() []database.RelayerKey {
+	var keys []database.RelayerKey
 	for _, s := range c.SourceBlockchains {
 		keys = append(keys, s.GetRelayerKeys()...)
 	}
@@ -488,10 +472,10 @@ func (s *SourceBlockchain) GetBlockchainID() ids.ID {
 	return s.blockchainID
 }
 
-func (s *SourceBlockchain) GetRelayerKeys() []RelayerKey {
-	var keys []RelayerKey
+func (s *SourceBlockchain) GetRelayerKeys() []database.RelayerKey {
+	var keys []database.RelayerKey
 	for _, dst := range s.GetSupportedDestinations().List() {
-		keys = append(keys, RelayerKey{
+		keys = append(keys, database.RelayerKey{
 			SourceBlockchainID:      s.GetBlockchainID(),
 			DestinationBlockchainID: dst,
 			OriginSenderAddress:     common.Address{}, // TODO: populate with allowed sender/receiver addresses
--- a/config/config.go
+++ b/config/config.go
@@ -244,6 +244,31 @@ func (c *Config) GetSubnetID(blockchainID ids.ID) ids.ID {
 	return c.blockchainIDToSubnetID[blockchainID]
 }
 
+// TODONOW: Put this somewhere else
+type RelayerKey struct {
+	SourceBlockchainID      ids.ID
+	DestinationBlockchainID ids.ID
+	OriginSenderAddress     common.Address
+	DestinationAddress      common.Address
+}
+
+func (k RelayerKey) CalculateRelayerKey() common.Hash {
+	return utils.CalculateRelayerKey(
+		k.SourceBlockchainID,
+		k.DestinationBlockchainID,
+		k.OriginSenderAddress,
+		k.DestinationAddress,
+	)
+}
+
+func (c *Config) GetAllRelayerKeys() []RelayerKey {
+	var keys []RelayerKey
+	for _, s := range c.SourceBlockchains {
+		keys = append(keys, s.GetRelayerKeys()...)
+	}
+	return keys
+}
+
 func (m *ManualWarpMessage) GetUnsignedMessageBytes() []byte {
 	return m.unsignedMessageBytes
 }
@@ -461,6 +486,19 @@ func (s *SourceBlockchain) GetSubnetID() ids.ID {
 
 func (s *SourceBlockchain) GetBlockchainID() ids.ID {
 	return s.blockchainID
+}
+
+func (s *SourceBlockchain) GetRelayerKeys() []RelayerKey {
+	var keys []RelayerKey
+	for _, dst := range s.GetSupportedDestinations().List() {
+		keys = append(keys, RelayerKey{
+			SourceBlockchainID:      s.GetBlockchainID(),
+			DestinationBlockchainID: dst,
+			OriginSenderAddress:     common.Address{}, // TODO: populate with allowed sender/receiver addresses
+			DestinationAddress:      common.Address{},
+		})
+	}
+	return keys
 }
 
 // Validatees the destination subnet configuration
--- a/database/database.go
+++ b/database/database.go
@@ -6,6 +6,8 @@
 package database
 
 import (
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/awm-relayer/utils"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/pkg/errors"
 )
@@ -24,4 +26,20 @@ var (
 type RelayerDatabase interface {
 	Get(relayerKey common.Hash, key []byte) ([]byte, error)
 	Put(relayerKey common.Hash, key []byte, value []byte) error
+}
+
+type RelayerKey struct {
+	SourceBlockchainID      ids.ID
+	DestinationBlockchainID ids.ID
+	OriginSenderAddress     common.Address
+	DestinationAddress      common.Address
+}
+
+func (k RelayerKey) CalculateRelayerKey() common.Hash {
+	return utils.CalculateRelayerKey(
+		k.SourceBlockchainID,
+		k.DestinationBlockchainID,
+		k.OriginSenderAddress,
+		k.DestinationAddress,
+	)
 }
--- a/database/database.go
+++ b/database/database.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/awm-relayer/config"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/pkg/errors"
@@ -63,4 +64,24 @@ func CalculateRelayerKey(
 			"-",
 		)),
 	)
+}
+
+func GetConfigRelayerKeys(cfg *config.Config) []RelayerKey {
+	var keys []RelayerKey
+	for _, s := range cfg.SourceBlockchains {
+		keys = append(keys, GetSourceConfigRelayerKeys(s)...)
+	}
+	return keys
+}
+func GetSourceConfigRelayerKeys(cfg *config.SourceBlockchain) []RelayerKey {
+	var keys []RelayerKey
+	for _, dst := range cfg.GetSupportedDestinations().List() {
+		keys = append(keys, RelayerKey{
+			SourceBlockchainID:      cfg.GetBlockchainID(),
+			DestinationBlockchainID: dst,
+			OriginSenderAddress:     common.Address{}, // TODO: populate with allowed sender/receiver addresses
+			DestinationAddress:      common.Address{},
+		})
+	}
+	return keys
 }
--- a/database/database.go
+++ b/database/database.go
@@ -20,15 +20,20 @@ const (
 )
 
 var (
-	ErrKeyNotFound              = errors.New("key not found")
-	ErrChainNotFound            = errors.New("no database for chain")
+	ErrDataKeyNotFound          = errors.New("data key not found")
+	ErrRelayerKeyNotFound       = errors.New("no database for relayer key")
 	ErrDatabaseMisconfiguration = errors.New("database misconfiguration")
 )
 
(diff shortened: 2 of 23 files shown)


config/config.go
```diff
@@ -95,10 +95,13 @@ type WarpQuorum struct {
...
-	LogLevel               string                   `mapstructure:"log-level" json:"log-level"`
-	PChainAPIURL           string                   `mapstructure:"p-chain-api-url" json:"p-chain-api-url"`
-	InfoAPIURL             string                   `mapstructure:"info-api-url" json:"info-api-url"`
-	StorageLocation        string                   `mapstructure:"storage-location" json:"storage-location"`
+	LogLevel        string `mapstructure:"log-level" json:"log-level"`
+	PChainAPIURL    string `mapstructure:"p-chain-api-url" json:"p-chain-api-url"`
+	InfoAPIURL      string `mapstructure:"info-api-url" json:"info-api-url"`
+	StorageLocation string `mapstructure:"storage-location" json:"storage-location"`
+	APIPort         uint64 `mapstructure:"api-port" json:"api-port"`
+	MetricsPort     uint64 `mapstructure:"metrics-port" json:"metrics-port"`
```
cam-schultz, line 103: These were added to support running multiple relayers on the same host in the new E2E test. They are optional, and will default to the same values that are currently used.

messages/off-chain-registry/message_manager.go
```diff
@@ -171,3 +171,11 @@ func (m *messageManager) SendMessage(signedMessage *warp.Message, destinationBlo
 func (m *messageManager) GetDestinationBlockchainID(unsignedMessage *warp.UnsignedMessage) (ids.ID, error) {
 	return unsignedMessage.SourceChainID, nil
 }
+
+func (m *messageManager) GetOriginSenderAddress(unsignedMessage *warp.UnsignedMessage) (common.Address, error) {
```
michaelkaplan13, line 175: We probably will want to loop back on the off-chain message support at a higher level either way after the next milestone, but should we parse the `unsignedMessage` here for now and return an error if the source address isn't the expected value?

messages/teleporter/message_manager.go
```diff
@@ -105,6 +105,32 @@ func (m *messageManager) GetDestinationBlockchainID(unsignedMessage *warp.Unsign
...
+func (m *messageManager) GetOriginSenderAddress(unsignedMessage *warp.UnsignedMessage) (common.Address, error) {
+	teleporterMessage, err := m.parseTeleporterMessage(unsignedMessage)
+	if err != nil {
+		m.logger.Error(
+			"Failed get teleporter message.",
+			zap.String("warpMessageID", unsignedMessage.ID().String()),
+		)
+		return common.Address{}, err
+	}
+	m.teleporterMessageCache.Put(unsignedMessage.ID(), teleporterMessage)
```
michaelkaplan13, line 117: Can we move this repeated section to a helper somehow? Looks like we have it 4 or more times in this file now, so even though it doesn't save many lines it will be less error prone as far as the log message and adding the parsed message to the cache.

relayer/message_relayer.go
```diff
@@ -548,27 +591,176 @@ func (r *messageRelayer) aggregateSignatures(signatureMap map[int]blsSignatureBu
...
+// 1) The database contains the latest processed block data for the chain
+//   - In this case, we return the maximum of the latest processed block and the configured processHistoricalBlocksFromHeight
+//
+// 2) The database has been configured for the chain, but does not contain the latest processed block data
+//   - In this case, we return the configured processHistoricalBlocksFromHeight
+//
+// 3) The database does not contain the any information for the chain.
+//   - In this case, we return the configured processHistoricalBlocksFromHeight if it is set, otherwise
+//     we return the chain head.
+func (r *messageRelayer) calculateStartingBlockHeight(processHistoricalBlocksFromHeight uint64) (uint64, error) {
```
michaelkaplan13, line 608: Think ahead to having a individual `MessageRelayers` (or `ApplicationRelayers`) for each source contract->destination contraction, are you thinking we should allow for specifying `processHistoricalBlocksFromHeight` in the configuration for each pair? I haven't thought about the pros/cons of that vs always using the DB value, or if it doesn't existing using the current block height as an initial starting point.

michaelkaplan13 (commented):
I think these changes are great 🙌 

Haven't had a chance to go through it all closely yet, but left a few questions/comments.

geoff-vball (commented):
So overall, I don't think this change will allow us to unlock the level of concurrency we're looking for. The problem is that this implementation will still bottleneck if there are multiple transactions sent between the same two applications. 

In order to achieve proper concurrency, we're going to have to spawn a new ephemeral goroutine per transaction. I think it would be easier to have one persistent-struct per (source chain, destination chain) pair. That struct could filter-out messages based on a list of (source address, destination address) pairs.

I think the issue is going to come down to where we coalesce the information about which blocks have been processed. We have one `Relayer` per source subnet that processes blocks, so I don't think it makes sense for each tuple to keep track of it's own last seen block.

michaelkaplan13:
> So overall, I don't think this change will allow us to unlock the level of concurrency we're looking for. The problem is that this implementation will still bottleneck if there are multiple transactions sent between the same two applications.
>
> In order to achieve proper concurrency, we're going to have to spawn a new ephemeral goroutine per transaction. I think it would be easier to have one persistent-struct per (source chain, destination chain) pair. That struct could filter-out messages based on a list of (source address, destination address) pairs.
>
> I think the issue is going to come down to where we coalesce the information about which blocks have been processed. We have one Relayer per source subnet that processes blocks, so I don't think it makes sense for each tuple to keep track of it's own last seen block.

That's correct. There's a separate issue to processes messages concurrently [here](https://github.com/ava-labs/awm-relayer/issues/31), which will be the primary throughput increase for an individual "application relayer" instance.

The motivation for this specific change is that it's the first step to enable relayer deployment in a more enterprise-grade environment with a single shared external database instance for multiple relayer processes. If there are multiple relayer instances running between two subnets but supporting different application contracts, they should have unique entries in the database to track which checkpoint blocks they have relayed all of the messages up to since they are intended to relay a different set of messages.

database/database.go
```diff
@@ -22,6 +27,67 @@ var (
 
 // RelayerDatabase is a key-value store for relayer state, with each blockchainID maintaining its own state
 type RelayerDatabase interface {
-	Get(blockchainID ids.ID, key []byte) ([]byte, error)
-	Put(blockchainID ids.ID, key []byte, value []byte) error
+	Get(relayerKey common.Hash, key []byte) ([]byte, error)
```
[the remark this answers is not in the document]
cam-schultz: Agreed the name `key` on it's own is no longer descriptive enough. I'm hesistant on the name `keyType`, since it is still a key and not a "type" of key. I'd prefer something like `dataKey` to indicate that it's used to access data to be returned to the caller.

As for changing it to an enum, my only hesistation there is that it makes the interface less general. It would slightly increase type safety, but we do already have consts for the single key we use.  I don't feel strongly about this though, so happy to make this change if you think it's the right call.

> we'll probably want to prepend the key type prefix or something

Yup, that's pretty much what I was thinking. It shouldn't matter so much, as long as it's consistent in `Get` and `Put`.

messages/off-chain-registry/message_manager.go
cam-schultz: Good call. No need to assume the sender is what we expect here.

relayer/message_relayer.go
cam-schultz: I thought about this some, and where I landed was that we should keep `processHistoricalBlocksFromHeight` associated with the source chain, and not the application relayer. For one, the changes to implement it that way are very straightforward, and to do otherwise we'd either have to rearchitect the application to perform catch up at the application relayer level, or introduce a heuristic to determine the actual starting point for catch up. If the latter choice, we essentially arrive at what we have here, but with a config that doesn't as closely reflect the implemtation.

If there's a compelling use case for providing different heights to application relayers on startup, then we can reevaluate, but my opinion is that it's more trouble than it's worth atm.

messages/teleporter/message_manager.go
cam-schultz: Sure thing. The most straightforward change would be to update the cache in the call to `parseTeleporterMessage` if it's not already there.

relayer/message_relayer.go
```diff
@@ -51,48 +57,77 @@ var (
...
-		destinationBlockchainID: destinationBlockchainID,
+		logger:                  logger,
+		metrics:                 metrics,
+		network:                 network,
+		messageCreator:          messageCreator,
+		responseChan:            responseChan,
+		sourceBlockchain:        sourceBlockchain,
+		destinationBlockchainID: relayerKey.DestinationBlockchainID,
+		originSenderAddress:     relayerKey.OriginSenderAddress,
+		destinationAddress:      relayerKey.DestinationAddress,
```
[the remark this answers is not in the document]
cam-schultz: Removed the address fields as they're not used anywhere. I left in `destinationBlockchainID` as a separate field since it's used extensively throughout the type

messages/teleporter/message_manager.go
```diff
@@ -93,16 +93,36 @@ func (m *messageManager) GetDestinationBlockchainID(unsignedMessage *warp.Unsign
...
+			zap.String("warpMessageID", unsignedMessage.ID().String()),
+		)
+		return common.Address{}, err
+	}
+	return teleporterMessage.OriginSenderAddress, nil
+}
 
-	return teleporterMessage.DestinationBlockchainID, nil
+func (m *messageManager) GetDestinationAddress(unsignedMessage *warp.UnsignedMessage) (common.Address, error) {
+	teleporterMessage, err := m.parseTeleporterMessage(unsignedMessage)
```
bernard-avalabs, line 117: nit: GetDestinationBlockchainID, GetOriginSenderAddress and GetDestinationAddress differ by just the last line as far as I can tell. Should consider abstracting the shared part out to a helper function.

relayer/message_relayer.go
```diff
@@ -548,27 +586,176 @@ func (r *messageRelayer) aggregateSignatures(signatureMap map[int]blsSignatureBu
...
+//
+
+// Determines the height to process from. There are three cases:
+// 1) The database contains the latest processed block data for the chain
+//   - In this case, we return the maximum of the latest processed block and the configured processHistoricalBlocksFromHeight
+//
+// 2) The database has been configured for the chain, but does not contain the latest processed block data
+//   - In this case, we return the configured processHistoricalBlocksFromHeight
+//
+// 3) The database does not contain the any information for the chain.
```
bernard-avalabs, line 600:
```suggestion
// 3) The database does not contain any information for the chain.
```

relayer/message_relayer.go
```diff
@@ -548,27 +586,176 @@ func (r *messageRelayer) aggregateSignatures(signatureMap map[int]blsSignatureBu
 	return aggSig, vdrBitSet, nil
 }
 
+//
+// Database access
+//
+
+// Determines the height to process from. There are three cases:
+// 1) The database contains the latest processed block data for the chain
+//   - In this case, we return the maximum of the latest processed block and the configured processHistoricalBlocksFromHeight
```
bernard-avalabs, line 595: In what case would processHistoricalBlocksFromHeight > latest processed block?

relayer/message_relayer.go
```diff
@@ -548,27 +586,176 @@ func (r *messageRelayer) aggregateSignatures(signatureMap map[int]blsSignatureBu
...
+		return processHistoricalBlocksFromHeight, nil
+	} else if err != nil {
+		// Otherwise, we've encountered an unknown database error
+		r.logger.Error(
+			"failed to get latest block from database",
+			zap.String("relayerKey", r.relayerKey.CalculateRelayerKey().String()),
+			zap.Error(err),
+		)
+		return 0, err
+	}
```
bernard-avalabs, line 621: Just to confirm my understanding: We process/relay blocks in order. As a result, if we have processed the latest block height, then we know that there are no blocks prior to the latest block height that we have not processed. Is this correct?

bernard-avalabs (commented):
Generally makes sense to me. I've left a few comments, some of which are just questions to help me understand the design.

database/database.go
```diff
@@ -15,13 +20,79 @@ const (
...
+func GetConfigRelayerKeys(cfg *config.Config) []RelayerKey {
+	var keys []RelayerKey
+	for _, s := range cfg.SourceBlockchains {
+		keys = append(keys, GetSourceConfigRelayerKeys(s)...)
+	}
+	return keys
+}
+
+// Calculate all of the possible relayer keys for a given source blockchain
+func GetSourceConfigRelayerKeys(cfg *config.SourceBlockchain) []RelayerKey {
```
minghinmatthewlam, line 87: nit: can this be renamed to `GetSourceBlockchainRelayerKeys`? and possibly renaming argument parameter too

database/json_file_storage.go
```diff
@@ -80,41 +80,41 @@ func NewJSONFileStorage(logger logging.Logger, dir string, sourceBlockchains []*
...
-func (s *JSONFileStorage) getCurrentState(blockchainID ids.ID) (chainState, bool, error) {
+func (s *JSONFileStorage) getCurrentState(relayerKey common.Hash) (chainState, bool, error) {
 	currentState := make(chainState)
-	fileExists, err := s.read(blockchainID, &currentState)
+	fileExists, err := s.read(relayerKey, &currentState)
 	if err != nil {
 		s.logger.Error(
 			"failed to read file",
-			zap.String("blockchainID", blockchainID.String()),
+			zap.String("blockchainID", relayerKey.String()),
```
minghinmatthewlam, line 117: update log

relayer/message_relayer.go
cam-schultz: That would occur only if the user sets `processHistoricalBlocksFromHeight` to be greater than the stored height. If, for instance, the user intentionally takes their relayer offline for a period of time and doesn't want to process messages below a certain height.

The max function is much more relevant to the case in which a user sets `processHistoricalBlocksFromHeight` on initial startup, but we want to resume from the latest processed height rather than the configured height on subsequent startups. Taking the max of `processHistoricalBlocksFromHeight` and the stored height makes this possible without having to change the config.

relayer/message_relayer.go
cam-schultz: That's not necessarily the case when the relayer is processing historical blocks. We process historical messages and live incoming messages concurrently, so we can't make assumptions about ordering until we're done catching up. Live incoming messages are always processed in order relative to each other.

These ordering assumptions can be tricky to reason about. We have some ideas on how to relax those ordering requirements for both historical and live incoming messges. See the section `Relax Ordering Requirement` in https://github.com/ava-labs/awm-relayer/issues/234

main/main.go
```diff
@@ -149,10 +144,10 @@ func main() {
 
 	// start the health check server
 	go func() {
-		log.Fatalln(http.ListenAndServe(fmt.Sprintf(":%d", defaultApiPort), nil))
+		log.Fatalln(http.ListenAndServe(fmt.Sprintf(":%d", cfg.APIPort), nil))
 	}()
 
-	startMetricsServer(logger, gatherer, defaultMetricsPort)
+	startMetricsServer(logger, gatherer, uint32(cfg.MetricsPort))
```
minghinmatthewlam, line 150: can the type from config match one passed to `startMetricsServer`? Also, are ports valid up to `uint32`? From what I'm seeing online `uint16` would cover valid ports

messages/message_manager.go
```diff
@@ -31,6 +31,12 @@ type MessageManager interface {
 
 	// GetDestinationBlockchainID returns the destination chain ID of the destination chain for the given message
 	GetDestinationBlockchainID(unsignedMessage *warp.UnsignedMessage) (ids.ID, error)
+
+	// GetOriginSenderAddress returns the address of the sender on the origin chain
```
minghinmatthewlam, line 35:
```suggestion
	// GetOriginSenderAddress returns the address of the sender on the source blockchain
```

messages/message_manager.go
```diff
@@ -31,6 +31,12 @@ type MessageManager interface {
...
+	GetOriginSenderAddress(unsignedMessage *warp.UnsignedMessage) (common.Address, error)
+
+	// GetDestinationAddress returns the address of the recipient on the destination chain
```
minghinmatthewlam, line 38:
```suggestion
	// GetDestinationAddress returns the address of the recipient on the destination blockchain
```

relayer/relayer.go
```diff
@@ -114,37 +86,46 @@ func NewRelayer(
 	// scenario.
 	catchUpResultChan := make(chan bool, 1)
 
+	// Create the message relayers
+	messageRelayers := make(map[common.Hash]*messageRelayer)
+	for _, relayerKey := range database.GetSourceConfigRelayerKeys(&sourceBlockchain) {
+		messageRelayer, err := newMessageRelayer(logger, metrics, network, messageCreator, responseChan, relayerKey, db, sourceBlockchain, cfg)
```
minghinmatthewlam, line 92: nit should the parameters to the line be separated lines?

relayer/relayer.go
```diff
@@ -114,37 +86,46 @@ func NewRelayer(
...
+		if err != nil {
+			logger.Error(
+				"Failed to create message relayer",
+				zap.String("relayerKey", relayerKey.CalculateRelayerKey().String()),
+				zap.Error(err),
+			)
+			return nil, err
+		}
+		messageRelayers[relayerKey.CalculateRelayerKey()] = messageRelayer
```
minghinmatthewlam, line 101: nit: would consdier renaming function to `Hash` or similar. Seems strange to need to "calculate relayer key" when already have a `relayerKey`

database/database.go
minghinmatthewlam: does this function and the underlying functions it calls need to take in a pointer?

messages/teleporter/message_manager.go
cam-schultz: These methods already abstract most of their code to `parseTeleporter`. We could remove the logs emitted in GetDestinationBlockchainID, GetOriginSenderAddress and GetDestinationAddress so that they are just the call to `parseTeleporter` and the field access, but I also think there's value in keeping error logs verbose.

database/database.go
cam-schultz: > does this function and the underlying functions it calls need to take in a pointer?

`Config.GetSupportedDestinations` takes a pointer receiver, so the pointer arguments just prevent us from having to make that conversion in this call.

I'm not sure if this is what you were thinking about, but I'd definitely like to take steps to increase the immutability of the `Config` singleton. In the absence of const pointers, we'd have to refactor the `Config` type itself a bit to provide tighter access control.

main/main.go
cam-schultz: Yup, the max Unix port number is 2^16-1, so it could fit into a uint16.

relayer/relayer.go
cam-schultz: I agree the naming is strange. The `common.Hash` _is_ the key used to access the correct database partition, so I think the method should remain named `CalculateRelayerKey` This is also the key used to access the `messageRelayer` [mapping](https://github.com/ava-labs/awm-relayer/blob/114ec14d49e64ab2640db2ccc26703045d8a6dd3/relayer/relayer.go#L353) based on the parsed data from the Warp message.

How about renaming the type `RelayerKey` to `RelayerKeyParameters`? That way there's a clear distinction between the actual key and the data used to construct the key.

cam-schultz: Or maybe even just change the name of the method from `CalculateRelayerKey` to `GetKey` or even just `Key`.

relayer/message_relayer.go
michaelkaplan13: That makes sense to me. I definitely think we should have a single listener/filterer per source chain, which requires us to choose a single start block for all application relayers on that source chain. The implementation in this PR is reasonable in that it chooses the lowest checkpoint of all application relayers on that source chain, since there is no harm is re-scanning older messages for other application relayers. Given that though, I agree it makes sense to keep a single `processHistoricalBlocksFromHeight` for each source chain, rather than one per application relayer on that chain. 👍

database/database.go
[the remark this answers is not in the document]
michaelkaplan13: > I'd prefer something like dataKey to indicate that it's used to access data to be returned to the caller.

Devils advocate here, but isn't that true of both/all DB keys? 😅 

Changing it to an enum does make the interface less general, but I think that's actually a good thing here, since it will prevent it from being used incorrectly. Add new values to the enum is trivial for any code changes, but with a generic `[]byte` type, it could be easy to confuse what value should be passed.

relayer/message_relayer.go
```diff
@@ -336,8 +372,9 @@ func (r *messageRelayer) createSignedMessageAppRequest(requestID uint32) (*avala
 		if responsesExpected > 0 {
 			// Handle the responses. For each response, we need to call response.OnFinishedHandling() exactly once.
 			// Wrap the loop body in an anonymous function so that we do so on each loop iteration
-			for response := range r.relayer.responseChan {
-				r.relayer.logger.Debug(
+			// TODO: In order to run this concurrently, we need to route to each message relayer from the relayer responseChan
```
michaelkaplan13, line 375: Is this something we plan on making concurrent? It seems reasonable to me to handle responses serially as long as the requests are issued in a non-blocking manner. I'm not sure what you meant by "route to each message relayer from the relayer responseChan"

database/database.go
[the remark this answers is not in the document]
cam-schultz: On the naming front, my only strong feeling is that the name should end in "key" to denote that that parameter is exactly a key, and not some other descriptor. An alternative to `relayerKey/dataKey` would be `relayerID/key`. `relayerID` seems appropriate to me since that key is far less likely to change than the `key` in the future.

relayer/message_relayer.go
```diff
@@ -548,27 +586,176 @@ func (r *messageRelayer) aggregateSignatures(signatureMap map[int]blsSignatureBu
...
+// Store the block height in the database. Does not check against the current latest processed block height.
+func (r *messageRelayer) storeBlockHeight(height uint64) error {
+	return r.db.Put(r.relayerKey.GetKey(), database.LatestProcessedBlockKey, []byte(strconv.FormatUint(height, 10)))
+}
+
+// Stores the block height in the database if it is greater than the current latest processed block height.
+func (r *messageRelayer) storeLatestBlockHeight(height uint64) error {
+	// First, check that the stored height is less than the current block height
+	// This is necessary because the relayer may be processing blocks out of order on startup
+	latestProcessedBlock, err := r.getLatestProcessedBlockHeight()
```
michaelkaplan13, line 707: It will be nice to hopefully remove this extra read (even if just from memory) once we have #234 and #31

relayer/message_relayer.go
cam-schultz: In the global external handler, incoming responses are [routed](https://github.com/ava-labs/awm-relayer/blob/a5470472072226d3ed7d18c1e53f442052bc9502/peers/external_handler.go#L126) to the appropriate application relayer channel. At present, there's one channel per source blockchain, since there's at most one active application relayer per source blockchain. When we update the application relayers to process messages concurrently, we'll have to route the response to the correct signature aggregator.

The simplest approach would be to register a new channel per signature aggregator, then tear it down once the signature is constructed.

relayer/relayer.go
cam-schultz: > Or maybe even just change the name of the method from CalculateRelayerKey to GetKey or even just Key.

I've gone with this approach. Lmk if you agree.

database/database.go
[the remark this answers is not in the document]
michaelkaplan13: Gotcha, I was think of the data format closer to:
```
{
    "latestBlock": {
        "<relayer_key_1>": 42,
        "<relayer_key_2>": 43
    }
}
````

Rather than:
```
{
    "<relayer_key_1>": {
        "latestBlock: 42
    },
    "<relayer_key_2>": {
        "latestBlock: 43    
    }
}
```

But I see now that neither are correct. In the local JSON DB set up, we each `relayer_key` gets it's own file, and in the external flat key value store DB, the keys will be along the lines of "latestBlock-<relayer_key>".

**Review by geoff-vball — COMMENTED** [review]
Overall this looks really good! Just a couple small comments

File: relayer/message_relayer.go (hunk 18 above)
Comment (michaelkaplan13) [reply relayer/message_relayer.go:375]: Got it, thanks for the clarification

File: relayer/relayer.go [hunk 20]
```diff
@@ -404,35 +327,60 @@ func (r *Relayer) RelayMessage(warpLogInfo *vmtypes.WarpLogInfo, storeProcessedH
...
 		return nil
 	}
 
-	// Create and run the message relayer to attempt to deliver the message to the destination chain
-	messageRelayer, err := newMessageRelayer(r, unsignedMessage, destinationBlockchainID)
-	if err != nil {
+	messageRelayerID := database.CalculateRelayerID(
+		r.sourceBlockchain.GetBlockchainID(),
+		destinationBlockchainID,
+		common.Address{}, // TODO: Populate with the proper sender/receiver address
```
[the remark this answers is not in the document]
cam-schultz: This is populated in https://github.com/ava-labs/awm-relayer/pull/245

relayer/message_relayer.go
```diff
@@ -51,48 +56,73 @@ var (
...
 // Each messageRelayer runs in its own goroutine.
 type messageRelayer struct {
-	relayer                 *Relayer
-	warpMessage             *avalancheWarp.UnsignedMessage
+	logger                  logging.Logger
+	metrics                 *MessageRelayerMetrics
+	network                 *peers.AppRequestNetwork
+	messageCreator          message.Creator
+	responseChan            chan message.InboundMessage
+	sourceBlockchain        config.SourceBlockchain
```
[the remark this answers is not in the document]
cam-schultz: I originally decided to keep `destinationBlockchainID` in the struct to reduce the diff in this PR, but on second thought that's not worth having redundant data. Removed now.

geoff-vball (approved):
👍

PR #245, opened 2024-03-20
Source and Destination allowed addresses

### LINKED ISSUE #79: Support address allow/deny lists on source subnet
**Context and scope**
Relayers are configurable to relay Warp messages created via different messaging protocols, e.g. Teleporter. However, relayer operators may wish to only relay messages sent by a specific set of addresses, for instance if they want to tie relayer gas expenditure due to message delivery to transactions initiated by known addresses. Relayers may also be liable for compliance violations if they deliver messages sent by a sanctioned address.

We should add configuration options on the sending subnet to allow relayers to apply filters at the address level. 

**Open Questions**
- The contract that calls the warp precompile to send the message is the address listed in the Warp `AddressedCall` payload type. For example, this is the `TeleporterMessenger` contract address for Teleporter messages. How would we get the original transaction sender in the relayer? Is that included in the message logs/receipt?
- There are two addresses to consider: the EOA that initiates the transaction on the source blockchain, and the address that calls Teleporter (or another messaging protocol). It makes sense to filter on both, but they may need to be treated differently.

### LINKED ISSUE #80: Support address allow list on destination subnets
**Context and scope**
Relayers are configurable to relay Warp messages created via different messaging protocols, e.g. Teleporter. However, relayer operators may wish to only relay messages to a particular set of Teleporter `destinationAddress`s. This would be useful if a relayer operator only wants to interact with certain contracts on the destination side.

We should add configuration options on the sending subnet to allow relayers to apply filters at the address level. This would likely look like an additional condition in `MessageManager.ShouldSendMessage`.

cam-schultz:
## Why this should be merged
Fixes #79 
Fixes #80

## How this works
Adds `allowed-origin-sender-addresses` and `allowed-destination-addresses` configuration options. Each combination of (source blockchainID, destination blockchainID, allowedSourceAddress, allowedDestinationAddress) results in a unique database key. If either of these lists are empty, then all addresses are allowed, which is represented by the zero address in the database key.

## How this was tested
CI, added unit tests and e2e test
## How is this documented
Update README


relayer/relayer.go
```diff
@@ -387,3 +460,29 @@ func (r *Relayer) CheckSupportedDestination(destinationBlockchainID ids.ID) bool
 	supportedDsts := r.sourceBlockchain.GetSupportedDestinations()
 	return supportedDsts.Contains(destinationBlockchainID)
 }
+
+func (r *Relayer) CheckAllowedOriginSenderAddress(addr common.Address) bool {
+	allowedAddresses := r.globalConfig.GetSourceBlockchainAllowedAddresses(r.sourceBlockchain.GetBlockchainID())
+	if bytes.Equal(allowedAddresses[0].Bytes(), common.Address{}.Bytes()) {
```
geoff-vball, line 466: What's the rationale behind putting the empty address in the slice, instead of just checking for an empty slice?

geoff-vball (commented):
Just one question, otherwise LGTM

relayer/relayer.go
cam-schultz: Really just to encapsulate the "all allowed" relayer key case in the `Config`, especially w.r.t. [fetching the relayer keys](https://github.com/ava-labs/awm-relayer/blob/5982596abfbd082441be59418089d9ae676d154a/database/database.go#L87).

The approach I went with requires the caller to understand that the zero address represents a wildcard, whereas keeping the slice empty would require the caller to understand that an empty slice represents a wildcard.

geoff-vball (commented):
Could you fix the merge conflicts with `shared-db-keys`, and then I'll take another pass at this

cam-schultz:
> Could you fix the merge conflicts with `shared-db-keys`, and then I'll take another pass at this

Resolved

relayer/relayer.go
geoff-vball: I'm not a huge fan of the sentinel value, but not too picky. I think having the behaviour line up with what we expect in the config (empty list in the config means all allowed). As-is we could also panic if `GetSourceBlockchainAllowedAddresses` returns an empty slice. At the very least we should clearly document in/above each of the functions the expected behaviour.

geoff-vball: If we're worried about accidentally not having initialized the slice, we could check for a empty non-nil slice?

README.md
```diff
@@ -211,15 +215,19 @@ The relayer is configured via a JSON file, the path to which is passed in via th
...
+  - The RPC endpoint of the destination blockchains's API node.
 
   `"account-private-key": string`
 
-  - The hex-encoded private key to use for signing transactions on the destination Subnet. May be provided by the environment variable `ACCOUNT_PRIVATE_KEY`. Each `destination-subnet` may use a separate private key by appending the blockchain ID to the private key environment variable name, for example `ACCOUNT_PRIVATE_KEY_11111111111111111111111111111111LpoYY`
+  - The hex-encoded private key to use for signing transactions on the destination blockchain. May be provided by the environment variable `ACCOUNT_PRIVATE_KEY`. Each `destination-subnet` may use a separate private key by appending the cb58 encoded blockchain ID to the private key environment variable name, for example `ACCOUNT_PRIVATE_KEY_11111111111111111111111111111111LpoYY`
+
+  `"allowed-destination-addresses": []string`
+
+  - List of addresses that the relayer is allowed to deliver Warp messages to on this destination blockchain. The destination address is defined by the message protocol. For example, it could be the address called from the message protocol contract. If empty, then all addresses are allowed.
```
michaelkaplan13, line 230:
```suggestion
  - List of addresses that the relayer supports delivering Warp messages to on this destination blockchain. The destination address is defined by the message protocol. For example, it could be the address called from the message protocol contract. If empty, then all addresses are allowed.
```

config/config.go
```diff
@@ -67,26 +67,30 @@ type SourceBlockchain struct {
...
-	BlockchainID      string `mapstructure:"blockchain-id" json:"blockchain-id"`
-	VM                string `mapstructure:"vm" json:"vm"`
-	RPCEndpoint       string `mapstructure:"rpc-endpoint" json:"rpc-endpoint"`
-	AccountPrivateKey string `mapstructure:"account-private-key" json:"account-private-key"`
+	SubnetID                    string   `mapstructure:"subnet-id" json:"subnet-id"`
+	BlockchainID                string   `mapstructure:"blockchain-id" json:"blockchain-id"`
+	VM                          string   `mapstructure:"vm" json:"vm"`
+	RPCEndpoint                 string   `mapstructure:"rpc-endpoint" json:"rpc-endpoint"`
+	AccountPrivateKey           string   `mapstructure:"account-private-key" json:"account-private-key"`
+	AllowedDestinationAddresses []string `mapstructure:"allowed-destination-addresses" json:"allowed-destination-addresses"`
```
michaelkaplan13, line 85: Would it make more sense to define the allowed destinations as part of the `supportedDestinations` of the `SourceBlockchain` configuration? I think it's a bit confusing to have supported destinations defined separately from allowed destination addresses, and I think is not as specific as could be required.

For instance, if a relayer supports relaying messages bidirectionally between `A<->B`, and `A<->C`, then the allowed destination addresses defined for on Subnet A will apply for messages from both Subnets B and C, which doesn't map will into the intended application relayer instances that should be created.

I'd recommend defining the `SupportedDestinations` of a `SourceSubnet` fully along the lines of:
```
type SupportedDestination {
    BlockchainID ids.ID
    DestinationAddresses []common.Address
}
```

And then not having any additional configuration of the `DestinationBlockchain` struct.

cam-schultz: I went ahead and made this change; I think it makes a lot of sense. It also brings us one step closer to having a config where the application relayer configuration is standalone at the top level. I'm thinking something like:

`type Config struct {
    ApplicationRelayers []ApplicationRelayer
    SourceBlockchains []SourceBlockchain
    DestinationBlockchains []DestinationBlockchain
}
`

where `SourceBlockchain` and `DestinationBlockchain` specify how to connect to the chains, and `ApplicationRelayer` specifies a distinct relayer path.

relayer/relayer.go
```diff
@@ -394,6 +467,43 @@ func (r *Relayer) RelayMessage(warpLogInfo *vmtypes.WarpLogInfo, storeProcessedH
...
+	}
+	return false
+}
+
+func (r *Relayer) CheckAllowedOriginSenderAddress(addr common.Address) bool {
+	allowedAddresses := r.globalConfig.GetSourceBlockchainAllowedAddresses(r.sourceBlockchain.GetBlockchainID())
+	if bytes.Equal(allowedAddresses[0].Bytes(), common.Address{}.Bytes()) {
+		return true
+	}
+	for _, allowedAddress := range allowedAddresses {
```
michaelkaplan13, line 483: Seems like it may make sense to make `allowedAddresses` a `Set` rather than having to iterate it each time we see a message here.

config/config.go
```diff
@@ -565,3 +602,12 @@ func (c *Config) GetWarpQuorum(blockchainID ids.ID) (WarpQuorum, error) {
 	}
 	return WarpQuorum{}, errFailedToGetWarpQuorum
 }
+
+func (c *Config) GetSourceBlockchainAllowedAddresses(blockchainID ids.ID) []common.Address {
```
michaelkaplan13, line 606: It probably doesn't make a huge performance difference as long as there aren't a large number of supported source chains by an individual relayer, but should we consider creating a map here so we don't need to iterate `SourceBlockchains` each time? It looks like this is called each time the relayer sees a message, so I think it's probably worth a TODO if it's larger of a change than you'd like to make now.

relayer/relayer.go
```diff
@@ -284,6 +285,62 @@ func (r *Relayer) ReconnectToSubscriber() error {
...
+	)
+	if messageRelayer, ok := r.messageRelayers[messageRelayerID]; ok {
+		return messageRelayer, ok
+	}
+
+	// Check for a match on sourceBlockchainID and destinationBlockchainID, with any originSenderAddress and any destinationAddress
+	messageRelayerID = database.CalculateRelayerID(
+		sourceBlockchainID,
+		destinationBlockchainID,
+		common.Address{},
```
michaelkaplan13, line 315: Could we define a `allAllowedDestinationAddress` const and use that here instead?

relayer/relayer.go
```diff
@@ -343,26 +409,33 @@ func (r *Relayer) RelayMessage(warpLogInfo *vmtypes.WarpLogInfo, storeProcessedH
...
 			"Message destination chain ID not supported. Not relaying.",
 			zap.String("blockchainID", r.sourceBlockchain.GetBlockchainID().String()),
-			zap.String("sourceBlockchain.GetBlockchainID()", destinationBlockchainID.String()),
+			zap.String("destinationBlockchainID", destinationBlockchainID.String()),
 		)
 		return nil
 	}
 
-	messageRelayerID := database.CalculateRelayerID(
+	messageRelayer, ok := r.getMessageRelayer(
```
michaelkaplan13, line 432: Is this check sufficient for removing the `CheckSupportedDestination`, `CheckAllowedDestinationAddress` and `CheckAllowedOriginSenderAddress` checks above? I think if it's not a supported source/destination, then we won't have a message relayer for it.

relayer/relayer.go
cam-schultz: I went ahead and implemented your suggestion, and agree it's cleaner. It makes sense to have the relayer ID constructor interpret the empty list, rather than modifying the config object directly.

relayer/relayer.go
cam-schultz: It totally is, great call. Removed all of those checks. That should also resolve a handful of your other comments :)

relayer/relayer.go
cam-schultz: Added

config/config.go
cam-schultz: This method has been removed.

relayer/relayer.go
cam-schultz: This method has been removed.

README.md
```diff
@@ -147,11 +147,11 @@ The relayer is configured via a JSON file, the path to which is passed in via th
 
   `"source-blockchain-id": string`
 
-  - cb58-encoded Blockchain ID of the source Subnet.
+  - cb58-encoded Blockchain ID of the source blockchain.
 
   `"destination-blockchain-id": string`
 
-  - cb58-encoded Blockchain ID of the destination Subnet.
+  - cb58-encoded Blockchain ID of the destination blockchain.
```
minghinmatthewlam, line 154: nit: don't think blockchain needs to be capitalized, and other places they are not.

config/config.go
```diff
@@ -58,20 +58,38 @@ type ManualWarpMessage struct {
 	destinationAddress      common.Address
 }
 
+type SupportedDestination struct {
```
minghinmatthewlam, line 61: I think it'd be a good practice for us to start adding documentation above structs

config/config.go
```diff
@@ -432,32 +446,45 @@ func (s *SourceBlockchain) Validate(destinationBlockchainIDs *set.Set[string]) e
...
-
 	// If the list of supported destinations is empty, populate with all of the configured destinations
 	if len(s.SupportedDestinations) == 0 {
 		for _, blockchainIDStr := range destinationBlockchainIDs.List() {
-			blockchainID, err := ids.FromString(blockchainIDStr)
-			if err != nil {
-				return fmt.Errorf("invalid blockchainID in configuration. error: %w", err)
-			}
-			s.supportedDestinations.Add(blockchainID)
+			s.SupportedDestinations = append(s.SupportedDestinations, &SupportedDestination{
```
minghinmatthewlam, line 452: I think we should avoid cases of not initializing the full struct. How about a `NewSupportedDestination` func that takes in both fields and does the conversion for blockchainID, and sets the access variables.

database/database.go
```diff
@@ -100,13 +103,27 @@ func GetConfigRelayerIDs(cfg *config.Config) []RelayerID {
...
+	}
+	for _, srcAddress := range srcAddresses {
+		for _, dst := range sourceBlockchain.SupportedDestinations {
+			dstAddresses := dst.GetAddresses()
+			// If no addresses are provided, use the zero address to construct the relayer ID
+			if len(dstAddresses) == 0 {
+				dstAddresses = append(dstAddresses, AllAllowedAddress)
+			}
+			for _, dstAddress := range dstAddresses {
+				ids = append(ids, RelayerID{
```
minghinmatthewlam, line 119: I think it'd be better to have `NewRelayerID` that initializes here too

relayer/relayer.go
```diff
@@ -284,6 +284,62 @@ func (r *Relayer) ReconnectToSubscriber() error {
 	return nil
 }
 
+// Fetch the appropriate message relayer
+// Checks for the following registered keys. Exactly one of these keys should be registered, or none of them.
+// 1. An exact match on sourceBlockchainID, destinationBlockchainID, originSenderAddress, and destinationAddress
+// 2. A match on sourceBlockchainID and destinationBlockchainID, with a specific originSenderAddress and any destinationAddress
+// 3. A match on sourceBlockchainID and destinationBlockchainID, with any originSenderAddress and a specific destinationAddress
+// 4. A match on sourceBlockchainID and destinationBlockchainID, with any originSenderAddress and any destinationAddress
+func (r *Relayer) getMessageRelayer(
```
minghinmatthewlam, line 293: instead of checking for all 4 cases, can something with prior knoweldge of the source blockchain's supported destinations, save information about which of the 4 cases each supported destination is?

config/config.go
cam-schultz: Up until the loop starting below this one, none of the `SupportedDestinations` are fully populated, so this check is equivalent to the user manually providing all of the destination chain IDs with no allowed addresses specified. Then in the following loop we populate all of the elements of the list.

Adding a `NewSupportedDestination` method call here doesn't make much sense IMO, since the `SupportedDestination` objects already exist.

database/database.go
```diff
@@ -19,6 +19,9 @@ var (
 	ErrKeyNotFound              = errors.New("key not found")
 	ErrRelayerIDNotFound        = errors.New("no database entryfor relayer id")
 	ErrDatabaseMisconfiguration = errors.New("database misconfiguration")
+
+	// AllAllowedAddress is used to construct relayer IDs when all addresses are allowed
+	AllAllowedAddress = common.Address{}
```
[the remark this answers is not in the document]
cam-schultz: Both of the suggested names would fit better with a `bool` variable. `AllAllowedAddress`, while clunky, gets across that the variable is an `Address`.

database/database.go
cam-schultz: I don't think there's any real benefit to adding a standalone constructor on a POD datatype. It would result in a net increase of code, since it would take the exact same arguments as the composite literal here.

relayer/relayer.go
cam-schultz: We could probably optimize this function by implementing a more intelligent matcher, but that seems like premature optimization. We could also add a cache of registered relayer IDs that we could check instead of re-computing them each time, but that also seems premature. This function isn't exactly expensive when considering that message relaying is heavily network-bound.

config/config.go
```diff
@@ -432,32 +457,45 @@ func (s *SourceBlockchain) Validate(destinationBlockchainIDs *set.Set[string]) e
...
 				blockchainID)
 		}
-		s.supportedDestinations.Add(blockchainID)
+		dest.blockchainID = blockchainID
+		for _, address := range dest.Addresses {
+			destinationAddress, err := hex.DecodeString(utils.SanitizeHexString(address))
+			if err != nil {
+				return err
+			}
+			dest.addresses = append(dest.addresses, common.BytesToAddress(destinationAddress))
```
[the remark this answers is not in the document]
michaelkaplan13: Also below for the allowed sender addresses

relayer/relayer.go
```diff
@@ -284,6 +284,62 @@ func (r *Relayer) ReconnectToSubscriber() error {
 	return nil
 }
 
+// Fetch the appropriate message relayer
+// Checks for the following registered keys. Exactly one of these keys should be registered, or none of them.
```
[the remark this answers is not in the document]
michaelkaplan13: Is the invariant that at most one of these cases can match enforced at the code level, and if not, should it be?

IIUC, I think the configuration could set an allowed addresses value of `[0x123, 0x000]`, and multiple message relayers would be created that have overlapping sets of messages to relay. 

We may want to consider explicitly disallowing the `0x000` address from being set in the configuration such that the only way to have that behavior is setting an empty list, which excludes the possibility of multiple cases matching here.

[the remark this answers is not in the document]
cam-schultz: The invariant is enforced at the code level by `GetSourceBlockchainRelayerIDs`. From that function, it's clear there's only 4 possible "types" of key:
- any source + any destination
- any source + specific destination
- specific source + any destination
- specific source + specific destination

However, you're correct that the zero address would count as a "specific" address if provided in the configuration. We should disallow that in the config validation.

database/database.go
minghinmatthewlam: I was originally thinking to initialize the `RelayerID.id` field in the construtor instead of in `GetID`, but looking back not important to add here.

config/config.go
minghinmatthewlam: makes sense since the object already exists. My main gripe with this file is that the types `SourceBlockchinID` `DestinationBlockchainID` `ManualWarpMessage` `Config`, etc. and all their helper and validation functions all stuck into one file. Since the code has expanded more than we originally had, would want to distribute to different files. Feel free to defer to new issue though.

relayer/relayer.go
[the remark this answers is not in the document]
cam-schultz: Pushed the change to disallow the zero address in the config.

config/config.go
```diff
@@ -432,33 +455,53 @@ func (s *SourceBlockchain) Validate(destinationBlockchainIDs *set.Set[string]) e
...
-			s.supportedDestinations.Add(blockchainID)
+			s.SupportedDestinations = append(s.SupportedDestinations, &SupportedDestination{
+				BlockchainID: blockchainIDStr,
+			})
 		}
 	}
-
-	for _, blockchainIDStr := range s.SupportedDestinations {
-		blockchainID, err := ids.FromString(blockchainIDStr)
+	zeroAddress := common.Address{}
```
cam-schultz, line 466: We unfortunately can't use `database.AllAllowedAddresses` here without introducing an import cycle. I don't want to move that variable to an upstream package, since it logically belongs to the `database` package. So I elected to explicitly disallow the zero address in the config, and not reference that variable here at all.

config/config.go
cam-schultz: Agreed that this file should be split up. I'll defer that to a later issue though: https://github.com/ava-labs/awm-relayer/issues/251

database/database.go
cam-schultz: Good point re: initializing the `id` field. I added a constructor, and made `ID` exported so that it can be referenced directly.

config/config.go
```diff
@@ -432,33 +455,53 @@ func (s *SourceBlockchain) Validate(destinationBlockchainIDs *set.Set[string]) e
...
 	}
 
+	// Validate and store the allowed origin source addresses
+	allowedOriginSenderAddresses := make([]common.Address, len(s.AllowedOriginSenderAddresses))
+	for i, addressStr := range s.AllowedOriginSenderAddresses {
+		if !common.IsHexAddress(addressStr) {
+			return fmt.Errorf("invalid allowed origin sender address in source blockchain configuration: %s", addressStr)
+		}
+		address := common.HexToAddress(addressStr)
+		zeroAddress := common.Address{}
```
michaelkaplan13, line 497: Don't think we need to re-define zero address here

michaelkaplan13 (approved):

minghinmatthewlam (approved):
LGTM

PR #249, opened 2024-03-27
KMS integration

### LINKED ISSUE #194: Support dynamic key management
**Context and scope**
Rather than reading the wallet private key from the configuration or environment, we should integrate a key management system such as AWS KMS.

cam-schultz:
## Why this should be merged
Fixes #194 

## How this works
Adds KMS transaction signing support. Introduces a new `Signer` interface with implementations for KMS and local signing. Adds configuration checks to ensure that only of the KMS configuration and an account private key are provided.

## How this was tested
CI, new unit tests
Tested manually against a hosted KMS key

## How is this documented
Updated README


config/config.go
```diff
@@ -151,6 +151,24 @@ func BuildConfig(v *viper.Viper) (Config, bool, error) {
 		return Config{}, false, fmt.Errorf("failed to unmarshal source subnets: %w", err)
 	}
 
+	// Explicitly overwrite the configured KMS configuration
```
michaelkaplan13, line 154: Should we support setting different KMS configurations for each supported destination chain via env vars, similarly to how we do the raw private keys?

config/config.go
```diff
@@ -474,13 +492,21 @@ func (s *DestinationBlockchain) Validate() error {
...
 
-	if len(s.AccountPrivateKey) != relayerPrivateKeyBytes*2 {
-		return errors.New("invalid account private key hex string")
-	}
-
-	if _, err := hex.DecodeString(s.AccountPrivateKey); err != nil {
-		return fmt.Errorf("invalid account private key hex string: %w", err)
+		if _, err := hex.DecodeString(s.AccountPrivateKey); err != nil {
+			return fmt.Errorf("invalid account private key hex string: %w", err)
+		}
```
michaelkaplan13, line 509: I think the behavior we're going for here is:
- if `AccountPrivateKey` is set, it must be valid, `KMSKeyID` cannot be set, and `AWSRegion` cannot be set.
- otherwise, both `KMSKeyID` and `AWSRegion` must be set and valid.

For validating the account private key, I recommend we use `crypto.HexToECDSA` to provide more robust validation.

config/config.go
cam-schultz: I think we should hold off on more advanced/fine grained KMS integration until we have explicit requests for such features. I imagine we'll want a mechanism to select different keys for different purposes or at different times, but I'm not quite sure what that would be. For instance, I know round-robin cycling through a set of keys is a common technique.

cam-schultz: Discussed some more offline, and concluded that the global KMS option is unnecessarily complicated. The KMS settings are not sensitive, so they can safely be provided via config file.

sample-relayer-config.json
```diff
@@ -1,6 +1,7 @@
 {
   "info-api-url": "https://api.avax-test.network",
   "p-chain-api-url": "https://api.avax-test.network",
+  "log-level": "debug",
```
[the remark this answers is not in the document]
cam-schultz: This and all the other changes to this file have been reverted.

vms/evm/signer/tx_signer_test.go
```diff
@@ -0,0 +1,80 @@
...
+		{
+			name: "invalid 0x prefix",
+			dst: config.DestinationBlockchain{
+				AccountPrivateKey: "0x56289e99c94b6912bfc12adc093c9b51124f0dc54ac7a766b2bc5ccf558d8027",
+			},
+			expectedResult: retStruct{
+				pk: &ecdsa.PrivateKey{
+					D: big.NewInt(-5567472993773453273),
+				},
+				addr: common.HexToAddress("0x8db97C7cEcE249c2b98bDC0226Cc4C2A57BF52FC"),
```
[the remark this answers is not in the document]
cam-schultz: We do sanitize the private key string when parsing the config, but there's certainly no harm in defensively doing it again here. Made this change and removed this test case.

vms/evm/signer/kms_signer.go
```diff
@@ -0,0 +1,149 @@
...
+	sigBytes, err := getEthereumSignature(asn1pubk.PublicKey.Bytes, h, sigAsn1.R.Bytes, sigAsn1.S.Bytes)
+	if err != nil {
+		return nil, err
+	}
+	return tx.WithSignature(signer, sigBytes)
+}
+
+func (s *KMSSigner) Address() common.Address {
+	return common.Address{}
+}
```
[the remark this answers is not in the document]
cam-schultz: D'oh this was an oversight on my part. I had the zero address as a placeholder and neglected to return to it. This is fixed now: I updated the constructor to cache the pubKey and EOA, and use those in both `Address` and `SignTx`

geoff-vball (approved):

vms/evm/signer/kms_signer.go
```diff
@@ -0,0 +1,149 @@
...
+	signature := append(rsSignature, []byte{0}...)
+
+	recoveredPublicKeyBytes, err := crypto.Ecrecover(txHash, signature)
+	if err != nil {
+		return nil, err
+	}
+	s0 := hex.EncodeToString(recoveredPublicKeyBytes)
+
+	if hex.EncodeToString(recoveredPublicKeyBytes) != hex.EncodeToString(expectedPublicKeyBytes) {
+		signature = append(rsSignature, []byte{1}...)
```
bernard-avalabs, line 131: Seems like there would be less duplicated code if we create a getEthereumSignature helper that takes V as a parameter, and have getEthereumSignature call the helper with V=0 and then V=1 if necessary.

bernard-avalabs (approved):
Generally looks good to me. Left a few minor comments.

config/config.go
```diff
@@ -526,13 +526,17 @@ func (s *DestinationBlockchain) Validate() error {
...
+	if s.KMSKeyID != "" {
+		if s.KMSAWSRegion == "" {
+			return errors.New("KMS key ID provided without an AWS region")
+		}
+		if s.AccountPrivateKey != "" {
+			return errors.New("only one of account private key or KMS key ID can be provided")
+		}
+	} else {
+		if _, err := crypto.HexToECDSA(s.AccountPrivateKey); err != nil {
+			return fmt.Errorf("invalid account private key hex string: %w", err)
```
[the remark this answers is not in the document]
cam-schultz: That's a great call. The other place we handle the private key is when [signing the transaction ](https://github.com/ava-labs/awm-relayer/blob/fcba4316b9df50ab43eb1305bec5dc8f8dac5ac3/vms/evm/signer/tx_signer.go#L45). In both cases, the function we call does not include the private key in the error message, but there's no guarantee that that will be true in future releases. As a defensive measure, I've updated the error messages returned by the caller to _not_ wrap these errors.

vms/evm/signer/kms_signer.go
cam-schultz: Refactored these helpers to be a bit cleaner. Thanks for the suggestion

michaelkaplan13 (commented):
A few small comments, but otherwise LGTM 👍

vms/evm/signer/kms_signer.go
```diff
@@ -0,0 +1,155 @@
...
+// Trim null bytes and pad with zeros to 32 bytes
+//
+// This function is adapted from https://github.com/welthee/go-ethereum-aws-kms-tx-signer
+// and is reproduced here under welthee's MIT license
+func adjustSignatureLength(buffer []byte) []byte {
+	buffer = bytes.TrimLeft(buffer, "\x00")
+	for len(buffer) < 32 {
+		zeroBuf := []byte{0}
+		buffer = append(zeroBuf, buffer...)
+	}
```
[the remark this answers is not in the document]
cam-schultz: Updated this function to be a bit more efficient. Added a check to potentially skip the initial trim, and reduced to a single allocation after that.

vms/evm/destination_client.go
```diff
@@ -145,9 +126,18 @@ func (c *destinationClient) SendTx(signedMessage *avalancheWarp.Message,
 	gasFeeCap := baseFee.Mul(baseFee, big.NewInt(BaseFeeFactor))
 	gasFeeCap.Add(gasFeeCap, big.NewInt(MaxPriorityFeePerGas))
 
+	evmChainID, err := c.client.ChainID(context.Background())
```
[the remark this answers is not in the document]
cam-schultz: Yes, good call

vms/evm/signer/kms_signer.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// Trim null bytes and pad with zeros to 32 bytes
+func adjustSignatureLength(buffer []byte) []byte {
+	if len(buffer) > 32 {
+		buffer = bytes.TrimLeft(buffer, "\x00")
+	}
+
+	// pad to 32 bytes
+	pad := 32 - len(buffer)
```
michaelkaplan13, line 171: May want a defensive check that the length here is <= 32, which is may not be the case with certain inputs, but should never occur.

cam-schultz: I elected to make that check at the [caller](https://github.com/ava-labs/awm-relayer/blob/1aca8b40c2181862a7fd70c0fd1d5d6353263d18/vms/evm/signer/kms_signer.go#L129) instead, that way the call to `adjustSignatureLength` could still be made inline.

michaelkaplan13 (approved):

geoff-vball (approved):

Issue #225, opened 2024-03-08
Integrate an external database


cam-schultz:
**Context and scope**
The relayer currently stores the checkpointed block height for each configured source chain in a thin inline "database" that is just a JSON file saved to disk on the same host as the application. At the time of development, this approach had the advantage of being human-readable and modifiable outside of the application, allowing a relayer operator to tailor their instance to relay messages from a specific height. The JSON is written to on the order of once per second, and only read once at startup, so there were no real bandwidth considerations either. At the time, an external DBMS was deemed overkill.

However, disks are generally not considered durable on their own in cloud environments, which is why most cloud providers recommend decoupling compute from storage by integrating an external DBMS. In our use case, a failed disk would be completely safe, as the relayer is able to specify a historical block from which to backprocess Warp messages, but it would still require manual intervention to address. An external DBMS would provide much stronger data persistence guarantees, as well as provide a more scalable solution if we ever want to expand the relayer's storage footprint in the future.

**Discussion and alternatives**
Given the low storage and throughput requirements, we should opt for the solution that 1) provides the simplest integration and 2) offers the lowest operating costs.

PR #252, opened 2024-04-02
Redis support

cam-schultz:
## Why this should be merged
Fixes https://github.com/ava-labs/awm-relayer/issues/225 by adding a Redis integration.

## How this works
- Adds a type `RedisDatabase` that implements `RelayerDatabase`.
- Adds "redis-url" ocnfiguration option that takes precedence over "storage-location" if set.

## How this was tested
CI. Manually verified block heights update against a local Redis instance.

 Open for discussion, but modifying the E2E setup to spin up a Redis server would likely be much more complex than the feature changes. My preference would be to test outside of CI, and then circle back in a serparate ticket if it makes sense.

## How is this documented
Updated README


geoff-vball (dismissed):
This all LGTM. Adding redis to the testing setup shouldn't be too bad though. Super okay leaving it for another ticket. 

I think we have 2 different configurations we should support? These could each be tackled in their own ticket.

1. Start a redis instance from with `tests/utils/utils.go`. We probably just need to check that it's installed when calling `e2e_test.sh`.
2. Deploying a redis instance from a docker container. This could also maybe be used in step 1 above.

bernard-avalabs (approved):
LGTM

cam-schultz:
> This all LGTM. Adding redis to the testing setup shouldn't be too bad though. Super okay leaving it for another ticket.
> 
> I think we have 2 different configurations we should support? These could each be tackled in their own ticket.
> 
> 1. Start a redis instance from with `tests/utils/utils.go`. We probably just need to check that it's installed when calling `e2e_test.sh`.
> 2. Deploying a redis instance from a docker container. This could also maybe be used in step 1 above.

Are you referring to the `teleporter` docker setup in 2? I think that's a bit overkill. We should definitely investigate testing in the e2e tests though.

database/redis.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"context"
+	"strings"
+
+	"github.com/ava-labs/avalanchego/utils/logging"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/redis/go-redis/v9"
+	"go.uber.org/zap"
+)
+
```
michaelkaplan13, line 12:
```suggestion

var _ RelayerDatabase = & RedisDatabase{}

```

michaelkaplan13 (dismissed):
All the changes LGTM.

In the future if we want to add support for any other database implementations, we may want to consider making the configuration explicit set a "database type" instead of inferring it from a combination of settings. We could also then have a single `NewRelayerDatabase` method called from `main` that handles the creation of any of the underlying implementations. All just future considerations though.

meeshavax (approved):
👍🏼

database/redis.go
```diff
@@ -0,0 +1,74 @@
...
+func NewRedisDatabase(logger logging.Logger, redisURL string, relayerIDs []RelayerID) (*RedisDatabase, error) {
+	opts, err := redis.ParseURL(redisURL)
+	if err != nil {
+		logger.Error("Failed to parse Redis URL", zap.Error(err), zap.String("url", redisURL))
+		return nil, err
+	}
+
+	// Create a new Redis client.
+	// The server address, password, db index, and protocol version are extracted from the URL
+	// Request timeouts use the default value of 3 seconds
```
minghinmatthewlam, line 32: Looking over `PraseURL` documentation, I think this can be set in the redis url as well, and not always defaulting to 3 seconds. `redis://user:password@localhost:6789/3?dial_timeout=3&db=1&read_timeout=6s&max_retries=2`

database/redis.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (r *RedisDatabase) Get(relayerID common.Hash, key DataKey) ([]byte, error) {
+	ctx := context.Background()
+	compositeKey := constructCompositeKey(relayerID, key)
+	val, err := r.client.Get(ctx, compositeKey).Result()
+	if err != nil {
+		r.logger.Debug("Error retrieving key from Redis",
+			zap.String("key", compositeKey),
+			zap.Error(err))
+		if err == redis.Nil {
```
minghinmatthewlam, line 48: why do we need to return a different error here specifically?

database/redis.go
cam-schultz: If the timeout query parameters are not present, then the `opts` will be use the default value of 3 seconds. If the user passed in the URL above, then `opts` would use those timeouts instead.

database/redis.go
cam-schultz: We have branching logic depending on if the DB returned a key not found error or some other error:
https://github.com/ava-labs/awm-relayer/blob/af25ff13915796cb208044215c9c5c33edf4f9bc/database/database.go#L49
https://github.com/ava-labs/awm-relayer/blob/acb4428b365176b5b89d003b054e66e29fccbd13/relayer/message_relayer.go#L603

main/main.go
```diff
@@ -169,13 +169,25 @@ func main() {
...
 	// Initialize the database
-	db, err := database.NewJSONFileStorage(logger, cfg.StorageLocation, database.GetConfigRelayerIDs(&cfg))
-	if err != nil {
-		logger.Error(
-			"Failed to create database",
-			zap.Error(err),
-		)
-		panic(err)
+	var db database.RelayerDatabase
+	if cfg.RedisURL != "" {
```
minghinmatthewlam, line 173: can we create a new method `NewDatabase` or `NewRelayerDatabase`, pass in the config there, and perform this logic there? The function can return the interface `RelayerDatabase` type

cam-schultz: Good call. Added.

geoff-vball (dismissed):

minghinmatthewlam (dismissed):

database/redis.go
minghinmatthewlam: nit: I'd suggest adding that clarification in the comment. The comment sounds as if it always uses 3 seconds regardless of url

cam-schultz: Clarified

minghinmatthewlam (approved):

Issue #244, opened 2024-03-20
Rename Relayer and messageRelayer types


cam-schultz:
**Context and scope**
As described in https://github.com/ava-labs/awm-relayer/issues/240, in order to support shared database access, we need to persist `messageRelayer` instances and have them handle a unique database key. That key identifies a particular (source chain, destination chain, source address, destination address) tuple. Another way to think about that tuple is that it represents Warp messaging between two distinct applications on different chains. We should replace "message relayer" with "application relayer" to better reflect the use of the type.

Additionally, the current `Relayer` type doesn't actual perform any relaying - this is done by `ApplicationRelayer`s. We should rename `Relayer` to `Listener` to better reflect its function as a subscriber to a particular source blockchain, and manager of the configured `ApplicationRelayer`s

**Discussion and alternatives**
We need to rename:
- `messageRelayer` type to `ApplicationRelayer`
  - Associated variables in the `Relayer` type
- `Relayer` to `Listener` 
- Metric names
- Comments throughout

PR #262, opened 2024-04-17
Rename relayer types

cam-schultz:
## Why this should be merged
Fixes https://github.com/ava-labs/awm-relayer/issues/244

## How this works
Renames relevant types

## How this was tested
CI

## How is this documented
N/A

### DIFF
--- a/relayer/listener.go
+++ b/relayer/listener.go
@@ -363,7 +363,7 @@ func (lstnr *Listener) RelayMessage(warpLogInfo *vmtypes.WarpLogInfo, storeProce
 		zap.String("warpMessageID", unsignedMessage.ID().String()),
 	)
 
-	// Check that the warp message is from a support message protocol contract address.
+	// Check that the warp message is from a supported message protocol contract address.
 	messageManager, supportedMessageProtocol := lstnr.messageManagers[warpLogInfo.SourceAddress]
 	if !supportedMessageProtocol {
 		// Do not return an error here because it is expected for there to be messages from other contracts
@@ -371,6 +371,7 @@ func (lstnr *Listener) RelayMessage(warpLogInfo *vmtypes.WarpLogInfo, storeProce
 		lstnr.logger.Debug(
 			"Warp message from unsupported message protocol address. Not relaying.",
 			zap.String("protocolAddress", warpLogInfo.SourceAddress.Hex()),
+			zap.String("warpMessageID", unsignedMessage.ID().String()),
 		)
 		return nil
 	}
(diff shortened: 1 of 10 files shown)


sample-relayer-config.json
```diff
@@ -9,10 +10,10 @@
       "rpc-endpoint": "https://api.avax-test.network/ext/bc/C/rpc",
       "ws-endpoint": "wss://api.avax-test.network/ext/bc/C/ws",
       "message-contracts": {
-        "0x253b2784c75e510dD0fF1da844684a1aC0aa5fcf": {
+        "0xF7cBd95f1355f0d8d659864b92e2e9fbfaB786f7": {
```
geoff-vball, line 13: Did you mean to include the changes in this file?

geoff-vball (commented):
Just one question, otherwise lgtm

sample-relayer-config.json
cam-schultz: Oops, no I did not. Reverted.

relayer/listener.go
```diff
@@ -156,30 +156,30 @@ func NewRelayer(
...
-		r.catchUpResultChan <- true
+		lstnr.catchUpResultChan <- true
 	}
 
-	return &r, nil
+	return &lstnr, nil
 }
 
-func (r *Relayer) calculateListenerStartingBlockHeight(processHistoricalBlocksFromHeight uint64) (uint64, error) {
+func (lstnr *Listener) calculateListenerStartingBlockHeight(processHistoricalBlocksFromHeight uint64) (uint64, error) {
```
michaelkaplan13, line 180: Nit: Can we rename this `calculateStartingBlockHeight` since the `Listener` is applied in the name now?

relayer/listener.go
```diff
@@ -233,141 +233,141 @@ func (r *Relayer) ProcessLogs(ctx context.Context) error {
...
-	messageRelayer, ok := r.messageRelayers[messageRelayerID]
-	return messageRelayer, ok
+	applicationRelayer, ok := lstnr.applicationRelayers[applicationRelayerID]
+	return applicationRelayer, ok
 }
 
 // RelayMessage relays a single warp message to the destination chain. Warp message relay requests from the same origin chain are processed serially
-func (r *Relayer) RelayMessage(warpLogInfo *vmtypes.WarpLogInfo, storeProcessedHeight bool) error {
-	r.logger.Info(
+func (lstnr *Listener) RelayMessage(warpLogInfo *vmtypes.WarpLogInfo, storeProcessedHeight bool) error {
```
michaelkaplan13, line 344: I find it a little odd that the `Listener` has a `RelayMessage` function (as opposed to that being contained in `ApplicationRelayer`. I don't think we should change anything functionally, but this function is really:
1. Filtering unsupported messages that won't be relayed.
2. Routing supported messages to the proper `ApplicationRelayer` to be relayed.

Any ideas for what we could rename it maybe?

cam-schultz: How about `RouteMessage`?

relayer/listener.go
```diff
@@ -386,53 +388,61 @@ func (r *Relayer) RelayMessage(warpLogInfo *vmtypes.WarpLogInfo, storeProcessedH
...
+		lstnr.logger.Debug(
+			"Application relayer not found. Skipping message relay.",
+			zap.String("blockchainID", lstnr.sourceBlockchain.GetBlockchainID().String()),
 			zap.String("destinationBlockchainID", destinationBlockchainID.String()),
 			zap.String("originSenderAddress", originSenderAddress.String()),
 			zap.String("destinationAddress", destinationAddress.String()),
+			zap.String("warpMessageID", unsignedMessage.ID().String()),
 		)
-		return fmt.Errorf("message relayer not found")
+		return nil
```
cam-schultz, line 421: We should not emit an error when deciding to skip an unsupported Warp message.

michaelkaplan13 (approved):

PR #263, opened 2024-04-18
DB Manager

### LINKED ISSUE #234: Add periodic database write routine

cam-schultz:
## Why this should be merged
Fixes #234 and unblocks #31 

The relaying logic of an Application Relayer is largely parallelizable. Specifically, parsing Warp messages, aggregating signatures, and constructing and issuing transactions require no synchronization. The one operation that does is database access. Note that the `RelayerDatabse` interface assumes that writes are thread safe, but the `RelayerDatabase` interface makes no assumptions about the consistency of concurrent writes. In its current form, this is an issue because concurrent Application Relayer worker threads may complete blocks out of order.

## How this works
To support consistent writes, this PR introduces a _database manager_ layer that prepares and commits writes when it is sure that there are no more pending jobs for a particular block height. It accomplishes this with the following changes:
- `Subscriber`s now subscribe to new blocks, rather than relevant Warp logs. After a new block is received, the number of relevant Warp logs for each application relayer are counted, and then dispatched for processing.
- Adds a `keyManager` type that is initialized with the number of expected jobs, and commits a height to the database once all expected jobs have completed. This decouples database write consistency from the order in which messages are processed. (By comparison, the existing implementation writes the current processing message's height to the database; this requires that messages are processed serially in block order)
- Adds a `DatabaseManager` type that periodically writes committed block heights to the associated key in the database. The write interval is configurable.

## How this was tested
CI

## How is this documented
Updated README


relayer/listener.go
```diff
@@ -371,78 +551,66 @@ func (lstnr *Listener) RouteMessage(warpLogInfo *vmtypes.WarpLogInfo, storeProce
...
-			zap.String("warpMessageID", unsignedMessage.ID().String()),
+			zap.String("warpMessageID", parsedMessageInfo.unsignedMessage.ID().String()),
 			zap.Error(err),
 		)
-		return err
 	}
 
 	// Increment the request ID for the next message relay request
 	lstnr.currentRequestID++
-	return nil
```
cam-schultz, line 447: This was an existing bug.

database/manager.go
```diff
@@ -0,0 +1,181 @@
...
+		return fmt.Errorf("key manager not found")
+	}
+	km.prepareHeight(height, totalMessages)
+	return nil
+}
+
+func (dm *DatabaseManager) Finished(id RelayerID, height uint64) {
+	km, ok := dm.keyManagers[id]
+	if !ok {
+		// This is not an error, as this will occur if the database is not eligible for writing
```
geoff-vball, line 89: I don't quite understand this comment/case. Why wouldn't a keyManager exist for a particular relayerID?

database/manager.go
```diff
@@ -0,0 +1,181 @@
...
+		counter, ok := km.queuedHeightsAndMessages[height]
+		if !ok {
+			continue
+		}
+		counter.processedMessages++
+		if counter.processedMessages == counter.totalMessages {
+			km.commitHeight(height)
+			delete(km.queuedHeightsAndMessages, height)
+		}
+	}
```
geoff-vball, line 133: So I'm pretty sure we have a race condition in here. Lets say the current highest processed height is 99 and we are currently processing blocks 100 and 101. If 101 finished processing before block 100, we will update the highest processed block to 101, while not having processed 100 yet. We have to check `queuedHeightsAndMessages` to make sure there is no lower height being processed. 

We could do something like... Let's say our highest processed block is N. If the height we're committing is N+1, set the value to that. Otherwise, throw that block height in a min heap. Then when we finally process block N+1, we look in the min heap for N+2, and keep removing values from the heap as long as we have the next block in sequence.

database/manager.go
```diff
@@ -0,0 +1,181 @@
...
+// PrepareHeight sets the total number of messages to be processed at a given height.
+// Once all messages have been processed, the height is eligible to be committed.
+// It is up to the caller to determine if a height is eligible to be committed.
+// This function is thread safe.
+func (km *keyManager) prepareHeight(height uint64, totalMessages uint64) {
+	km.logger.Debug("preparing height", zap.Uint64("height", height))
+	if totalMessages == 0 {
+		km.maxHeightLock.Lock()
+		defer km.maxHeightLock.Unlock()
+		km.commitHeight(height)
```
geoff-vball, line 153: This lock should be taken and released inside `commitHeight`. Right now only one of the calls to `commitHeight` is grabbing the lock.

database/manager.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+}
+
+// commitHeight marks a height as eligible to be written to the database.
+func (km *keyManager) commitHeight(height uint64) {
+	if height > km.maxCommittedHeight {
+		km.logger.Debug("committing height", zap.Uint64("height", height))
+		km.maxCommittedHeight = height
+	}
+}
```
geoff-vball, line 142: `commitHeight` sounds to me like it should be writing to the database. Maybe `heightProcessed`?

geoff-vball (commented):
Still going through, but will submit these comments for you to start thinking about

geoff-vball (commented):
A lot of the comments assume we're going to be processing logs async, but I know they're currently still synchronous per source chain.

README.md
```diff
@@ -246,7 +258,9 @@ The relayer consists of the following components:
 - At the global level:
   - P2P app network: issues signature `AppRequests`
   - P-Chain client: gets the validators for a Subnet
-  - JSON database: stores latest processed block for each Application Relayer
+  - Relayer database: stores latest processed block for each Application Relayer
+    - Currently supports Redis and local JSON file storage
```
minghinmatthewlam, line 262: does it by default use JSON file storage

database/manager.go
```diff
@@ -0,0 +1,181 @@
...
+		interval:    interval,
+		keyManagers: keyManagers,
+	}
+}
+
+// Run writes all committed keys to the database on the configured timer.
+// This function should only be called once.
+func (dm *DatabaseManager) Run() {
+	for range time.Tick(dm.interval) {
+		for id, km := range dm.keyManagers {
```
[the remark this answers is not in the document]
minghinmatthewlam: actually doing a quick readup seems like the ticks continue in background, and if the underlying for loop is not completed, then the tick iteration will get dropped and waits for the next. Is that a problem at all for short durations?

database/manager.go
```diff
@@ -0,0 +1,181 @@
...
+				}
+				if storedHeight >= km.maxCommittedHeight {
+					continue
+				}
+				dm.logger.Debug(
+					"writing height",
+					zap.Uint64("height", km.maxCommittedHeight),
+					zap.String("relayerID", id.ID.String()),
+				)
+				err = dm.db.Put(id.ID, LatestProcessedBlockKey, []byte(strconv.FormatUint(km.maxCommittedHeight, 10)))
```
minghinmatthewlam, line 62: can we add a comment in the interface definition that the functions handle syncing with locks, etc.

database/manager.go
```diff
@@ -0,0 +1,181 @@
...
+				if err != nil {
+					dm.logger.Error("Failed to write latest processed block height", zap.Error(err))
```
minghinmatthewlam, line 64: do the key managers have some identifier that can be logged here too?

minghinmatthewlam: how are you thinking of detecting if one of many key managers are in a stuck state and can't erroring when going to its iterations?

database/manager.go
```diff
@@ -0,0 +1,181 @@
...
+		dm.logger.Debug("Key manager not found", zap.String("relayerID", id.ID.String()))
+		return
+	}
+	km.finished <- height
+}
+
+//
+// keyManager commits keys to be written to the database in a thread safe manner.
+//
+
```
minghinmatthewlam, line 99: extra line

database/manager.go
```diff
@@ -0,0 +1,181 @@
...
+			km.commitHeight(height)
+			delete(km.queuedHeightsAndMessages, height)
+		}
+	}
+}
+
+// commitHeight marks a height as eligible to be written to the database.
+func (km *keyManager) commitHeight(height uint64) {
+	if height > km.maxCommittedHeight {
+		km.logger.Debug("committing height", zap.Uint64("height", height))
```
minghinmatthewlam, line 139: nit: capitalize log messages here and throughout some places in file

database/manager.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+}
+
+// Run listens for finished signals from application relayers, and commits the
+// height once all messages have been processed.
+// This function should only be called once.
+func (km *keyManager) run() {
+	for height := range km.finished {
+		counter, ok := km.queuedHeightsAndMessages[height]
+		if !ok {
```
minghinmatthewlam, line 125: should this ever happen? From what I can tell, the height should be committed if it were finished. If so, can you log an error?

database/manager.go
```diff
@@ -0,0 +1,181 @@
...
+			continue
+		}
+		counter.processedMessages++
+		if counter.processedMessages == counter.totalMessages {
+			km.commitHeight(height)
+			delete(km.queuedHeightsAndMessages, height)
```
minghinmatthewlam, line 131: do we need the lock before deleting?

database/manager.go
```diff
@@ -0,0 +1,181 @@
...
+		finished:                 make(chan uint64),
+	}
+}
+
+// Run listens for finished signals from application relayers, and commits the
+// height once all messages have been processed.
+// This function should only be called once.
+func (km *keyManager) run() {
+	for height := range km.finished {
+		counter, ok := km.queuedHeightsAndMessages[height]
```
minghinmatthewlam, line 124: would we want a read/write lock, and to acquire readlock here?

database/manager.go
```diff
@@ -0,0 +1,181 @@
...
+
+//
+// keyManager commits keys to be written to the database in a thread safe manner.
+//
+
+type keyManager struct {
+	logger                   logging.Logger
+	queuedHeightsLock        *sync.Mutex
+	queuedHeightsAndMessages map[uint64]*messageCounter
+	maxHeightLock            *sync.Mutex
```
bernard-avalabs, line 104: Is having a separate queuedHeightsLock and maxHeightLock necessary from a correctness/performance perspective? Both locks are held only very briefly, so separating into separate locks would probably not make a significant performance difference. 

Usually a "monitor" based model where this is a single lock per object and the lock is acquired at the start of its member functions, is less error-prone compared to more fine-grained locking approaches.

database/manager.go
bernard-avalabs: If I'm understanding the code correctly, it does look like the commitHeight function needs to do some buffering to avoid gaps in the committed block height.

A circular buffer could be used in place of a min heap if the difference between the highest processed block block and the current height is bounded.

database/manager.go
```diff
@@ -0,0 +1,181 @@
...
+	"go.uber.org/zap"
+)
+
+//
+// DatabaseManager writes all committed keys to the database on a timer.
+//
+
+type DatabaseManager struct {
+	logger      logging.Logger
+	keyManagers map[RelayerID]*keyManager
```
bernard-avalabs, line 22: Do we plan to allow additional keyManagers to be added dynamically? It doesn't look like we do right now, but if we want to add that later, we will need to add a mutex in DatabaseManager.

bernard-avalabs (commented):
I've left a few concurrency-related comments.

database/manager.go
cam-schultz: That's a great catch. I've implemented the min-heap based approach you suggested.

On a related note, I'm debating how we should handle the case in which a block height never calls `commitHeight`. In that case, the min-heap would grow unbounded, and no new data would be written to the DB. One approach would be to kill the entire application if the min-heap grows too large, and attempt to re-process the missing block on restart. If the issue persists, the operator would have to intervene by either setting `process-missed-blocks` to false, or manually updating the DB. One thing I like about this approach is that all blocks up to the height stored in the DB are guaranteed to be processed; only with manual intervention can this be made untrue.

database/manager.go
```diff
@@ -0,0 +1,181 @@
...
+		db:          db,
+		interval:    interval,
+		keyManagers: keyManagers,
+	}
+}
+
+// Run writes all committed keys to the database on the configured timer.
+// This function should only be called once.
+func (dm *DatabaseManager) Run() {
+	for range time.Tick(dm.interval) {
```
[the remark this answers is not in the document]
cam-schultz: I think this is captured implicitly in the function's comment that it should only be called once.

database/manager.go
```diff
@@ -0,0 +1,181 @@
...
+		for id, km := range dm.keyManagers {
+			// Ensure we're not writing the default value
+			if km.maxCommittedHeight > 0 {
+				storedHeight, err := getLatestProcessedBlockHeight(dm.db, id)
+				if err != nil && !IsKeyNotFoundError(err) {
+					dm.logger.Error("Failed to get latest processed block height", zap.Error(err))
```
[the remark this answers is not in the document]
cam-schultz: I think an in-memory cache would be appropriate to include in the `RelayerDatabase` implementations. It would be good to verify through profiling if disk reads are even a bottleneck worth optimizing.

In any case, we need to have some kind of error check here. `IsKeyNotFoundError` is meant to filter expected versus unexpected errors.

database/manager.go
cam-schultz: I think this is out of date and pertains to a previous design. Removed the comment and upgraded the log to `Error`

relayer/listener.go
```diff
@@ -194,14 +248,44 @@ func (lstnr *Listener) calculateStartingBlockHeight(processHistoricalBlocksFromH
 
 func (lstnr *Listener) setAllProcessedBlockHeightsToLatest() error {
 	for _, relayer := range lstnr.applicationRelayers {
-		_, err := relayer.setProcessedBlockHeightToLatest()
+		height, err := lstnr.getCurrentHeight()
+		if err != nil {
+			return err
+		}
+		relayer.dbManager.PrepareHeight(relayer.relayerID, height, 0)
```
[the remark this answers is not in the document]
cam-schultz: `keyManager.prepareHeight` immediately commits the height if the number of messages to process is `0`. I don't think we should directly expose `commitHeight` to `dbManager` callers. I can add separate method with a clearer name that makes this call internally though.

relayer/listener.go
```diff
@@ -337,32 +516,33 @@ func (lstnr *Listener) getApplicationRelayer(
 		database.AllAllowedAddress,
 		database.AllAllowedAddress,
 	)
-	applicationRelayer, ok := lstnr.applicationRelayers[applicationRelayerID]
```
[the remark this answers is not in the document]
cam-schultz: Sorry, I'm not quite following your suggestion. The four checks are already in a helper `getApplicationRelayer`. Are you suggesting splitting this function up further?

vms/evm/subscriber.go
```diff
@@ -179,55 +117,55 @@ func (s *subscriber) ProcessFromHeight(height *big.Int, done chan bool) {
 	done <- true
 }
 
+func (s *subscriber) newWarpBlockInfo(header *types.Header, isCatchUp bool) (*relayerTypes.WarpBlockInfo, error) {
+	var (
+		logs []types.Log
+		err  error
+	)
+	if header.Bloom.Test(warpPrecompileLogFilter[:]) {
```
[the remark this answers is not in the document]
cam-schultz: Yup exactly. False positives are ok, since the subsequent call to `ethclient.FilterLogs` would just return empty.

database/manager.go
```diff
@@ -0,0 +1,181 @@
...
+	"time"
+
+	"github.com/ava-labs/avalanchego/utils/logging"
+	"go.uber.org/zap"
+)
+
+//
+// DatabaseManager writes all committed keys to the database on a timer.
+//
+
```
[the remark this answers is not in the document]
cam-schultz: This line is necessary to prevent the go auto-formatter from cropping the empty comment lines. I wanted to make it very obvious what parts of this file corresponded to what types.

database/manager.go
cam-schultz: See https://github.com/ava-labs/awm-relayer/pull/263#discussion_r1578589613

database/manager.go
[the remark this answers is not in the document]
cam-schultz: The tick interval is on the order of seconds (or 10's of seconds), so the only expected scenario this could occur is if there's network issues reaching the remote database. Even so, this isn't really a problem since a missed write to the database is perfectly safe.

database/manager.go
cam-schultz: Added a comment to `RelayerDatabase`.

database/manager.go
cam-schultz: Yes definitely, nice catch

database/manager.go
cam-schultz: Good call, added this in.

config/config.go
```diff
@@ -147,6 +148,7 @@ func SetDefaultConfigValues(v *viper.Viper) {
 	v.SetDefault(ProcessMissedBlocksKey, true)
 	v.SetDefault(APIPortKey, 8080)
 	v.SetDefault(MetricsPortKey, 9090)
+	v.SetDefault(DBWriteIntervalSecondsKey, 1)
```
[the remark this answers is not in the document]
cam-schultz: The performance should actually be the same regardless of the write interval, since we always write the same number of keys.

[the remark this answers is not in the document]
cam-schultz: 10 seconds seems reasonable to me. Only under very extreme edge cases would the relayer not be able to make progress on _some_ timescale, so the more important question is what's the oldest data we're willing to have in the relayer database?

types/types.go
```diff
@@ -15,5 +16,10 @@ type WarpLogInfo struct {
 	SourceTxID       []byte
 	UnsignedMsgBytes []byte
 	BlockNumber      uint64
-	IsCatchUpMessage bool
+}
+
+type WarpBlockInfo struct {
```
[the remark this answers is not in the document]
cam-schultz: Added additional commente.

With the current form of `ProcessLogs` we'll still need `IsCatchUpBlock`. At minimum, `ProcessLogs` would need to be modified to count all of the Warp messages to be processed for all of the catch-up blocks in order to fully populate the `keyManager` pending queues. If we do that, then the write validity would be fully encapsulated by the `DatabaseManager` and `keyManager`. I will explore making that change in this PR, since it should be equally valid here.

relayer/listener.go
```diff
@@ -220,28 +304,73 @@ func (lstnr *Listener) ProcessLogs(ctx context.Context) error {
...
+					if err != nil {
+						lstnr.logger.Error(
+							"Failed to prepare height",
+							zap.String("relayerID", relayerID.ID.String()),
+							zap.Error(err),
+						)
+						continue
+					}
+				}
+			}
```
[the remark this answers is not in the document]
cam-schultz: I agree - `keyManager` and `applicationRelayer` are very tightly coupled. I'll refactor these types to better reflect this.

database/manager.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+}
+
+// Run writes all committed keys to the database on the configured timer.
+// This function should only be called once.
+func (dm *DatabaseManager) Run() {
+	for range time.Tick(dm.interval) {
+		for id, km := range dm.keyManagers {
+			// Ensure we're not writing the default value
+			if km.maxCommittedHeight > 0 {
```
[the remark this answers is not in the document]
cam-schultz: Added RW locks to all accesses of `maxCommittedHeight` and `queuedHeightsAndMessages`

database/manager.go
cam-schultz: No, the set of `keyManagers` is fully specified by the configuration (as is the case with the set of application relayers that interact with the `keyManagers`)

database/manager.go
cam-schultz: Great call. I've simplified the class to have a single object lock.

database/manager.go
geoff-vball: Hmmm... Yeah I wonder what the best approach would be here to handle the case of a missing block. I don't really want to kill the whole application for one missing block for one application.

The approach that is coming to mind is setting some threshold, say 100 blocks, that if the min heap grows bigger than, we just pop the min value from the heap and set that as our most recently processed block, and throw an error for any missed blocks. We could also write the missed blocks to a table somewhere for potential reprocessing later.

relayer/listener.go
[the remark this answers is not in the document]
geoff-vball: So with the min-heap implemented, `prepareheight` with 0 messages should only write the height if the current height is N and we commit N+1. We need a separate helper to force-set the height to latest here

relayer/listener.go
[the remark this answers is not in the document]
geoff-vball: Yeah, definitely not necessary, but I see this function as two separate sub-steps. 

- Parse unsigned warp message
- Get application relayer

relayer/listener.go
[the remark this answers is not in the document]
cam-schultz: `commitHeight` automatically commits if `km.maxCommittedHeight == 0`, which is the case on initialization, and should be the case when this function is called, since it is called before the main processing loop in `ProcessLogs` is initiated.

I agree that this is not at all clear though. I'll add a more specific method that can be called here, and update `commitHeight` to accept a "force" option.

types/types.go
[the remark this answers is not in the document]
geoff-vball: @cam-schultz Not sure if you wrote this comment before reading through my feedback, but if we implement the min-heap for updating the most recent block, we shouldn't need to know if a block is a catchup block, since the height will just be added to the min-heap, and the non-catchup values will be processed out of the min-heap once all the catchup blocks have been processed.

database/manager.go
geoff-vball: Yeah I like the monitor pattern better unless we have a strong reason for separate locks 😄

database/manager.go
```diff
@@ -0,0 +1,267 @@
...
+	*h = append(*h, x.(uint64))
+}
+
+func (h *intHeap) Pop() any {
+	old := *h
+	n := len(old)
+	x := old[n-1]
+	*h = old[0 : n-1]
+	return x
+}
```
michaelkaplan13, line 128: Could we move this to it's own package/file?

database/manager.go
```diff
@@ -0,0 +1,267 @@
...
+
+//
+// keyManager commits keys to be written to the database in a thread safe manner.
+//
+
+type keyManager struct {
```
michaelkaplan13, line 134: Small suggestion that may be missing some context, but would `BlockCheckpointManager` be a more descriptive name rather than `keyManager`? i.e. each `ApplicationRelayer` instance would have a `BlockCheckpointManager` that it uses to determine the correct value to write to the database.

types/types.go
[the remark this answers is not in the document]
cam-schultz: You're absolutely right. Requiring in-order writes turns out to be a powerful invariant. I've made this change now.

README.md
minghinmatthewlam: Similar to how the other config values specifies the default value, can you add that for storage

database/manager.go
```diff
@@ -0,0 +1,267 @@
...
+	km, ok := dm.keyManagers[id]
+	if !ok {
+		dm.logger.Error("Key manager not found", zap.String("relayerID", id.ID.String()))
+		return
+	}
+	km.finished <- height
+}
+
+// intHeap adapted from https://pkg.go.dev/container/heap#example-package-IntHeap
+type intHeap []uint64
```
geoff-vball, line 109: Can we rename this to `uint64Heap`? The typecheck on line 119 will panic if we provide any other type of integer.

geoff-vball: Alternatively, there's a generic heap package [here](https://pkg.go.dev/github.com/fufuok/utils/generic/heap#section-readme)

README.md
cam-schultz: The default values and override behavior is already documented in the `Configuration` section. This section here is intended to give a high level overview of the architectural components.

database/manager_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/ava-labs/avalanchego/utils/logging"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCommitHeight(t *testing.T) {
+	testCases := []struct {
+		name              string
+		currentMaxHeight  uint64
+		commitHeight      uint64
+		pendingHeights    *intHeap
```
geoff-vball: nit: Does this have to be a pointer?

database/manager.go
```diff
@@ -79,178 +71,32 @@ func (dm *DatabaseManager) Run() {
...
-func (h intHeap) Peek() uint64       { return h[0] }
-
-func (h *intHeap) Push(x any) {
-	// Push and Pop use pointer receivers because they modify the slice's length,
-	// not just its contents.
-	*h = append(*h, x.(uint64))
+func (dm *DatabaseManager) CommitHeight(id RelayerID, height uint64) {
+	dm.committedHeightsLock.Lock()
+	defer dm.committedHeightsLock.Unlock()
+	dm.committedHeights[id] = height
```
geoff-vball: Bear with me... But I don't think we should grab the lock here. I think it's okay to lock in `RegisterRelayerID`, and then grabbing a read lock while iterating over them so that the map doesn't get any new keys while iterating, but I think in this function, where we should only be updating the value for existing keys, we don't need to lock. Right now, this function will be locked out during the entire execution of the update tick when I don't think that serves any purpose.

database/manager.go
```diff
@@ -79,178 +71,32 @@ func (dm *DatabaseManager) Run() {
...
-func (h *intHeap) Pop() any {
-	old := *h
-	n := len(old)
-	x := old[n-1]
-	*h = old[0 : n-1]
-	return x
+func (dm *DatabaseManager) GetCommittedHeight(id RelayerID) uint64 {
+	dm.committedHeightsLock.RLock()
+	defer dm.committedHeightsLock.RUnlock()
+	return dm.committedHeights[id]
```
geoff-vball: I also don't think we should be locking here

relayer/application_relayer.go
```diff
@@ -96,25 +95,29 @@ func newApplicationRelayer(
 		// Otherwise, the source subnet signs the message.
 		signingSubnet = sourceBlockchain.GetSubnetID()
 	}
+
+	sub := ticker.Subscribe()
+	checkpointManager := newCheckpointManager(logger, db, sub, relayerID)
+	checkpointManager.run()
```
geoff-vball, line 101: If `run()` should only be called once, maybe we should call it at the end of `newCheckpointManager`?

relayer/checkpoint.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+func (cm *checkpointManager) run() {
+	go cm.handleFinishedRelays()
+	go cm.writeToDatabase()
+}
+
+func (cm *checkpointManager) writeToDatabase() {
+	for range cm.writeSignal {
+		cm.lock.RLock()
```
geoff-vball, line 55: I would encapsulate the body of this for loop into a helper function. That way, you can call lock and then defer the unlock. I think almost any critical section should always be it's own function

relayer/checkpoint.go
```diff
@@ -0,0 +1,206 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package relayer
```
geoff-vball, line 4: Another thing to consider is to put this file into its own package (nested inside relayer). That way you can more finely tune which methods are available for the relayer to call on it, and which methods are just private helper functions that shouldn't be called outside of this file.

relayer/listener.go
```diff
@@ -285,17 +413,41 @@ func (lstnr *Listener) ReconnectToSubscriber() error {
 	return nil
 }
 
-// Fetch the appropriate application relayer
+// RouteMessage relays a single warp message to the destination chain.
+// Warp message relay requests from the same origin chain are processed serially
+func (lstnr *Listener) RouteMessage(warpLogInfo *relayerTypes.WarpLogInfo) error {
```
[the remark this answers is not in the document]
cam-schultz: Yeah... I'll take a pass through the updated methods for naming consistency.

relayer/application_relayer.go
cam-schultz: I elected to keep them separate to give the caller greater control over the checkpoint manager processing loop. For instance, the tests in `checkpoint_test.go` only call `newCheckpointManager`. Requiring the unit tests to also run the processing loop seems like overkill, and potentially less stable.

relayer/listener.go
```diff
@@ -217,31 +300,76 @@ func (lstnr *Listener) ProcessLogs(ctx context.Context) error {
...
+				appRelayer.checkpointManager.prepareHeight(block.BlockNumber, totalMessages)
+			}
+			for _, msgInfo := range msgsInfo {
+				lstnr.logger.Info(
+					"Relaying message",
+					zap.String("sourceBlockchainID", lstnr.sourceBlockchain.GetBlockchainID().String()),
+					zap.String("warpMessageID", msgInfo.unsignedMessage.ID().String()),
 				)
-				continue
+				err := lstnr.dispatchToApplicationRelayer(msgInfo, block.BlockNumber)
```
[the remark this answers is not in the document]
cam-schultz: I agree with the suggested changes, but if you don't mind, I'd like to defer to #31. There's a decent amount of refactoring for `Listener` and `ApplicationRelayer` that we'll have to do to support that change. The current incoming `AppResponse` routing logic assumes that at most one message per `Listener` is being processed at any given time, so we'll need to update that routing logic to account for concurrent message processing. My thought right now is to update the [external handler](https://github.com/ava-labs/awm-relayer/blob/eb9162b4233c4d995dbc1086834d50e4317b3271/peers/external_handler.go#L30) to route directly to `ApplicationRelayer`s instead of indirectly through the `Listener`. That'll require some changes to both types, so would rather make all the intended changes in one pass.

relayer/checkpoint.go
cam-schultz: Good call, I've updated the methods so that locks/unlocks are performed at function scope.

relayer/checkpoint.go
cam-schultz: I'm going to leave the packages as-is for now. #31 will require some refactoring of `Listener` and `ApplicationRelayer` that should result in the types becoming less coupled. At that point, all of the types in `package relayer` can likely be moved into their own packages.

database/ticker.go
```diff
@@ -0,0 +1,31 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package database
+
+import "time"
+
+type Ticker struct {
```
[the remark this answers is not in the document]
cam-schultz: Moved to `utils`

relayer/listener.go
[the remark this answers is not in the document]
cam-schultz: On second thought, I think the name `RouteMessage` is appropriate for now. #31 will require some refactoring to `Listener` and `ApplicationRelayer` that should make it possible for `ApplicationRelayer` methods to be called directly from `main`.

database/manager_test.go
geoff-vball: You don't need a pointer to use pointer receivers, do you?

relayer/checkpoint.go
```diff
@@ -0,0 +1,210 @@
...
+	for range cm.writeSignal {
+		cm.writeToDatabase()
+	}
+}
+
+func (cm *checkpointManager) incrementFinishedCounter(height uint64) {
+	cm.lock.Lock()
+	defer cm.lock.Unlock()
+	counter, ok := cm.queuedHeightsAndMessages[height]
+	if !ok {
```
minghinmatthewlam, line 98: should this case ever happen? Confirming when this happens right now we just return and wait for the next finished block height?

relayer/listener.go
```diff
@@ -217,31 +301,76 @@ func (lstnr *Listener) ProcessLogs(ctx context.Context) error {
...
 			}
-		case txLog := <-lstnr.Subscriber.Logs():
-			// Relay the message to the destination chain. Continue on failure.
+		case block := <-lstnr.Subscriber.Blocks():
+			// Relay the messages in the block to the destination chains. Continue on failure.
 			lstnr.logger.Info(
-				"Handling Teleporter submit message log.",
-				zap.String("txId", hex.EncodeToString(txLog.SourceTxID)),
+				"Processing block",
 				zap.String("originChainId", lstnr.sourceBlockchain.GetBlockchainID().String()),
```
minghinmatthewlam, line 309: nit: are we going with `sourceBlockchainID`?

relayer/listener.go
```diff
@@ -177,11 +203,40 @@ func NewListener(
 	return &lstnr, nil
 }
 
+// Gets the height of the chain head.
+func (lstnr *Listener) getCurrentHeight() (uint64, error) {
+	ethClient, err := ethclient.Dial(lstnr.sourceBlockchain.RPCEndpoint)
```
minghinmatthewlam, line 208: can we save the ethclient in `Listener` instead of dialing for every call?

tests/utils/utils.go
```diff
@@ -102,8 +102,13 @@ func CreateDefaultRelayerConfig(
 	fundedAddress common.Address,
 	relayerKey *ecdsa.PrivateKey,
 ) config.Config {
+	logLevel := logging.Info
+	if os.Getenv("LOG_LEVEL") == "debug" {
```
minghinmatthewlam, line 106:
```suggestion
	if os.Getenv("LOG_LEVEL") == logging.Debug.LowerString() {
```

minghinmatthewlam: actually should we do default to `logging.Info`, if env variable is not empty string, set to `logging.ToLevel`?

relayer/checkpoint.go
cam-schultz: This is actually expected to happen when processing manual Warp messages. In that case, the message is not associated with a block height, so it does not go through the usual prepare-complete step as normal Warp messages. I've demoted this to a debug log and added a comment explaining this.

And to clarify, when this case is hit, we no-op and continue listening for finished relay signals.

relayer/listener.go
```diff
@@ -61,7 +70,34 @@ func NewListener(
...
-	sub := vms.NewSubscriber(logger, sourceBlockchain)
+	blockchainID, err := ids.FromString(sourceBlockchain.BlockchainID)
+	if err != nil {
+		logger.Error(
+			"Invalid blockchainID provided to subscriber",
+			zap.Error(err),
+		)
+		return nil, err
+	}
+	ethWSClient, err := ethclient.Dial(sourceBlockchain.WSEndpoint)
```
minghinmatthewlam, line 81: nit: since ethWSClient doesn't seem to be anywhere else in the file, we can create the client in `NewSubscriber`

relayer/checkpoint.go
```diff
@@ -0,0 +1,212 @@
...
+func newCheckpointManager(logger logging.Logger, database database.RelayerDatabase, writeSignal chan struct{}, relayerID database.RelayerID) *checkpointManager {
+	h := &utils.UInt64Heap{}
+	heap.Init(h)
+	return &checkpointManager{
+		logger:                   logger,
+		database:                 database,
+		writeSignal:              writeSignal,
+		relayerID:                relayerID,
+		queuedHeightsAndMessages: make(map[uint64]*messageCounter),
+		lock:                     &sync.RWMutex{},
```
geoff-vball, line 42: So I think we need to initialize the checkpointManager with the correct starting block height as `committedHeight`. Right now we have a race condition where the first call to `stageCommittedHeight()` will set `committedHeight` to that height. So if we process a block from the chain head before any of the catchup blocks, we will write the chain head to the database accidentally

relayer/checkpoint.go
```diff
@@ -0,0 +1,212 @@
...
+	if cm.committedHeight == 0 {
+		cm.logger.Debug(
+			"Committing initial height",
+			zap.Uint64("height", height),
+			zap.String("relayerID", cm.relayerID.ID.String()),
+		)
+		cm.committedHeight = height
+		return
+	}
+
```
geoff-vball, line 147: We actually need to ensure that `height` > `cm.committedHeight` here. I think with the race condition i described above it's currently possible. I think it should at least emit an error log, as it means we've probably made a logic error here. A `height` <= `cm.committedHeight` would mean we would never pop anything from the heap.

relayer/listener.go
```diff
@@ -177,11 +215,30 @@ func NewListener(
 	return &lstnr, nil
 }
 
+// Gets the height of the chain head.
+func (lstnr *Listener) getCurrentHeight() (uint64, error) {
+	latestBlock, err := lstnr.ethClient.BlockNumber(context.Background())
```
geoff-vball, line 220: Can we just return this line?

relayer/listener.go
```diff
@@ -194,60 +251,130 @@ func (lstnr *Listener) calculateStartingBlockHeight(processHistoricalBlocksFromH
...
-			err := lstnr.RouteMessage(&txLog, doneCatchingUp || txLog.IsCatchUpMessage)
-			if err != nil {
-				lstnr.logger.Error(
-					"Error relaying message",
-					zap.String("originChainID", lstnr.sourceBlockchain.GetBlockchainID().String()),
-					zap.Error(err),
+			// Iterate over the Warp logs in two passes:
+			// The first pass extracts the information needed to relay from the log, but does not initiate relaying
+			// This is so that the number of messages to be processed can be registered with the database before
+			// any messages are processed
```
geoff-vball, line 318:
```suggestion
			// any messages are processed.
```

relayer/listener.go
```diff
@@ -371,78 +521,76 @@ func (lstnr *Listener) RouteMessage(warpLogInfo *vmtypes.WarpLogInfo, storeProce
...
+		messageManager:     messageManager,
+		applicationRelayer: applicationRelayer,
+	}, nil
+}
 
+func (lstnr *Listener) dispatchToApplicationRelayer(parsedMessageInfo *parsedMessageInfo, blockNumber uint64) error {
 	// Relay the message to the destination. Messages from a given source chain must be processed in serial in order to
 	// guarantee that the previous block (n-1) is fully processed by the listener when processing a given log from block n.
 	// TODO: Add a config option to use the Warp API, instead of hardcoding to the app request network here
-	err = applicationRelayer.relayMessage(
```
geoff-vball, line 427: I think the comment above is now out of date

tests/allowed_addresses.go
```diff
@@ -187,6 +188,8 @@ func AllowedAddresses(network interfaces.LocalNetwork) {
 	)
 	height1, err := subnetAInfo.RPCClient.BlockNumber(ctx)
 	Expect(err).Should(BeNil())
+	// Sleep for some time to make sure the DB is updated
+	time.Sleep(2 * time.Second)
```
geoff-vball, line 192: nit: can we extract this timing to be `2 * databaseUpdatePeriod` or something?

types/types.go
```diff
@@ -0,0 +1,28 @@
...
+	WarpLogs    []types.Log
+}
+
+// WarpLogInfo describes the transaction information for the Warp message
+// sent on the source chain, and includes the Warp Message payload bytes
+// WarpLogInfo instances are either derived from the logs of a block or
+// from the manual Warp message information provided via configuration
+type WarpLogInfo struct {
+	SourceAddress    common.Address
+	SourceTxID       []byte
```
geoff-vball, line 25: I don't think `SourceTxID` is used anywhere. If it is, should it be a `common.Hash`?

vms/evm/subscriber.go
```diff
@@ -179,55 +117,55 @@ func (s *subscriber) ProcessFromHeight(height *big.Int, done chan bool) {
...
+		BlockNumber: header.Number.Uint64(),
+		WarpLogs:    logs,
+	}, nil
+}
+
 // Filter logs from the latest processed block to the latest block
 // Since initializationFilterQuery does not modify existing fields of warpFilterQuery,
 // we can safely reuse warpFilterQuery with only a shallow copy
 func (s *subscriber) processBlockRange(
-	ethClient ethclient.Client,
```
geoff-vball, line 186: Nit: In the function comment can we include that it process [fromBlock, toBlock] inclusive?

relayer/listener.go
cam-schultz: We pass in an ethClient in the constructor so that we can mock it in the unit tests.

bernard-avalabs (commented):
Left a concurrency-related comment/question.

relayer/checkpoint.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+}
+
+func (cm *checkpointManager) run() {
+	go cm.listenForFinishedRelays()
+	go cm.listenForWriteSignal()
+}
+
+func (cm *checkpointManager) writeToDatabase() {
+	cm.lock.RLock()
```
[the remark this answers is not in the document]
geoff-vball: Strictly speaking, the read lock isn't necessary here, as long as we copy `cm.committedHeight` and use the same value throughout the function. `cm.committedHeight` is monotonically increasing, and we write periodically and don't care about a little lossiness, so we don't really care about other threads reading/writing to it.

We only have one thread per relayer that can write, but even if we had more, I don't think it would matter.

`database.LatestProcessedBlockKey` is a const so can't be modified. Unless I misinterpreted your question.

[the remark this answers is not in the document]
geoff-vball: Actually, to answer most of your questions more succinctly... If two threads have the read lock simultaneously, they necessarily have the same values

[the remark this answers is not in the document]
cam-schultz: > For my own understanding: we are acquiring a read lock here to protect fields such as committedHeight, but it is also safe to perform write operations to the database (e.g., cm.database.Put) since the database has its own concurrency control mechanism?

Yes, write concurrency control is assumed to be provided by the `RelayerDatabase` implementation. 

> Two concurrent threads calling writeToDatabase would potentially call cm.database.Put in a different order than their reads to fields in earlier parts of the function. This looks safe, but I just want to confirm.

Within a single `awm-relayer` process, there is a one-to-one mapping between database keys and checkpoint managers, so there should only ever be a single thread writing to a key at any given time. Therefore, the scenario you described should not be possible _within a single `awm-relayer` process_. However, it's possible to deploy a fleet of relayers that all share the same database instance, in which case concurrent writes to the same key are possible. We rely on the database's implementation to provide write consistency in that scenario.

> Can database.LatestProcessedBlockKey be modified by a different concurrent thread? If it can, then it appears we are reading it in our cm.database.Put without providing mutual exclusion unless it is meant to be protected by cm.lock.

Hm that's a good call-out. As mentioned above, only one thread per process is able to access a particular key at a time, but it may be worth adding an `OrderedPut` or a `SafePut` that accepts a predicate and only writes if the predicate passes. We could then more easily atomically perform the check in this funciton. I think there are more considerations here than are worth including in this PR, but I'll create a ticket to track.

types/types.go
cam-schultz: Removed

relayer/checkpoint.go
cam-schultz: Great call out. I've refactored the checkpoint manager constructor to initialize `committedHeight`. We fetch this for each application relayer using the `CalculateStartingBlockHeight`, which I've moved to `database/utils.go`. The `Listener` begins listening from the minimum starting block height of all its composite application relayers. The checkpoint manager will only prepare heights that are greater than its `committedHeight` to account for possible re-processing of already committed blocks on startup.

relayer/checkpoint.go
cam-schultz: See https://github.com/ava-labs/awm-relayer/pull/263#discussion_r1585291255

database/utils.go
```diff
@@ -0,0 +1,83 @@
...
+//   - In this case, we return the configured processHistoricalBlocksFromHeight
+//
+// 3) The database does not contain any information for the chain.
+//   - In this case, we return the configured processHistoricalBlocksFromHeight if it is set, otherwise
+//     we return the chain head.
+func CalculateStartingBlockHeight(
+	logger logging.Logger,
+	db RelayerDatabase,
+	relayerID RelayerID,
+	processHistoricalBlocksFromHeight uint64, currentHeight uint64) (uint64, error) {
```
geoff-vball, line 33: nit: can you put the return values on a newline?

relayer/checkpoint.go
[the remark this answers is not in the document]
bernard-avalabs: Thanks for the answers. I think I have a better understanding of the code now. The reason why I asked about performing writes to the DB after acquiring just a read lock is that it is an uncommon pattern. It did look safe, but I just wanted to make sure I wasn't missing anything. 

I didn't realize that database.LatestProcessedBlockKey is a const. I assumed with the word "Latest" in the name that the value would change over time. Perhaps we might want to consider changing its name.

From Cam's answer, it seems that multiple awm-relayer processes can potentially access the same key concurrently (although I assume this is something that we want to avoid in most cases). Adding some type of predicate could be useful as a external check, but I agree it is not something that is critical to add.

relayer/checkpoint.go
```diff
@@ -135,16 +143,6 @@ func (cm *checkpointManager) listenForFinishedRelays() {
...
-	if cm.committedHeight == 0 {
-		cm.logger.Debug(
-			"Committing initial height",
-			zap.Uint64("height", height),
-			zap.String("relayerID", cm.relayerID.ID.String()),
-		)
-		cm.committedHeight = height
-		return
-	}
-
```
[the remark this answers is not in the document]
cam-schultz: For sure. I also took you up on your earlier suggestion to put `CheckpointManager` into its own package to better enforce correct usage and make this condition less likely to be hit when changes are made to the code in the future.

geoff-vball (dismissed):
🚀

geoff-vball (dismissed):

bernard-avalabs (approved):
LGTM

utils/ticker.go
```diff
@@ -0,0 +1,44 @@
...
+// when subscribing.
+type Ticker struct {
+	interval      time.Duration
+	subscriptions []chan struct{}
+	lock          *sync.Mutex
+}
+
+func NewTicker(writeIntervalSeconds uint64) *Ticker {
+	return &Ticker{
+		interval: time.Duration(writeIntervalSeconds) * time.Second,
```
michaelkaplan13, line 22:
```suggestion
func NewTicker(intervalSeconds uint64) *Ticker {
	return &Ticker{
		interval: time.Duration(intervalSeconds) * time.Second,
```

Nit: now that this in `utils` the interval isn't necessarily tied to writes to the DB

relayer/checkpoint/checkpoint.go
```diff
@@ -0,0 +1,223 @@
...
+
+	counter.processedMessages++
+	cm.logger.Debug(
+		"Received finished signal",
+		zap.Uint64("height", height),
+		zap.String("relayerID", cm.relayerID.ID.String()),
+		zap.Uint64("processedMessages", counter.processedMessages),
+		zap.Uint64("totalMessages", counter.totalMessages),
+	)
+	if counter.processedMessages == counter.totalMessages {
```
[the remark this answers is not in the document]
cam-schultz: Added a comment noting that the counter is incremented in place

relayer/checkpoint/checkpoint.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+}
+
+func (cm *CheckpointManager) listenForWriteSignal() {
+	for range cm.writeSignal {
+		cm.writeToDatabase()
+	}
+}
+
+func (cm *CheckpointManager) incrementFinishedCounter(height uint64) {
```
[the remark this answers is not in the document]
cam-schultz: Added a comment above this function, and changed the name to `incrementProcessedMessageCounter` to match the name of the variable that's incremented.

michaelkaplan13 (dismissed):

geoff-vball (approved):

relayer/checkpoint/checkpoint.go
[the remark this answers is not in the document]
bernard-avalabs: Having a "->" to dereference a pointer and access a field would be nice in this situation.

bernard-avalabs (approved):
LGTM

geoff-vball (approved):

Issue #186, opened 2024-02-15
Support query parameters and HTTP headers for API requests


cam-schultz:
**Context and scope**
The relayer makes RPC requests to a handful of Avalanche APIs via the Golang clients. Specifically, the relayer uses:
- [platformvm.Client](https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/client.go#L27)
- [info.Client](https://github.com/ava-labs/avalanchego/blob/master/api/info/client.go#L19)

These clients support HTTP headers and query parameters in the form of the optional `rpc.Option` parameters to each of their method calls.

We should add `query-parameters` and `http-headers` fields to the top-level config to support these options. This would allow us to provide a rate limit bypass token, for example.

The relayer also uses [ethclient.Client](https://github.com/ava-labs/subnet-evm/blob/master/ethclient/ethclient.go#L70), which supports query parameters in the form of a fully specified URL. HTTP headers are supported via `rpc.Client.SetHeader`. It might make sense to add a `query-parameters` field to the `source-subnets` and `destination-subnets` lists for consistency.

najeal:
Hi @cam-schultz , please assign it to me.

The values provided by the new fields need to be passed to every requests right?

cam-schultz:
Hey @najeal, thanks for offering to take a look at this. @minghinmatthewlam is going to take an initial pass at this to support query parameters for `p-chain-api-url` and `info-api-url` to reach parity with the subnet RPC and WS API fields.

Please feel free to open a PR with any further proposed changes. To close this ticket out, we'll also need to support HTTP headers provided via configuration.

> The values provided by the new fields need to be passed to every requests right?

For the [Info API and P-Chain API](https://github.com/ava-labs/awm-relayer/blob/main/main/main.go#L74-L75) calls, this is correct. The [subnet RPC API](https://github.com/ava-labs/awm-relayer/blob/main/vms/evm/destination_client.go#L51) takes a raw URL, so query parameters are directly supported. Further investigation needs to be done for supporting HTTP headers for subnet RPC calls.

najeal:
@cam-schultz thank for the answer.
This means I need to wait for @minghinmatthewlam PR that will partially resolve the issue, then I can complete the rest?

cam-schultz:
@najeal, yes that would be great. Appreciate your patience on this 🙏

PR #268, opened 2024-04-25
Pchain/Info API query params

minghinmatthewlam:
## Why this should be merged
Adding query params to using pchain and info api clients. #186 

## How this works
Adds config options to pass in query parameters to using p-chain and info api clients. Then passing these query parameters as options when making calls to the p-chain/info api clients.

## How is this documented
README


config/config.go
```diff
@@ -408,6 +433,48 @@ func (c *Config) InitializeWarpQuorums() error {
 	return nil
 }
 
+func (p *PChainAPI) Validate() error {
+	if _, err := url.ParseRequestURI(p.BaseURL); err != nil {
+		return fmt.Errorf("invalid base URL: %w", err)
+	}
+	p.options = make([]rpc.Option, 0, len(p.QueryParams))
+	for key, value := range p.QueryParams {
+		p.options = append(p.options, rpc.WithQueryParam(key, value))
```
najeal, line 442: Option creation could be moved into client implementation/usage as it is the one aware of rpc usage.

For example in `validators/canonical_validator_client.go` you create a config struct being a subset of the global one
```
type Config struct {
  QueryParams map[string]string
  HttpHeaders map[string]string
}
```
This Config object becomes an argument to the `NewCanonicalValidatorClient` factory. Then the factory is responsible to create the Options and store it the same way you did.

minghinmatthewlam: Totally can see that being an option, and thanks for the suggestion. I don't anticipate there being too many differences in the usage of query parameters throughout the relayer, so prefer to leave a unified format when parsing the config under `PChainAPI` and `InfoAPI`.

peers/app_request_network.go
```diff
@@ -59,7 +56,8 @@ func NewNetwork(
 		),
 	)
 
-	networkID, err := infoClient.GetNetworkID(context.Background())
+	infoClient := cfg.InfoAPI.GetClient()
```
cam-schultz, line 59: Calls to `info.Peers`, `info.GetNodeID`, and `info.GetNodeIP` are not using the configured `rpc.Option`s. 

We should compose `AppRequestNetwork` with `InfoAPI` rather than `info.Client` to make it more obvious that `InfoApi.GetClient` should be used in conjunction with `InfoApi.GetOptions`.

config/config.go
```diff
@@ -408,6 +433,48 @@ func (c *Config) InitializeWarpQuorums() error {
...
+}
+
+func (i *InfoAPI) Validate() error {
+	if _, err := url.ParseRequestURI(i.BaseURL); err != nil {
+		return fmt.Errorf("invalid base URL: %w", err)
+	}
+	i.options = make([]rpc.Option, 0, len(i.QueryParams))
+	for key, value := range i.QueryParams {
+		i.options = append(i.options, rpc.WithQueryParam(key, value))
+	}
```
cam-schultz, line 456: Just a stylistic nit, but I've often been burned by 1) initializing a slice with a specified _size_ and 2) appending to that slice in a loop. This results in a slice with the first half of the elements empty.

This isn't the case here, since you're initializing the slice with 0 _size_ and a specified _capacity_, but I had to double check to confirm this wasn't an instance of that bug.

I'd instead advocate for either `make` + assigning via index or `append`ing to an empty slice.

cam-schultz: Actually, I think we can simply remove `i.options = make([]rpc.Option, 0, len(i.QueryParams))`, since the empty slice will be allocated already.

config/config.go
```diff
@@ -408,6 +433,48 @@ func (c *Config) InitializeWarpQuorums() error {
 	return nil
 }
 
+func (p *PChainAPI) Validate() error {
+	if _, err := url.ParseRequestURI(p.BaseURL); err != nil {
+		return fmt.Errorf("invalid base URL: %w", err)
+	}
+	p.options = make([]rpc.Option, 0, len(p.QueryParams))
```
cam-schultz, line 440: Same comment here re: `make`

cam-schultz (commented):
Just a few comments, but overall LGTM! This will be easily extendable to support HTTP headers in the future.

config/config.go
najeal: What he did is safe as far a he uses the append func. The only difference if he removes the `make` will be the underlying array of the slice being reallocated X times depending of the input array size. But as this action is done only once in the program and I suspect the array size to be low it is finally pretty the same.
But as you said it's stylistic, I don't want to question your review, just wanted to demystify the point. 👍

minghinmatthewlam: yup what @najeal said. I'd pref initializing the slice with its capacity since we know the number of elements at initialization, and can help optimize not needing x memory allocations.

README.md
```diff
@@ -105,29 +105,45 @@ The relayer binary accepts a path to a JSON configuration file as the sole argum
 
 ### Configuration
 
-The relayer is configured via a JSON file, the path to which is passed in via the `--config-file` command line argument. Top level configuration options are also able to be set via environment variable. To get the environment variable corresponding to a key, upper case the key and change the delimiter from "-" to "_". For example, `LOG_LEVEL` sets the `"log-level"` JSON key. The following configuration options are available:
+The relayer is configured via a JSON file, the path to which is passed in via the `--config-file` command line argument. Top level configuration options are also able to be set via environment variable. To get the environment variable corresponding to a key, upper case the key and change the delimiter from "-" to "\_". For example, `LOG_LEVEL` sets the `"log-level"` JSON key. The following configuration options are available:
```
michaelkaplan13, line 108: Did you mean to add this "\"? Curious why it's needed if so.

config/config.go
```diff
@@ -122,16 +125,34 @@ type WarpQuorum struct {
...
+}
+
+// Info API configuration, contains the base URL and query parameters
+type InfoAPI struct {
+	BaseURL     string            `mapstructure:"base-url" json:"base-url"`
+	QueryParams map[string]string `mapstructure:"query-parameters" json:"query-parameters"`
+
+	options []rpc.Option
+	client  info.Client
+}
```
michaelkaplan13, line 144: I think it could it be cleaner to have a single `APIConfig` type for the configuration, and keep the actual client instances out of `config` package. It seems a little off to have the `rpc/platformvm/info` packages all be dependencies of `config`.

minghinmatthewlam: I changed it to `APIConfig` for both P-Chain and Info APIs, then moved the logic of initializing client and options to where it's used in `peers`. Also moved validator client logic to `peers` since they're closely tied.

@najeal Also related to your previous suggestion https://github.com/ava-labs/awm-relayer/pull/268#discussion_r1579897585, implemented a similar change

README.md
minghinmatthewlam: nope thanks for finding

peers/api_clients.go
```diff
@@ -0,0 +1,47 @@
...
+	client  info.Client
+	options []rpc.Option
+}
+
+func NewPChainAPI(apiConfig *config.APIConfig) (*PChainAPI, error) {
+	client := platformvm.NewClient(apiConfig.BaseURL)
+	options := make([]rpc.Option, 0, len(apiConfig.QueryParams))
+	for key, value := range apiConfig.QueryParams {
+		options = append(options, rpc.WithQueryParam(key, value))
+	}
```
michaelkaplan13, line 30: Nit: Lets move this duplicated piece to a `initializeOptions` (or another name) helper

michaelkaplan13 (commented):
One nit, but otherwise looks good

najeal (approved):

config/config.go
cam-schultz: Apologies if my original comment was misleading. For non-performance critical initialization code, code readability should take priority over small performance optimizations. In any case, I see this section has been removed.

peers/api_clients.go
cam-schultz, line 1: In another comment, I suggested moving the logic in `NewPChainAPI` directly into `NewCanonicalValidatorClient`. This file would then only contain `InfoAPI`. I'd suggest renaming this file to `info_client.go`, and implementing the client as a wrapper around `info.Client` that encapsulated the RPC options. The result would be that we could call the info API like

`infoAPI.GetNetworkID(context.Background())`

instead of 

`infoAPI.client.GetNetworkID(context.Background(), infoAPI.options...)`

peers/api_clients.go
minghinmatthewlam: there was some refactoring but I added a `utils` for `InitializeOptions`

peers/app_request_network.go
```diff
@@ -100,12 +105,21 @@ func NewNetwork(
...
-	validatorClient := validators.NewCanonicalValidatorClient(logger, pChainClient)
+	pChainAPI, err := NewPChainAPI(cfg.PChainAPI)
+	if err != nil {
+		logger.Error(
+			"Failed to create P-Chain API",
+			zap.Error(err),
+		)
+		return nil, nil, err
+	}
+	validatorClient := validators.NewCanonicalValidatorClient(logger, pChainAPI.client, pChainAPI.options)
```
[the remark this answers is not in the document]
minghinmatthewlam: Updated canonical validator client to hold p chain client interaction logic. Also wrapped additional functions we use with the p-chain to the canonical validator client.

peers/api_clients.go
minghinmatthewlam: added info client wrapping for options.

peers/app_request_network.go
```diff
@@ -91,6 +78,23 @@ func NewNetwork(
...
 
+	infoAPI, err := NewInfoAPI(cfg.InfoAPI)
+	if err != nil {
+		logger.Error(
+			"Failed to create info API",
+			zap.Error(err),
+		)
+		return nil, nil, err
+	}
+	networkID, err := infoAPI.client.GetNetworkID(context.Background(), infoAPI.options...)
```
cam-schultz, line 89:
```suggestion
	networkID, err := infoAPI.GetNetworkID(context.Background())
```

peers/app_request_network.go
```diff
@@ -152,7 +156,7 @@ func (n *AppRequestNetwork) ConnectPeers(nodeIDs set.Set[ids.NodeID]) set.Set[id
 	// re-adding connections to already tracked peers.
 
 	// Get the list of peers
-	peers, err := n.infoClient.Peers(context.Background())
+	peers, err := n.infoAPI.client.Peers(context.Background(), n.infoAPI.options...)
```
cam-schultz, line 159:
```suggestion
	peers, err := n.infoAPI.Peers(context.Background())
```

peers/app_request_network.go
```diff
@@ -184,13 +188,13 @@ func (n *AppRequestNetwork) ConnectPeers(nodeIDs set.Set[ids.NodeID]) set.Set[id
 
 	// If the Info API node is in nodeIDs, it will not be reflected in the call to info.Peers.
 	// In this case, we need to manually track the API node.
-	if apiNodeID, _, err := n.infoClient.GetNodeID(context.Background()); err != nil {
+	if apiNodeID, _, err := n.infoAPI.client.GetNodeID(context.Background(), n.infoAPI.options...); err != nil {
```
cam-schultz, line 191:
```suggestion
	if apiNodeID, _, err := n.infoAPI.GetNodeID(context.Background()); err != nil {
```

peers/app_request_network.go
```diff
@@ -184,13 +188,13 @@ func (n *AppRequestNetwork) ConnectPeers(nodeIDs set.Set[ids.NodeID]) set.Set[id
...
 		n.logger.Error(
 			"Failed to get API Node ID",
 			zap.Error(err),
 		)
 	} else if nodeIDs.Contains(apiNodeID) {
-		if apiNodeIP, err := n.infoClient.GetNodeIP(context.Background()); err != nil {
+		if apiNodeIP, err := n.infoAPI.client.GetNodeIP(context.Background(), n.infoAPI.options...); err != nil {
```
cam-schultz, line 197:
```suggestion
		if apiNodeIP, err := n.infoAPI.GetNodeIP(context.Background()); err != nil {
```

cam-schultz (commented):
Just need to update the `InfoAPI` callsites, then LGTM

cam-schultz (approved):
LGTM!

peers/validators/canonical_validator_client.go
```diff
@@ -59,15 +66,27 @@ func (v *CanonicalValidatorClient) GetCurrentCanonicalValidatorSet(subnetID ids.
...
+func (v *CanonicalValidatorClient) GetCurrentValidators(ctx context.Context, subnetID ids.ID, nodeIDs []ids.NodeID) ([]platformvm.ClientPermissionlessValidator, error) {
+	return v.client.GetCurrentValidators(ctx, subnetID, nodeIDs, v.options...)
+}
+
+func (v *CanonicalValidatorClient) GetValidatorsAt(
+	ctx context.Context,
+	subnetID ids.ID,
+	height uint64,
+) (map[ids.NodeID]*validators.GetValidatorOutput, error) {
+	return v.client.GetValidatorsAt(ctx, subnetID, height, v.options...)
```
michaelkaplan13, line 89: Is there a reason why it's necessary to export these methods? My worry is there are:
- `GetCurrentCanonicalValidatorSet`
- `GetValidatorsAt`
- `GetCurrentValidators`
- `getCurrentValidatorSet`
- `GetValidatorSet`

Which all do slightly different things but make it confusing to know which should be used. If only called internally here, I think it'd prefer to directly call `v.client.GetValidatorsAt` and `v.client.GetCurrentValidators` in the existing implementations rather than add additional similarly named wrapped.

cam-schultz: `CanonicalValidatorClient` needs to implement `warp.ValidatorState` in order to be able to call `GetCanonicalValidatorSet`. It looks like that interface has been updated in AvalancheGo since this was last updated here, so `GetValidatorSet` should be the only method we need to export.

michaelkaplan13 (approved):
LGTM!

cam-schultz (approved):

PR #273, opened 2024-05-01
EIP-2 cap s-value

cam-schultz:
## Why this should be merged
Fixes a bug in the KMS signature recovery logic where we were not capping the signature's s-value according to [EIP-2](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2.md#specification).

## How this works
Ensures that the signature's s-value is not greater than `secp256k1n/2`. See reference implementation here: https://github.com/welthee/go-ethereum-aws-kms-tx-signer/blob/main/signer.go#L79

## How this was tested
Manually verified previously invalid signatures are now valid with the s-value cap.

## How is this documented
N/A

### DIFF
--- a/vms/evm/signer/kms_signer.go
+++ b/vms/evm/signer/kms_signer.go
@@ -109,11 +109,11 @@ func (s *KMSSigner) SignTx(tx *types.Transaction, evmChainID *big.Int) (*types.T
 	if err != nil {
 		return nil, err
 	}
-
-	sigBytes, err := s.getEthereumSignature(h, sigAsn1.R.Bytes, sigAsn1.S.Bytes)
+	sigBytes, err := s.recoverEIP155Signature(h, sigAsn1.R.Bytes, sigAsn1.S.Bytes)
 	if err != nil {
 		return nil, err
 	}
+
 	return tx.WithSignature(signer, sigBytes)
 }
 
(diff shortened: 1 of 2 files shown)


michaelkaplan13 (approved):
Would be good to add unit tests for `recoverEIP155Signature`, but the changes LGTM

geoff-vball (dismissed):
👍

vms/evm/signer/kms_signer.go
```diff
@@ -109,22 +109,31 @@ func (s *KMSSigner) SignTx(tx *types.Transaction, evmChainID *big.Int) (*types.T
...
 
-// Recover the EIP-155 signature from the KMS signature
+// Recover the EIP-155 signature from the KMS signature.
 // KMS returns the signature in ASN.1 format, but the EIP-155 signature is in R || S || V format,
-// so we need to test both V = 0 and V = 1 against the recovered public key
-func (s *KMSSigner) getEthereumSignature(txHash []byte, rBytes []byte, sBytes []byte) ([]byte, error) {
+// so we need to test both V = 0 and V = 1 against the recovered public key.
+// Additionally, with EIP-2 S-values are capped at secp256k1n/2, so adjust that if necessary.
+func (s *KMSSigner) recoverEIP155Signature(txHash []byte, rBytes []byte, sBytes []byte) ([]byte, error) {
+	sBigInt := new(big.Int).SetBytes(sBytes)
```
minghinmatthewlam, line 129: nit: what was our decision with using new(big.Int) vs big.NewInt()?

vms/evm/signer/kms_signer.go
```diff
@@ -109,22 +109,31 @@ func (s *KMSSigner) SignTx(tx *types.Transaction, evmChainID *big.Int) (*types.T
...
 	return s.eoa
 }
 
-// Recover the EIP-155 signature from the KMS signature
+// Recover the EIP-155 signature from the KMS signature.
 // KMS returns the signature in ASN.1 format, but the EIP-155 signature is in R || S || V format,
-// so we need to test both V = 0 and V = 1 against the recovered public key
-func (s *KMSSigner) getEthereumSignature(txHash []byte, rBytes []byte, sBytes []byte) ([]byte, error) {
+// so we need to test both V = 0 and V = 1 against the recovered public key.
+// Additionally, with EIP-2 S-values are capped at secp256k1n/2, so adjust that if necessary.
```
minghinmatthewlam, line 127:
```suggestion
// Additionally, EIP-2 S-values are capped at secp256k1n/2, so adjust that if necessary.
```

vms/evm/signer/kms_signer.go
cam-schultz: My preference is to use the constructor. Changed now.

cam-schultz:
> Would be good to add unit tests for `recoverEIP155Signature`, but the changes LGTM

Added unit tests

geoff-vball (approved):

PR #278, opened 2024-05-04
Rpc options

najeal:
## Why this should be merged

Adding http-headers for p-chain and info-api clients.
Related to  https://github.com/ava-labs/awm-relayer/issues/186

## How this works

Use mechanism introduced in #268 to pass http-headers to p-chain & info-api clients.

## How is this documented

Details of the new Config fields are added in the README.

PR #279, opened 2024-05-06
Eth Client: Use http-headers and query-parameters

### LINKED ISSUE #186: Support query parameters and HTTP headers for API requests

najeal:
## Why this should be merged

It is proposing a way to use http-headers & query-params with eth client.
Closes #186 - Should be merged after #278 

## How this works

Instead of using eth client Factory from `subnet-evm`, it creates a local Factory expecting query-params & http-headers as arguments.
The factory package is named `ethclient_utils` instead of `ethclient` because both packages will be imported in go files.

## How this was tested

Unit tests have been added for the `ethclient_utils`.

## How is this documented

- [x]  Update the README


config/config.go
```diff
@@ -86,6 +87,8 @@ type SourceBlockchain struct {
 	BlockchainID                      string                           `mapstructure:"blockchain-id" json:"blockchain-id"`
 	VM                                string                           `mapstructure:"vm" json:"vm"`
 	RPCEndpoint                       string                           `mapstructure:"rpc-endpoint" json:"rpc-endpoint"`
```
cam-schultz, line 89: Let's use `APIConfig` to specify the URL, headers, and query parameters for `SourceBlockchain.RPCEndpoint`, `SourceBlockchain.WSEndpoint`, and `DestinationBlockchain.RPCEndpoint`. We currently use the same query params and headers for both WS and RPC connections, but the user may want these to be different.

ethclient_utils/client.go
```diff
@@ -0,0 +1,51 @@
...
+	"fmt"
+	"net/url"
+
+	"github.com/ava-labs/subnet-evm/ethclient"
+	"github.com/ava-labs/subnet-evm/rpc"
+)
+
+var ErrInvalidEndpoint = errors.New("invalid rpc endpoint")
+
+func DialWithConfig(ctx context.Context, endpoint string, httpHeaders, queryParams map[string]string) (ethclient.Client, error) {
```
cam-schultz, line 18: We can just pass in a `config.APIConfig` object here. That would be better aligned with the naming of the function, as well as encapsulate everything we need in a single type derived directly from the `Config` object.

ethclient_utils/client.go
```diff
@@ -0,0 +1,51 @@
...
+	values := uri.Query()
+	for key, value := range queryParams {
+		values.Add(key, value)
+	}
+	uri.RawQuery = values.Encode()
+	return uri.String(), nil
+}
+
+// newClientOptions creates a ClientOption slice from httpHeaders
+func newClientOptions(httpHeaders map[string]string) []rpc.ClientOption {
```
cam-schultz, line 45: Let's name this function `newClientHeaderOptions` or similar, since the returned `rpc.ClientOption` only contains headers.

vms/evm/subscriber.go
```diff
@@ -49,6 +50,8 @@ var warpFilterQuery = interfaces.FilterQuery{
 type subscriber struct {
 	nodeWSURL    string
 	nodeRPCURL   string
+	httpHeaders  map[string]string
+	queryParams  map[string]string
```
cam-schultz, line 54: We can maintain two `APIConfig` instances in this struct, one for the WS connection and one for the RPC connection.

ethclient_utils/client_test.go
```diff
@@ -0,0 +1,45 @@
+// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
```
cam-schultz, line 1:
```suggestion
// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
```

ethclient_utils/client.go
```diff
@@ -0,0 +1,51 @@
+// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
```
cam-schultz, line 1:
```suggestion
// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
```

ethclient_utils/client.go
cam-schultz, line 18:
```suggestion
// DialWithContext returns an ethclient.Client with the internal RPC client configured with the provided options.
func DialWithConfig(ctx context.Context, endpoint string, httpHeaders, queryParams map[string]string) (ethclient.Client, error) {
```

ethclient_utils/client.go
```diff
@@ -0,0 +1,51 @@
...
+// See the file LICENSE for licensing terms.
+
+package ethclient_utils
```
cam-schultz, line 4: First, Go convention is to name packages in all lowercase with no punctuation, so `ethclientutils`.
Second, I'm not a huge fan of `ethclientutils`, since this package wraps `ethclient.Client`, rather than providing utilities on `ethclient.Client` instances. Honestly I think `ethclient` on its own is differentiated enough, since it is how other `awm-relayer` packages will create (`subnet-evm`) `ethclient.Clients`.

cam-schultz (commented):
Overall this approach makes sense to me. Let's make use of the new `APIConfig` type introduced in https://github.com/ava-labs/awm-relayer/pull/268. I've left comments where we should make that change, as well as a handful of other minor comments. Thanks for putting this together!

ethclient_utils/client.go
najeal: I will make the change 👍 
In some places we still need to import `subnet-evm/ethclient` for type assertion. So I hope there will ne be confusion caused by package name collision between `subnet-evm` and `awm-relayer`.

ethclient_utils/client.go
najeal: I created a local struct. I cannot use the struct from the config package otherwise there is import cycle error.

ethclient_utils/client.go
cam-schultz: We can remove the dependency on `subnet-evm/ethclient` everywhere except for `ethclient/client.go` by wrapping the the `subnet-evm` client like so:
```
package ethclient
import "github.com/ava-labs/subnet-evm/ethclient"

type Client interface {
    ethclient.Client
}
```

Then everywhere in the rest of the repo that uses `subnet-evm/ethclient` can be replaced with `awm-relayer/ethclient`.

cam-schultz (commented):
Once the change I mentioned in https://github.com/ava-labs/awm-relayer/pull/279#discussion_r1596865910 is implemented, I think this is good to go.

ethclient_utils/client.go
najeal: I agree 👍

ethclient_utils/client.go
najeal: Are you sure? It looks cleaner with your initial proposition.
Go build takes care about the two structs consistency. It fails if the fields are not aligned.

cam-schultz: Yeah, I'm sure. Besides duplicating type definitions to satisfy the dependency graph, by specifying the struct in the parameter list (and requiring the caller to construct an instance of that struct) it is not immediately clear from the callsite that the struct does not have any side effects or additional functionality. By passing the parameters directly, it is clear that there are no such side effects, even though the callsite requires more parameters.

The best solution would be to incorporate `DialWithConfig` into `config.APIConfig` directly, so that we can use just that one type throughout the application. This refactor is outside the scope of this PR and issue though, so my preference is to defer that to a later change.

najeal:
@cam-schultz Created a local Client interface in `ethclient` package and the factory expects independent parameters now, instead of a struct.

relayer/application_relayer.go
```diff
@@ -191,7 +191,7 @@ func (r *applicationRelayer) relayMessage(
 func (r *applicationRelayer) createSignedMessage(unsignedMessage *avalancheWarp.UnsignedMessage) (*avalancheWarp.Message, error) {
 	r.logger.Info("Fetching aggregate signature from the source chain validators via API")
 	// TODO: To properly support this, we should provide a dedicated Warp API endpoint in the config
-	uri := utils.StripFromString(r.sourceBlockchain.RPCEndpoint, "/ext")
+	uri := utils.StripFromString(r.sourceBlockchain.RPCEndpoint.BaseURL, "/ext")
```
michaelkaplan13, line 194: Does this remove the possibility to include query params to be used by the `warpBackend` client create below? I'm not 100% sure if that was previously supported or not, or if that's something we should be aware of going forward. cc @cam-schultz

cam-schultz: The Warp client expects a base URI, from which it constructs the full path like so: `fmt.Sprintf("%s/ext/bc/%s/rpc", uri, chain)fmt.Sprintf("%s/ext/bc/%s/rpc", uri, chain)`.  Query params were never supported for the Warp client, so this change makes no difference there. We'd have to add a similar wrapper to support query params/https headers, but since the Warp endpoint is not currently accessible via configurable, it's not a high priority.

michaelkaplan13 (approved):

relayer/application_relayer.go
michaelkaplan13: Thanks for the context 🙏

PR #287, opened 2024-05-13
Version flag

### LINKED ISSUE #281: Add --version flag
**Context and scope**
I have a awm-relayer binary, but I have no way to see which version it uses and therefore what format of the config it expects.

Add a flag to query the version of the relayer

```bash
awm-relayer --version

awm-relayer -v
```

**Open questions**
None

najeal:
## Why this should be merged

It closes #281 

## How this works

This adds a new  `--version` flag. The binary print the version then exits when the flag is used.
The go value is set during goreleaser processing.

## How this was tested

It has been tested manually, tagging locally and run `goreleaser build` command.

## How is this documented


config/flags.go
```diff
@@ -8,5 +8,6 @@ import "github.com/spf13/pflag"
 func BuildFlagSet() *pflag.FlagSet {
 	fs := pflag.NewFlagSet("awm-relayer", pflag.ContinueOnError)
 	fs.String(ConfigFileKey, "", "Specifies the relayer config file")
+	fs.BoolP(VersionKey, "v", false, "Display awm-relayer version")
```
cam-schultz, line 11:
```suggestion
	fs.BoolP(VersionKey, "", false, "Display awm-relayer version")
```

Let's only allow the long form here to prevent confusion over `-v` which is commonly used for `--verbose`.

.goreleaser.yml
```diff
@@ -5,6 +5,8 @@ builds:
     binary: awm-relayer
     flags:
       - -v
+    ldflags:
+      - -X main.version={{.Version}}
```
cam-schultz, line 9:
```suggestion
      - -X main.version={{ .Version }}
```
For consistency

main/main.go
```diff
@@ -25,13 +25,20 @@ import (
...
 	"go.uber.org/atomic"
 	"go.uber.org/zap"
 	"golang.org/x/sync/errgroup"
 )
 
+var version = "v0.0.0-dev"
+
 func main() {
 	fs := config.BuildFlagSet()
+	if err := fs.Parse(os.Args[1:]); err != nil {
```
cam-schultz, line 38: Rather than calling `os.Args` twice, let's define `args := os.Args[1:]` as the first line of this function and reuse it.

main/main.go
```diff
@@ -309,3 +316,18 @@ func initializeMetrics() (prometheus.Gatherer, prometheus.Registerer, error) {
 	}
 	return gatherer, registry, nil
 }
+
+// handleInfoDisplayFlags is in charge of displaying cli informations
+// when an info display flag is enabled
+// It exits if one of the flags has been enabled by the user.
+// examples: version, authors, help ...
+func handleInfoDisplayFlags(fs *pflag.FlagSet) {
```
cam-schultz, line 324: Let's move this logic inline in `main`. I think it's more readable to have all of the initialization logic that could cause the program to exit in `main`.  The word "handle" in this function name also makes me think something is happening async, which it is not.

main/main.go
```diff
@@ -25,13 +25,20 @@ import (
...
 )
 
+var version = "v0.0.0-dev"
+
 func main() {
 	fs := config.BuildFlagSet()
+	if err := fs.Parse(os.Args[1:]); err != nil {
+		panic(fmt.Errorf("couldn't parse flags: %w", err))
+	}
+	handleInfoDisplayFlags(fs)
```
cam-schultz, line 41: Can you add a comment explaining that we check for the version key, and if it's found we display it and exit?

cam-schultz (commented):
The general approach makes sense to me. I left some suggestions mainly focused on readability.

We should also include the version ldflag in `build.sh`. Since `build.sh` isn't only run on release, we can use the most recent tag on the current branch like so: `git describe --tags --abbrev=0`. We already do that when [printing the version](https://github.com/ava-labs/awm-relayer/blob/main/scripts/build.sh#L52) that's being built, so adding that git command already be compatible with CI.

main/main.go
najeal: Nice catch ! I figured out `config.BuildViper` does not need anymore the os.Args, so it will only be used once after my changes

najeal:
> The general approach makes sense to me. I left some suggestions mainly focused on readability.
> 
> We should also include the version ldflag in `build.sh`. Since `build.sh` isn't only run on release, we can use the most recent tag on the current branch like so: `git describe --tags --abbrev=0`. We already do that when [printing the version](https://github.com/ava-labs/awm-relayer/blob/main/scripts/build.sh#L52) that's being built, so adding that git command already be compatible with CI.

I need one more information about what you expect.
Currently I use `{{ .Version }}` as a value to override the `version` var in the code. It is the Git Tag with the "v" prefix stripped.
What you ask to add in the `build.sh` is adding the tag itself (containing the "v")
For consistency, either I strip the "v" in the build.sh or I use `{{ .Tag }}` to keep the "v" in the `version` variable.

cam-schultz:
> I need one more information about what you expect. Currently I use `{{ .Version }}` as a value to override the `version` var in the code. It is the Git Tag with the "v" prefix stripped. What you ask to add in the `build.sh` is adding the tag itself (containing the "v") For consistency, either I strip the "v" in the build.sh or I use `{{ .Tag }}` to keep the "v" in the `version` variable.

Thanks for the call-out. Let's stick with `{{.Version}}` in `.goreleaser.yml`. Stripping the leading `v` seems to be the accepted convention for version flags.

As for `build.sh`, I think it's fine to just output the last known git tag, rather than stripping the leading `v`. My reasoning is 1) the git tag doesn't necessarily need to be semver (as may be the case if somebody is developing in a fork), and 2) the value populated by `build.sh` will only be seen by developers. I only care that the release has the proper version format.

scripts/build.sh
```diff
@@ -49,5 +49,6 @@ else
 fi
 
 # Build AWM Relayer, which is run as a standalone process
-echo "Building AWM Relayer Version: $(git describe --tags --abbrev=0) at $binary_path"
-go build -o "$binary_path" "main/"*.go
+last_git_tag=$(git describe --tags --abbrev=0)
```
cam-schultz, line 52: This will fail if no tag is found. I recommend setting it to the same default value hardcoded in the binary in that case.
```suggestion
last_git_tag=$(git describe --tags --abbrev=0 2>/dev/null) || last_git_tag="v0.0.0-dev"
```

main/main.go
```diff
@@ -39,8 +38,16 @@ func main() {
 	if err := fs.Parse(os.Args[1:]); err != nil {
 		panic(fmt.Errorf("couldn't parse flags: %w", err))
 	}
-	handleInfoDisplayFlags(fs)
-	v, err := config.BuildViper(fs, os.Args[1:])
+	// Having the version flag set, we display the version then exit
```
cam-schultz, line 41:
```suggestion
	// If the version flag set, display the version then exit
```
Prefer avoiding "we". Also think this phrasing is less awkward.

cam-schultz (commented):
One small issue with the build script, and a small nit. Otherwise LGTM!

main/main.go
```diff
@@ -31,9 +31,23 @@ import (
...
 
+var version = "v0.0.0-dev"
+
 func main() {
 	fs := config.BuildFlagSet()
-	v, err := config.BuildViper(fs, os.Args[1:])
+	if err := fs.Parse(os.Args[1:]); err != nil {
+		panic(fmt.Errorf("couldn't parse flags: %w", err))
+	}
+	// If the version flag set, display the version then exit
```
cam-schultz, line 41:
```suggestion
	// If the version flag is set, display the version then exit
```
My bad on the typo.

cam-schultz (commented):
One last tiny nit that was my fault for introducing

cam-schultz (dismissed):

cam-schultz:
I didn't realize this before, but goreleaser automatically sets `main.version` to the version with `v` stripped. We should be able to remove the `ldflags` from `.goreleaser.yml` altogether. https://goreleaser.com/cookbooks/using-main.version/

najeal:
@cam-schultz You are right 👍

cam-schultz (approved):

geoff-vball (approved):
Tested on a local build and looks good, thanks!

Issue #185, opened 2024-02-15
Characterize relayer throughput


cam-schultz:
**Context and scope**
The relayer processes Warp messages from separate chains concurrently, but within a single chain messages are processed in serial (see https://github.com/ava-labs/awm-relayer/issues/31). There is therefore a throughput limit per chain that we should mesaure and characterize.

**Discussion and alternatives**
- Throughput is very likely network bound, as the application side processing complexity is insignificant compared to the number of network round trips required to relay a message. We should aim to answer the following questions:
  - Is single message relaying in fact network bound?
  - What's the breakdown between network latency and application side processing latency for a single message?
  - What's the minimum and maximum number of sequential network round trips needed to relay a single message?

**Open questions**
At what load level do concurrent database writes become a bottleneck?

PR #288, opened 2024-05-13
Process messages concurrently

### LINKED ISSUE #31: Process messages from the same chain concurrently

cam-schultz:
## Why this should be merged
Fixes #31 

## How this was tested
CI
TODO: before/after througput testing (see #185)

## How is this documented


relayer/application_relayer.go
```diff
@@ -102,30 +106,77 @@ func newApplicationRelayer(
...
 
-func (r *applicationRelayer) relayMessage(
-	unsignedMessage *avalancheWarp.UnsignedMessage,
+// Process [msgs] at height [height] by relaying each message to the destination chain.
+// Checkpoints the height with the checkpoint manager when all messages are relayed.
+func (r *ApplicationRelayer) ProcessHeight(height uint64, msgs []messages.MessageHandler) error {
+	var eg errgroup.Group
+	for _, msg := range msgs {
+		eg.Go(func() error {
+			return r.ProcessMessage(msg)
```
cam-schultz, line 132: Upgraded to go 1.22 so that this closure captures the value of `msg` on each iteration, rather than the variable itself.

.golangci.yml
```diff
@@ -3,13 +3,13 @@
 run:
   timeout: 3m
   tests: true
-  # skip auto-generated files.
-  skip-files:
```
cam-schultz, line 7: The latest golangci-lint deprecates this option in favor of `exclude-files`

relayer/application_relayer.go
bernard-avalabs: That's an interesting change for a language upgrade. I'm surprised if this change doesn't break some apps.

relayer/application_relayer.go
```diff
@@ -102,30 +106,84 @@ func newApplicationRelayer(
...
+
+// Relays a message to the destination chain. Does not checkpoint the height.
+func (r *ApplicationRelayer) ProcessMessage(msg messages.MessageHandler) error {
+	// Increment the request ID. Make sure we don't hold the lock while we relay the message.
+	r.lock.Lock()
+	r.currentRequestID++
+	r.lock.Unlock()
+
+	err := r.relayMessage(
+		r.currentRequestID,
```
bernard-avalabs, line 163: Can more than one goroutine call ProcessMessage (on different messages) at the same time? If yes, then this could result in more than one relayMessage calls to use the same request ID. 

If the request ID needs to be distinct on each call, then the currentRequestID should be copied to a local variable inside the critical section above in between the lock/unlock. 

I also assume that the ordering between calls to relayMessage calls doesn't matter (or is provided in some other parts of the system).

bernard-avalabs (commented):
I left a few comments.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/ava-labs/awm-relayer
 
-go 1.21.10
+go 1.22.3
```
michaelkaplan13, line 3: Lets keep this at 1.21.10. We always aim to use the latest patch version of the minor version one behind latest.

types/types.go
```diff
@@ -53,24 +56,50 @@ func NewWarpBlockInfo(header *types.Header, ethClient ethclient.Client) (*WarpBl
...
 // Extract the Warp message information from the raw log
-func NewWarpLogInfo(log types.Log) (*WarpLogInfo, error) {
+func NewWarpMessageInfo(log types.Log) (*WarpMessageInfo, error) {
 	if len(log.Topics) != 3 {
 		return nil, ErrInvalidLog
 	}
 	if log.Topics[0] != WarpPrecompileLogFilter {
 		return nil, ErrInvalidLog
 	}
+	unsignedMsgBytes, err := UnpackWarpMessage(log.Data)
```
michaelkaplan13, line 82: Nit: 
```suggestion
	unsignedMsg, err := UnpackWarpMessage(log.Data)
```

types/types.go
```diff
@@ -53,24 +56,50 @@ func NewWarpBlockInfo(header *types.Header, ethClient ethclient.Client) (*WarpBl
...
-		UnsignedMsgBytes: log.Data,
+	return &WarpMessageInfo{
+		SourceAddress:   common.BytesToAddress(log.Topics[1][:]),
+		UnsignedMessage: unsignedMsgBytes,
 	}, nil
 }
+
+func UnpackWarpMessage(unsignedMsgBytes []byte) (*avalancheWarp.UnsignedMessage, error) {
+	unsignedMsg, err := warp.UnpackSendWarpEventDataToMessage(unsignedMsgBytes)
+	if err != nil {
```
michaelkaplan13, line 95: Could you clarify when we expect `warp.UnpackSendWarpEventDataToMessage` to be successful and when we'd expect this fallback to be necessary?

It looks like `UnpackSendWarpEventDataToMessage` uses the `ParseUnsignedMessage` under the hood, but it passes `event.Message` to it, so I'm not sure which is the expected case.

May want to consider adding a debug log for the fallback case here also.

types/types.go
```diff
@@ -20,19 +22,20 @@ var ErrInvalidLog = errors.New("invalid warp message log")
...
-	SourceAddress    common.Address
-	UnsignedMsgBytes []byte
+// WarpMessageInfo describes the transaction information for the Warp message
+// sent on the source chain.
+// WarpMessageInfo instances are either derived from the logs of a block or
+// from the manual Warp message information provided via configuration.
+type WarpMessageInfo struct {
+	SourceAddress   common.Address
+	UnsignedMessage *avalancheWarp.UnsignedMessage
+	MessageManager  messages.MessageManager
```
michaelkaplan13, line 38: Looks like `WarpMessageInfo.MessageManager` can be removed.

relayer/application_relayer.go
geoff-vball: @cam-schultz There's a flag you can set in `1.21` that enables this feature, so we can keep our versions consistent.

geoff-vball: https://tip.golang.org/wiki/LoopvarExperiment

peers/external_handler.go
```diff
@@ -147,23 +109,99 @@ func (h *RelayerExternalHandler) Disconnected(nodeID ids.NodeID) {
...
+	// Check for the expected number of responses, and clear from the map if all expected responses have been received
+	responses := h.responsesCount[reqID.RequestID]
+	received := responses.received + 1
+	if received == responses.expected {
+		close(h.responseChans[reqID.RequestID])
+		delete(h.responseChans, reqID.RequestID)
+		delete(h.responsesCount, reqID.RequestID)
+	} else {
+		h.responsesCount[reqID.RequestID] = expectedResponses{
+			expected: h.responsesCount[reqID.RequestID].expected,
```
michaelkaplan13, line 203:
```suggestion
			expected: responses.expected,
```

peers/external_handler.go
```diff
@@ -147,23 +109,99 @@ func (h *RelayerExternalHandler) Disconnected(nodeID ids.NodeID) {
...
+	}
+
+	// Register the response with the timeout manager
 	h.timeoutManager.Remove(reqID)
+
+	// Dispatch to the appropriate response channel
+	if responseChan, ok := h.responseChans[requestID]; ok {
+		responseChan <- inboundMessage
+	} else {
+		h.log.Debug("Could not find response channel for request", zap.Uint32("requestID", requestID))
```
michaelkaplan13, line 191: My understanding is that this should never be the case unless there is an implementation error. If that's correct, should we raise this up to an `Error` or possibly even `Fatal`?

peers/external_handler.go
```diff
@@ -147,23 +109,99 @@ func (h *RelayerExternalHandler) Disconnected(nodeID ids.NodeID) {
...
+		avalancheCommon.ErrTimeout.Message,
 	)
 	h.timeoutManager.Put(reqID, false, func() {
 		h.HandleInbound(context.Background(), inMsg)
 	})
 }
 
 // RegisterResponse registers an AppResponse with the timeout manager
-func (h *RelayerExternalHandler) RegisterResponse(reqID ids.RequestID) {
+func (h *RelayerExternalHandler) RegisterAppResponse(inboundMessage message.InboundMessage) {
```
michaelkaplan13, line 146: Looks like this function should be un-exported

relayer/checkpoint/checkpoint.go
michaelkaplan13, line 1: Really like the simplification to the `CheckpointManager` here 🙌

relayer/listener.go
```diff
@@ -259,6 +213,8 @@ func (lstnr *Listener) ProcessLogs(ctx context.Context) error {
 				return fmt.Errorf("failed to catch up on historical blocks")
 			}
 		case blockHeader := <-lstnr.Subscriber.Headers():
+			// Parse the logs in the block, and group by application relayer
```
michaelkaplan13, line 216: IIUC, we still make a blocking RPC call per block here to fetch all of the warp logs from it. As long as the latency of that request is significantly less that the block interval I don't think it will be a bottleneck, but would it be difficult to put it in its own goroutine either way? Something like a `go lstnr.processBlockHeader(blockHeader)` that encapsulates the logic here possibly.

relayer/listener.go
```diff
@@ -464,47 +363,40 @@ func (lstnr *Listener) getApplicationRelayer(
...
-// From the raw log information
-type parsedMessageInfo struct {
-	unsignedMessage    *avalancheWarp.UnsignedMessage
-	messageManager     messages.MessageManager
-	applicationRelayer *applicationRelayer
-}
-
-func (lstnr *Listener) parseMessage(warpLogInfo *relayerTypes.WarpLogInfo) (
-	*parsedMessageInfo,
+// TODONOW: Does this function make sense? There's probably a better way to organize this logic.
```
michaelkaplan13, line 366: Aside from the TODONOW, could you add a comment here describing what `GetAppRelayerMessageHandler` is meant to be used for? 🙏

main/main.go
```diff
@@ -229,69 +239,148 @@ func main() {
...
+	if err != nil {
+		logger.Error(
+			"Failed to get current block height",
+			zap.Error(err),
+		)
+		return err
+	}
+
+	// Each ApplicationRelayer determines its starting height based on the database state.
+	// The Listener back-processes from the minimum height across all of the ApplicationRelayers.
```
michaelkaplan13, line 290: Nit: I'm not sure "back-processes" is an accurate description here. Maybe just "begins processing messages starting from the minimum height across all of the application relayers"?

main/main.go
```diff
@@ -229,69 +239,148 @@ func main() {
...
 	network *peers.AppRequestNetwork,
-	responseChan chan message.InboundMessage,
 	destinationClients map[ids.ID]vms.DestinationClient,
 	messageCreator message.Creator,
 	relayerHealth *atomic.Bool,
-	manualWarpMessages []*relayerTypes.WarpLogInfo,
+	manualWarpMessages []*relayerTypes.WarpMessageInfo,
 	cfg *config.Config,
 ) error {
+	// Create the ApplicationRelayers
```
michaelkaplan13, line 258: It may be cleaner to have a `createApplicationRelayers` helper here and separate that out from `runListener`, which is fairly long and complex as is.

We could create a helper for `processManualWarpMessages` and split that logic out from `runListener` as well.

messages/message_manager.go
```diff
@@ -6,63 +6,37 @@
...
 
 // MessageManager is specific to each message protocol. The interface handles choosing which messages to send
 // for each message protocol, and performs the sending to the destination chain.
 type MessageManager interface {
+	// Create a message handler to relay the Warp message
+	NewMessageHandler(unsignedMessage *warp.UnsignedMessage) (MessageHandler, error)
+}
+
+// MessageHandlers relay a single Warp message. A new instance should be created for each Warp message.
+type MessageHandler interface {
```
michaelkaplan13, line 22: I'm a bit confused what the need is to separate `MessageManager` and `MessageHandler` here, and worried it could easily be confused what each is intended to be used for in the future. 

It seems like a `MessageManager` is just a wrapper for creating `MessageHandler`s. Would it be possible to remove it all together and just use `NewMessageHandler` directly with additional params as needed? Otherwise, I think we should consider renaming `MessageManager` to `MessageHandlerFactory` or something similar.

go.mod
michaelkaplan13: Sorry I missed your other comment explaining the need for the change here. Would still prefer to keep our golang version consistent across the board for now. The motivation to stay a minor version behind is avoid the increased risk of bugs in the latest version

relayer/application_relayer.go
```diff
@@ -374,7 +429,7 @@ func (r *applicationRelayer) createSignedMessageAppRequest(unsignedMessage *aval
 			// Handle the responses. For each response, we need to call response.OnFinishedHandling() exactly once.
 			// Wrap the loop body in an anonymous function so that we do so on each loop iteration
 			// TODO: In order to run this concurrently, we need to route to each application relayer from the relayer responseChan
```
michaelkaplan13, line 431: Should be able to remove this TODO for now 👍

relayer/application_relayer.go
```diff
@@ -102,30 +106,84 @@ func newApplicationRelayer(
...
+		lock:              &sync.RWMutex{},
 	}
 
 	return &ar, nil
 }
 
-func (r *applicationRelayer) relayMessage(
-	unsignedMessage *avalancheWarp.UnsignedMessage,
+// Process [msgs] at height [height] by relaying each message to the destination chain.
+// Checkpoints the height with the checkpoint manager when all messages are relayed.
```
michaelkaplan13, line 127: Could you add to this comment explaining that an `ApplicationRelayer` instance expects `ProcessHeight` to be called for every block greater than or equal to its start block height?

messages/teleporter/message_manager.go
```diff
@@ -23,20 +23,18 @@ import (
...
-	teleporterMessageCache *cache.LRU[ids.ID, *teleportermessenger.TeleporterMessage]
-	destinationClients     map[ids.ID]vms.DestinationClient
+	messageConfig      Config
+	protocolAddress    common.Address
+	destinationClients map[ids.ID]vms.DestinationClient
+	logger             logging.Logger
+}
 
-	logger logging.Logger
+type messageHandler struct {
```
michaelkaplan13, line 33: Can we move `messageHandler` out to its own file? I found it a bit hard to find under `message_manager.go`

michaelkaplan13 (commented):
Really like all of the high level changes here. Left a handful of comments for possible clean up suggestions.

When zooming out to determine if these changes will have the desired effect of increasing the throughput of a single relayer instance significantly, the other comment I have is if `DestinationClient.SendTx` needs to be further optimized. Right now, for the EVM destination client, it look like the implementation makes a number of blocking RPC requests on each call to `SendTx` (see [here](https://github.com/ava-labs/awm-relayer/blob/main/vms/evm/destination_client.go#L117)). I think the only part of `SendTx` that actually needs to be synchronized for the EVM destination client is the assignment of the nonce to use for a given transaction, so we likely want to ensure that we only hold the lock when getting a local copy of the nonce to be used and incrementing the value for the next call.

peers/external_handler.go
```diff
@@ -147,23 +109,99 @@ func (h *RelayerExternalHandler) Disconnected(nodeID ids.NodeID) {
...
+
+	// Dispatch to the appropriate response channel
+	if responseChan, ok := h.responseChans[requestID]; ok {
+		responseChan <- inboundMessage
+	} else {
+		h.log.Debug("Could not find response channel for request", zap.Uint32("requestID", requestID))
+	}
+
+	// Check for the expected number of responses, and clear from the map if all expected responses have been received
+	responses := h.responsesCount[reqID.RequestID]
```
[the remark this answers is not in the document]
cam-schultz: It's not quite an invariant on `RegisterAppResponse`, since an AppResponse can be sent to the peer for any reason; it doest not necessarily need to be a response to a direct AppRequest. That said, there is a requirement on the responseChan dispatch and responseCount increment that `RegisterRequestID` has been called previously. This can be checked by adding a `return` statement after the debug log when checking if the responseChan for the request ID exists.

peers/external_handler.go
cam-schultz: AppResponses can come from anywhere, so this is an expected case. It's a no-op for the relayer, so we should return here.

relayer/application_relayer.go
cam-schultz: Good catch. Yes, requestID needs to be distinct on each call to `ProcessMessage`, but the ordering does not matter. Assigning to a local var in the critical section as suggested addressed this.

types/types.go
cam-schultz: In the expected case, `unsignedMsgBytes` is actually to abi encoded `SendMessage` event. This is parsed, and the `Message` field passed to `ParseUnsignedMessage`. 

The fallback case pertains only to manual Warp messages, in which `unsignedMsgBytes` is the Warp message itself, so can be directly passed to `ParseUnsignedMessage`.

relayer/listener.go
cam-schultz: That shouldn't be difficult, but I don't know how much we'd gain. As you said, adding concurrency here would only make sense if the RPC rtt is comparable to the block interval, which is not the expected case (~100ms vs ~1s). It is an interesting optimization opportunity though. I will create a ticket enumerating possible optimization targets and include that there.

messages/message_manager.go
cam-schultz: I renamed `MessageManager` to `MessageHandlerFactory`, and put both it and `MessageHandler` in `message_handler.go`.

go.mod
cam-schultz: Reverted back to go 1.21.10

relayer/application_relayer.go
cam-schultz: Thanks for pointing that out @geoff-vball . I elected to go with the workaround of assigning a local variable in each iteration rather than include that experimental flag in the build, since we're trying to lock into stable language features.

relayer/application_relayer.go
```diff
@@ -374,7 +429,7 @@ func (r *applicationRelayer) createSignedMessageAppRequest(unsignedMessage *aval
 			// Handle the responses. For each response, we need to call response.OnFinishedHandling() exactly once.
 			// Wrap the loop body in an anonymous function so that we do so on each loop iteration
 			// TODO: In order to run this concurrently, we need to route to each application relayer from the relayer responseChan
-			for response := range r.responseChan {
+			for response := range responseChan {
```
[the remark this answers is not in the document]
cam-schultz: Yeah, definitely a messy implementation. The motivation as you noticed is to achieve safer defer semantics, and unfortunately Go has no way to trigger actions at the close of a scope (as is the case with C++ RAII). I've added a standalone helper to clean this up as suggested.

messages/off-chain-registry/message_handler.go
```diff
@@ -84,17 +103,17 @@ func (m *messageManager) ShouldSendMessage(unsignedMessage *warp.UnsignedMessage
...
 			zap.String("destination", destination.String()),
-			zap.String("configuredRegistry", m.registryAddress.String()),
+			zap.String("configuredRegistry", m.factory.registryAddress.String()),
 		)
 		return false, nil
 	}
 
 	// Get the correct destination client from the global map
-	destinationClient, ok := m.destinationClients[destinationBlockchainID]
+	destinationClient, ok := m.factory.destinationClients[destinationBlockchainID]
```
geoff-vball, line 116: Would it make sense to pass the pointer to the destination client to the `ApplicationRelayer` instance, since each `ApplicationRelayer` only serves on src/dest pair? Then we don't have to deal with trying to grab it here.

messages/off-chain-registry/message_handler.go
```diff
@@ -28,17 +29,23 @@ const (
 	revertVersionNotFoundString        = "TeleporterRegistry: version not found"
 )
 
-type messageManager struct {
+type factory struct {
```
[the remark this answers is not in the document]
cam-schultz: I'd like to keep it internal to the package, that way the type can only be interacted with via the `MessageHandlerFactory` interface. Note that we explicitly declare the lhs at the callsite to `NewMessageHandlerFactory` as `messages.MessageHandlerFactory`. 

Same reasoning for `messageHandler`.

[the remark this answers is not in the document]
geoff-vball: Ah, then we should be returning a `messages.MessageHandlerFactory` from `NewMessageHandlerFactory`

messages/off-chain-registry/message_handler.go
cam-schultz: Good call out. I went ahead and added a destination client reference to the application relayer composition, and updated the message handler methods to accept the destination client directly.

messages/off-chain-registry/message_handler.go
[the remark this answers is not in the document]
cam-schultz: Updated 👍

relayer/listener.go
```diff
@@ -300,38 +232,21 @@ func (lstnr *Listener) ProcessLogs(ctx context.Context) error {
...
-					)
-					continue
-				}
+				// Dispatch all messages in the block to the appropriate application relayer.
+				// An empty slice is still a valid argument to ProcessHeight; in this case the height is immediately committed.
+				handlers := messageHandlers[appRelayer.relayerID.ID]
+
+				// Process the height async. This is safe because the ApplicationRelayer maintains the threadsafe
+				// invariant that heights are committed to the database one at a time, in order, with no gaps.
+				go appRelayer.ProcessHeight(block.BlockNumber, handlers)
```
cam-schultz, line 249: The first version of this PR called `appRelayer.ProcessHeight` synchronously. `ProcessHeight` manages the `ProcessMessage` goroutines, and blocks until all message workers have finished. As a result, the call here would block for each application relayer that the `Listener` manages. 

The fix is to call `ProcessMessage` async. I also had to add a lock acquisition to `checkpoint.StageProcessedHeight`, since there is no longer a guarantee that blocks finish processing in order.

relayer/listener.go
```diff
@@ -300,38 +244,18 @@ func (lstnr *Listener) ProcessLogs(ctx context.Context) error {
...
-						"Error relaying message",
-						zap.String("sourceBlockchainID", lstnr.sourceBlockchain.GetBlockchainID().String()),
-						zap.Error(err),
-					)
-					continue
-				}
+				// Dispatch all messages in the block to the appropriate application relayer.
+				// An empty slice is still a valid argument to ProcessHeight; in this case the height is immediately committed.
+				handlers := messageHandlers[appRelayer.relayerID.ID]
+				appRelayer.ProcessHeight(block.BlockNumber, handlers)
```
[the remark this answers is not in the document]
cam-schultz: I updated this call to be async, see here: https://github.com/ava-labs/awm-relayer/pull/288/files#r1610647753.

We could add an error wait group here and use the returned error, but that would require us to block until the block is fully processed by all application relayers before proceeding to the next block. There's no invariant that requires that. I'll think some more about how to best implement error handling here.

relayer/application_relayer.go
```diff
@@ -125,12 +129,12 @@ func NewApplicationRelayer(
 // Process [msgs] at height [height] by relaying each message to the destination chain.
 // Checkpoints the height with the checkpoint manager when all messages are relayed.
 // ProcessHeight is expected to be called for every block greater than or equal to the [startingHeight] provided in the cosntructor
-func (r *ApplicationRelayer) ProcessHeight(height uint64, messages []messages.MessageHandler) error {
+func (r *ApplicationRelayer) ProcessHeight(height uint64, handlers []messages.MessageHandler) error {
 	var eg errgroup.Group
-	for _, message := range messages {
-		m := message
+	for _, handler := range handlers {
+		h := handler
```
geoff-vball, line 135: Can we leave a comment on this line that this is to capture the loop variable, and can be removed when we move to go `1.22`

messages/message_handler.go
```diff
@@ -22,11 +23,11 @@ type MessageHandlerFactory interface {
 type MessageHandler interface {
 	// ShouldSendMessage returns true if the message should be sent to the destination chain
 	// If an error is returned, the boolean should be ignored by the caller.
-	ShouldSendMessage(destinationBlockchainID ids.ID) (bool, error)
+	ShouldSendMessage(destinationClient vms.DestinationClient) (bool, error)
 
 	// SendMessage sends the signed message to the destination chain. The payload parsed according to
 	// the VM rules is also passed in, since MessageManager does not assume any particular VM
-	SendMessage(signedMessage *warp.Message, destinationBlockchainID ids.ID) error
+	SendMessage(signedMessage *warp.Message, destinationClient vms.DestinationClient) error
```
geoff-vball: Now that the destination client is a field in the message handler, we don't need to pass it into these functions anymore

cam-schultz: The destination client is a field in the application relayer, not the message handler. The two objects are decoupled (see how they're fetched for a particular message in `Listener.GetAppRelayerMessageHandler`) and it would take some refactoring to couple them more closely. I'm not convinced that they should be coupled. In any case, I'd prefer to defer any more refactoring that's not strictly related to concurrent message processing to a later change.

relayer/listener.go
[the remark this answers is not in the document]
cam-schultz: I added an error channel parameter to `ProceeHeight`. The listener marks the health bool to false when that channel receives an error.

messages/message_handler.go
geoff-vball: Oh, got it.

vms/evm/destination_client.go
```diff
@@ -114,10 +114,6 @@ func (c *destinationClient) SendTx(signedMessage *avalancheWarp.Message,
 	gasLimit uint64,
 	callData []byte,
 ) error {
-	// Synchronize teleporter message requests to the same destination chain so that message ordering is preserved
```
geoff-vball, line 117: nit: can you put the first parameter of this function on a newline?

main/main.go
```diff
@@ -254,76 +265,178 @@ func main() {
...
-		zap.String("originBlockchainID", sourceSubnetInfo.BlockchainID),
+		zap.String("originBlockchainID", sourceBlockchain.BlockchainID),
 	)
 
 	// Wait for logs from the subscribed node
 	// Will only return on error or context cancellation
 	return listener.ProcessLogs(ctx)
 }
 
+func createApplicationRelayers(
```
michaelkaplan13, line 363: Can we add a doc comment about this function and its intended use? I think it creates the application relayers for a specific source subnet, in which case may want to rename to make that clear.

main/main.go
```diff
@@ -254,76 +265,178 @@ func main() {
...
+			"Failed to connect to node via RPC",
+			zap.String("blockchainID", sourceBlockchain.BlockchainID),
+			zap.Error(err),
+		)
+		return err
+	}
 
-	listener, err := relayer.NewListener(
+	// Create the ApplicationRelayers
+	applicationRelayers, minHeight, err := createApplicationRelayers(
```
michaelkaplan13, line 301: May want to consider moving the creation of application relayers out of `runListener` such that the application relayer instances can be used/passed else where if needed. 

For instance, in the case of an API to manually trigger the relaying of a message, I think it would want to route the message relay request to the respective application relayer if applicable. This way, the reference to the application relayers isn't isolated to the listeners.

main/main.go
```diff
@@ -254,76 +265,178 @@ func main() {
...
+		zap.String("blockchainID", sourceBlockchain.BlockchainID),
 	)
-
-	// Send any messages that were specified in the configuration
-	for _, warpMessage := range manualWarpMessages {
-		logger.Info(
-			"Relaying manual Warp message",
-			zap.String("blockchainID", sourceSubnetInfo.BlockchainID),
-			zap.String("warpMessageBytes", hex.EncodeToString(warpMessage.UnsignedMsgBytes)),
+	err = listener.ProcessManualWarpMessages(logger, manualWarpMessages, sourceBlockchain)
```
michaelkaplan13, line 344: I think the structure of the initialization flow here would be more readable and extendable if it roughly broke down as:
- Create the necessary application relayer instances.
- Create the necessary listener instances with the application relayers.
- Start `ProcessLogs` routines for each listener instance.
- Process the manual warp messages with the respective listener/application relayers. (I think it's a bit odd that `ProcessManualWarpMessages` is part of the `listener` since it isn't actually related to listening for events on that chain.)

geoff-vball (approved):

michaelkaplan13 (dismissed):

geoff-vball (approved):

geoff-vball (approved):

minghinmatthewlam (approved):

bernard-avalabs (commented):
Looks like this was merged before I finished my second pass, but it LGTM. Just a minor nit comment.

PR #302, opened 2024-05-24
Config split

najeal:
## Why this should be merged

This splits the config go file into multiple ones easier to read.
It shorten the BuildConfig function
It provides more informations about the config overwritten options.

## How this works

Split does not change logic.
About `BuildConfig()` changes:
- Instead of using multiple `viper.UnmarshalKey()`, a unique `viper.Unmarshal()` is done.
- a new function `NewConfig()` wraps `BuildConfig()` + `config Validation`. This makes the test of viper unmarshaling easier mechanism from `BuildConfig()`.
- **BuildConfig** does not return additional var to notify if some options have been overwritten.

 Information details about overwritten options are now embedded in a Config private field. It can be accessed from new methods `HasOverwrittenOptions()` & `GetOverwrittenOptions()`

## How this was tested

Modified unmarshal mechanism is tested in `viper_test.go`

## How is this documented


geoff-vball (dismissed):
Thanks for this! It overall looks good to me. @cam-schultz should also take a look at this

config/viper_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/stretchr/testify/require"
+
+	_ "embed"
+)
+
+//go:embed test_files/config.json
+var configFile string
+
+func TestBuildConfig(t *testing.T) {
+	v := viper.New()
```
cam-schultz, line 17: We can get rid of `test_files/config.json` and re-use the top-level `sample-relayer-config.json` for this test. The following changes are necessary:
1. Read in `"../sample-relayer-config.json` using `os.ReadFile`. `go:embed` does not support `..`.
```
cfgBytes, err := os.ReadFile("../sample-relayer-config.json")
configFile := string(cfgBytes)
```
2. Modify the expected `DestinationBlockchain` `AccountPrivateKey` to match the partial key provided in `sample-relayer-config.json`
3. Remote `test_files/config.json`

cam-schultz (commented):
Thanks for putting this together! Overall the changes look great. I have a couple of minor change requests.

config/viper_test.go
```diff
@@ -2,19 +2,17 @@ package config
...
-
-	_ "embed"
 )
 
-//go:embed test_files/config.json
-var configFile string
-
 func TestBuildConfig(t *testing.T) {
 	v := viper.New()
+	cfgBytes, err := os.ReadFile("../sample-relayer-config.json")
```
cam-schultz, line 14: Let's check the error here. This is also the cause of the failing lint job.

najeal: Sorry my mind was elsewhere. It should be ok now

cam-schultz: No worries! Thanks for the fix.

cam-schultz (approved):
Much appreciated 🙏

geoff-vball (approved):

PR #309, opened 2024-05-31
Refactor application relayers

geoff-vball:
## Why this should be merged
Integrates the first 3 bullet points from this comment https://github.com/ava-labs/awm-relayer/pull/288#discussion_r1619465642

## How this works
The main function now creates the application relayers instead of the listener. The application relayers are then passed to the listener instance.

## How is this documented

### DIFF
--- a/main/main.go
+++ b/main/main.go
@@ -234,27 +234,67 @@ func main() {
 			)
 			panic(err)
 		}
-		subnetInfo := s
+		sourceBlockchain := s
 
 		health := atomic.NewBool(true)
 		relayerHealth[blockchainID] = health
 
 		// errgroup will cancel the context when the first goroutine returns an error
 		errGroup.Go(func() error {
-			// runListener runs until it errors or the context is cancelled by another goroutine
-			return runListener(
+			// Dial the eth client
+			ethClient, err := ethclient.DialWithConfig(
+				context.Background(),
+				sourceBlockchain.RPCEndpoint.BaseURL,
+				sourceBlockchain.RPCEndpoint.HTTPHeaders,
+				sourceBlockchain.RPCEndpoint.QueryParams,
+			)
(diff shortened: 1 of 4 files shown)


main/main.go
```diff
@@ -234,27 +234,67 @@ func main() {
 			)
 			panic(err)
 		}
-		subnetInfo := s
+		sourceBlockchain := s
```
michaelkaplan13, line 237: Do we need this local var? I think we can just use `s` directly where needed, and maybe rename `s` to `sourceBlockchain` for clarity.

main/main.go
```diff
@@ -234,27 +234,67 @@ func main() {
...
+				logger.Error(
+					"Failed to connect to node via RPC",
+					zap.String("blockchainID", sourceBlockchain.BlockchainID),
+					zap.Error(err),
+				)
+				return err
+			}
+
+			// Create the ApplicationRelayers
+			applicationRelayers, minHeight, err := createApplicationRelayers(
```
michaelkaplan13, line 261: Should we move the creation of the `ethClient` and `applicationRelayers` out of the anonymous function run use `errGroup.Go`? I know we don't need references to them elsewhere right now, but could be useful in the future for routing messages to the application relayers otherwise.

michaelkaplan13 (commented):
Two small comments but otherwise LGTM.

main/main.go
geoff-vball: The thought process is that they're easy to hoist out of the goroutine if we need to, but right now processing them concurrently will save a bit of time.

main/main.go
michaelkaplan13: Ah, gets me every time...makes sense thanks

michaelkaplan13 (approved):

PR #310, opened 2024-05-31
update relayer gas estimate

### LINKED ISSUE #305: Relayer gas limit should account for message size
**Describe the bug**
Currently, the relayer EVM gas approximation for delivering Teleporter messages take into account the messages `requiredGasLimit`, a static overhead for calling `receiveCrossChainMessage` (including handling up to 5 receipts), and the number of warp signers. However, it does not account for the size of the message. Gas is charged per-byte of the of the message by the Warp precompile [here](https://github.com/ava-labs/subnet-evm/blob/master/precompile/contracts/warp/contract_warp_handler.go#L72). 

As a result, the implementation may not provide enough gas when trying to deliver relatively large messages, which could result in the transaction sent by the relayer unexpectedly reverting (see [example](https://43113.testnet.snowtrace.dev/tx/0x0c625bc0e075d9c502bbeadc72067bf44b64c11d1fa1746a231bfa5fa85fecea)).

This gas utils in the `teleporter` repo should be update to account for this, and the updated implementation used by the relayer.
https://github.com/ava-labs/teleporter/blob/main/utils/gas-utils/gas_utils.go#L26

minghinmatthewlam:
## Why this should be merged
Fixes #305 

## How this works
Uses updated gas estimate from `teleporter` that takes into account message size, receipt count, etc.

## How this was tested
CI

## How is this documented
n/a


michaelkaplan13 (approved):
Thanks for fixing this 🙏

geoff-vball (approved):

PR #317, opened 2024-06-04
Retry fetch logs

cam-schultz:
## Why this should be merged
The websocket connection serving new blocks to process and the RPC request to fetch logs from a block containing Warp messages may be served by different nodes that have indexed different heights. If the node serving the RPC request is behind the node serving blocks, then the `FilterLogs` call will fail. Adding retries the the `FilterLogs` RPC request works around this.

## How this works
See above

## How this was tested
CI

## How is this documented
N/A


types/types.go
```diff
@@ -43,14 +47,22 @@ func NewWarpBlockInfo(header *types.Header, ethClient ethclient.Client) (*WarpBl
...
 	if header.Bloom.Test(WarpPrecompileLogFilter[:]) {
-		logs, err = ethClient.FilterLogs(context.Background(), interfaces.FilterQuery{
-			Topics:    [][]common.Hash{{WarpPrecompileLogFilter}},
-			Addresses: []common.Address{warp.ContractAddress},
-			FromBlock: big.NewInt(int64(header.Number.Uint64())),
-			ToBlock:   big.NewInt(int64(header.Number.Uint64())),
-		})
-		if err != nil {
-			return nil, err
+		// The node serving the filter logs request may be behind the node serving the block header request,
```
michaelkaplan13, line 50: Would it be worth moving this out to a helper? I could see use tinkering with it in the future to adjust the delay/number of retries, etc.

types/types.go
```diff
@@ -43,14 +47,22 @@ func NewWarpBlockInfo(header *types.Header, ethClient ethclient.Client) (*WarpBl
...
+		// so we retry a few times to ensure we get the logs
+		for i := 0; i < filterLogsRetries; i++ {
+			logs, err = ethClient.FilterLogs(context.Background(), interfaces.FilterQuery{
+				Topics:    [][]common.Hash{{WarpPrecompileLogFilter}},
+				Addresses: []common.Address{warp.ContractAddress},
+				FromBlock: big.NewInt(int64(header.Number.Uint64())),
+				ToBlock:   big.NewInt(int64(header.Number.Uint64())),
```
geoff-vball, line 57: I think these can both just be `header.Number`

types/types.go
cam-schultz: Good call, added a helper

michaelkaplan13 (approved):

minghinmatthewlam (dismissed):

geoff-vball (approved):

michaelkaplan13 (approved):

bernard-avalabs (approved):
LGTM. Just a suggestion about using an exponential back off if the height difference can be significant.

types/types.go
```diff
@@ -101,3 +102,28 @@ func UnpackWarpMessage(unsignedMsgBytes []byte) (*avalancheWarp.UnsignedMessage,
...
+			Topics:    [][]common.Hash{{WarpPrecompileLogFilter}},
+			Addresses: []common.Address{warp.ContractAddress},
+			FromBlock: header.Number,
+			ToBlock:   header.Number,
+		})
+		if err == nil {
+			return logs, nil
+		}
+		if i != numRetries-1 {
+			time.Sleep(retryInterval)
```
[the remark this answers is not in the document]
cam-schultz: As part of the nonce ordering fix, I'm going to implement a generic retry helper, and will consider adding exponential backoffs there.

PR #323, opened 2024-06-06
Check receipt of sent transaction

### LINKED ISSUE #236: Emit error on tx revert
**Context and scope**
Transactions sent by the relayer should not revert under normal circumstances, but it doesn't hurt to emit an error to aid in debugging. Some scenarios in which the transaction may revert include:
- Bug in the message protocol contract
- Misconfigured source or destination blockchain (e.g. validators improperly registered)
- Misconfigured relayer (e.g. invalid private key)

In order to do this, we'll have to listen for the transaction to be included in a block and check the receipt status. For reference, we use this technique in many of the Teleporter and AWM Relayer end-to-end tests: https://github.com/ava-labs/teleporter/blob/8a71d0bbd121915d54792923a58d95e09bb837ba/tests/utils/utils.go#L545

### LINKED ISSUE #321: Check if transaction was mined
**Context and scope**
We currently don't check if a transaction was mined before incrementing the in-memory nonce. If the transaction was ejected from the mempool instead, then there will be a nonce mismatch, causing future transactions to fail to be mined as well.

We should retrieve the transaction receipt to confirm that it was mined, and return an error if it is not mined within some time period.

cam-schultz:
## Why this should be merged
Fixes #321 
Fixes #236

Modifies the retry mechanism introduced in #317 to retry a generic function in a loop. It runs until the context provided by the caller is cancelled, or the call is successful. The utility uses the same retry pattern implemented here: https://github.com/ava-labs/teleporter/blob/main/tests/utils/utils.go#L519

## How this works
The `MessageHandler.SendMessage` method waits until the API node returns the transaction receipt. It returns an error if either the receipt is not found within a time bound, or the receipt status is not successful. That error will be propagated to `ApplicationRelayer.ProcessHeight`, causing the height to not be committed to the database. The error will also be propagated to the `Listener`, which will mark the relayer as unhealthy.

Also cleans up some of the `awm-relayer/ethclient` usage. That package really only contains a custom `subnet-evm/ethclient` constructor that enables query parameters and http headers. Wrapping the `subnet-evm` type was overkill.

## How this was tested
CI. In the `AllowedAddresses` E2E test, the wait is increased before killing the relayer application to allow time for receipt fetching + the database write interval.

## How is this documented
N/A


vms/destination_client.go
```diff
@@ -22,7 +22,7 @@ import (
 type DestinationClient interface {
 	// SendTx contructs the transaction from warp primitives, and send to the configured destination chain endpoint
 	// TODO: Make generic for any VM.
-	SendTx(signedMessage *warp.Message, toAddress string, gasLimit uint64, callData []byte) error
+	SendTx(signedMessage *warp.Message, toAddress string, gasLimit uint64, callData []byte) (common.Hash, error)
```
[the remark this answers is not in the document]
cam-schultz: Sure thing, added.

geoff-vball (dismissed):

ethclient/client.go
```diff
@@ -15,10 +15,8 @@ import (
 
 var ErrInvalidEndpoint = errors.New("invalid rpc endpoint")
 
-type Client ethclient.Client
```
michaelkaplan13, line 18: If we're removing the `Client` type here, should we rename the package/file? 

I think would better belong under `utils` now, since it is just a utility for creating a `subnet-evm/ethclient` instance with headers/query params.

messages/teleporter/message_handler.go
```diff
@@ -236,6 +240,37 @@ func (m *messageHandler) SendMessage(signedMessage *warp.Message, destinationCli
...
+	if receipt.Status != types.ReceiptStatusSuccessful {
+		m.logger.Error(
+			"Transaction failed",
+			zap.String("destinationBlockchainID", destinationBlockchainID.String()),
+			zap.String("warpMessageID", signedMessage.ID().String()),
+			zap.String("teleporterMessageID", teleporterMessageID.String()),
+			zap.String("txHash", txHash.String()),
+		)
+		return fmt.Errorf("transaction failed with status: %d", receipt.Status)
+	}
```
michaelkaplan13, line 272: Can we move this section out to a helper? It also is close to duplicating the utility functions we have in Teleporter for use in the E2E such as [WaitMined](https://github.com/ava-labs/teleporter/blob/main/tests/utils/utils.go#L584). Would it make sense to export those from a non-test package to use here?

messages/teleporter/message_handler.go
```diff
@@ -236,6 +240,37 @@ func (m *messageHandler) SendMessage(signedMessage *warp.Message, destinationCli
...
+
 	m.logger.Info(
```
michaelkaplan13, line 274: Should this message be kept directly after the `SendTx` call? Alternatively, we could re-word it to say "delivered" rather than "sent". Would be great to add the `txHash` to it now that it's available also.

utils/utils.go
```diff
@@ -41,6 +47,33 @@ func CheckStakeWeightExceedsThreshold(accumulatedSignatureWeight *big.Int, total
...
+// Calls f until it returns a non-error result or the context is canceled, with a 200ms delay between calls.
+func CallWithRetry[T any](ctx context.Context, f func() (T, error)) (T, error) {
+	queryTicker := time.NewTicker(200 * time.Millisecond)
+	defer queryTicker.Stop()
+	var (
+		t   T
+		err error
+	)
+	for {
+		t, err = f()
```
bernard-avalabs, line 63: Should we state in the comments that we assume the time to call f() is significant less than the ticker delay? This should generally be true unless f() itself blocks or has some sleep logic. 

Otherwise the ticker will not have the intended effect, and we should only start the ticker after f() completes in each loop iteration.

bernard-avalabs (dismissed):
LGTM. Just a minor comment about the relationship between f() and our chosen tick interval.

messages/teleporter/message_handler.go
cam-schultz: I think this logic should continue to live here for a couple reasons:
- Moving `WaitMined` to a non-test utility in `teleporter` doesn't make a ton of sense, since there's nothing specific to Teleporter in that function. It's really a utility to support our current test framework
- `awm-relayer` has slightly different semantics. In `WaitMined`, we also wait for receipt's block height to account for differences in network views among nodes behind the public API. This is necessary to avoid races in the tests when run on Fuji. In the relayer, we instead assume that a valid receipt indicates that the tx will eventually be seen by all nodes.

messages/teleporter/message_handler.go
cam-schultz: Reworded to "delivered". This log message is intended to signal successful delivery.

utils/utils.go
cam-schultz: Go's `time.Ticker` type "will adjust the time interval or drop ticks to make up for slow receivers" ([docs](https://pkg.go.dev/time#NewTicker)). This is implemented using a buffered channel with a single element; if the client falls behind, new ticks are dropped.

I don't think we should make any assumptions on `f`, especially since `f` is usually a blocking network request. `time.Ticker` should account for arbitrarily expensive `f`s though.

geoff-vball (approved):

michaelkaplan13 (approved):
🚀

bernard-avalabs (approved):

Issue #314, opened 2024-06-03
Add API Handler for "relay message" requests


michaelkaplan13:
**Context and scope**
Add an API handler to the AWM relayer implementation that supports requesting the relayer relay a specific message via RPC. The relayer should looking up the message, decide if it's something that it is able to relay, and if so, initiate the `applicationRelayer` routine to relay it. This will be very helpful as a "catch-all" retry mechanism, and also allow the removal of manually configured off-chain warp messages in the future.

Currently, the relayer is only able to learn of messages to be relayed through it's listener on each source chain, or via manual configuration of off-chain warp messages. Providing an API to request a message get relayed provides a much more light weight mechanism for triggering re-tries in the event that a message is was not successfully relayed. That could occur in hypothetical situations were the message didn't include sufficient fees, or the transaction it sent previously to deliver a message revert for any reason. 

**Discussion and alternatives**
To make it easier for the API handler to route messages to the proper `ApplicationRelayer` instance (if one exists), we should likely move the logic to [get an application relayer and message handler ](https://github.com/ava-labs/awm-relayer/blob/app-relayer-worker/relayer/listener.go#L370) out of the `listener` and into a global singleton. Both the `listener` and API handler will then use that singleton to look up the application relayer / create a message handler to be used for a given message. This should pare down the `listener` to have less message handling logic also, making it easier to reason about. 

**Open questions**
- What are the required/option RPC params? Current thinking is:
    - Required: source subnet ID, warp message ID
    - Optional: block number message was sent in 
- Flag to enable/disable the API?
- Documentation around potential DOS vectors of the API should it be open to the public networking wise?
- Should any authentication be required? Probably makes sense to recommend it is only an RPC available locally.

cam-schultz:
Relevant ticket: https://github.com/ava-labs/awm-relayer/issues/171

PR #327, opened 2024-06-10
API endpoint for manually relaying warp message

geoff-vball:
## Why this should be merged
Closes https://github.com/ava-labs/awm-relayer/issues/314

## How this works
Creates a global `MessageCoordinator` that contains maps to Source Clients, Application Relayers, and Message Handler Factories.

TODOs:
- Make block number optional (implement iterative look-back for the message)
- Flag to enable/disable the API?
- Documentation around potential DOS vectors of the API should it be open to the public networking wise?
- Should any authentication be required? Probably makes sense to recommend it is only an RPC available locally.
- Currently we have no way to distinguish between the transaction being successfully sent and `ShouldSendMessage()` returning false.

## How this was tested
E2E Test

## How is this documented
TODO


relayer/relay_message_api_handler.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"encoding/json"
+	"math/big"
+	"net/http"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+const RelayMessageApiPath = "/relay-message"
```
cam-schultz, line 12: I prefer single name paths. Here, I think `/relay` would suffice.

relayer/relay_message_api_handler.go
```diff
@@ -0,0 +1,53 @@
...
+
+type RelayMessageRequest struct {
```
cam-schultz, line 14: Let's document that all of these params are required.

relayer/relay_message_api_handler.go
```diff
@@ -0,0 +1,53 @@
...
+	// cb58 encoding of the blockchain ID
+	BlockchainID string `json:"blockchain-id"`
+	// Hex encoding of the warp message ID
+	MessageID string `json:"message-id"`
+	// Integer representation of the block number
+	BlockNum string `json:"block-num"`
```
cam-schultz, line 20: Off-chain Warp messages do not have a block number associated with them. We should provide a route for fully specifying a Warp message so that we don't have to fetch from the source chain at a specific height.

I'm thinking we have `/relay` that accepts a Warp message ID as well as a height, and a /relay/message` that accepts the full unsigned message bytes, but no height.

relayer/relay_message_api_handler.go
cam-schultz, line 1: Let's document this endpoint in the top-level README

relayer/relay_message_api_handler.go
```diff
@@ -0,0 +1,53 @@
...
+
+		err := json.NewDecoder(r.Body).Decode(&req)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+
+		blockchainID, err := ids.FromString(req.BlockchainID)
+		if err != nil {
+			http.Error(w, "invalid blockchainID"+err.Error(), http.StatusBadRequest)
```
cam-schultz, line 35:
```suggestion
			http.Error(w, "invalid blockchainID: "+err.Error(), http.StatusBadRequest)
```

cam-schultz (commented):
I left a handful of comments and asks, and I'd like to think a bit more about the new `MessageCoordinator` type. But overall, the changes make sense and I'm a big fan of the trend of decoupling all of the components and coordinating them in `main`.

main/main.go
```diff
@@ -162,55 +167,72 @@ func main() {
...
+	relayer.SetMessageCoordinator(logger, messageHandlerFactories, applicationRelayers, sourceClients)
+
+	// Initialize the API after the message coordinator is set
+	http.Handle("/health", health.NewHandler(checker))
+	http.Handle(relayer.RelayMessageApiPath, relayer.RelayMessageAPIHandler())
+
+	// start the health check server
+	go func() {
+		log.Fatalln(http.ListenAndServe(fmt.Sprintf(":%d", cfg.APIPort), nil))
+	}()
```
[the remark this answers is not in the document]
geoff-vball: I'm not sure exactly what you mean here. Do you mean instantiating a new `ServeMux` instead of using the default one for everything here? Or encapsulating all this logic in a separate package?

main/main.go
```diff
@@ -102,18 +106,22 @@ func main() {
...
-		logger.Error(
-			"Failed to create destination clients",
-			zap.Error(err),
-		)
+		logger.Fatal("Failed to create destination clients", zap.Error(err))
+		panic(err)
+	}
+
+	// Initialize all source clients
+	logger.Info("Initializing destination clients")
```
minghinmatthewlam, line 114:
```suggestion
	logger.Info("Initializing source clients")
```

main/main.go
```diff
@@ -162,55 +167,72 @@ func main() {
...
 		panic(err)
 	}
 
 	// Initialize the global write ticker
 	ticker := utils.NewTicker(cfg.DBWriteIntervalSeconds)
 	go ticker.Run()
 
+	messageHandlerFactories, err := createMessageHandlerFactories(logger, &cfg)
+	if err != nil {
+		logger.Fatal("Failed to create Message Handler Factories", zap.Error(err))
```
minghinmatthewlam, line 198:
```suggestion
		logger.Fatal("Failed to create message handler factories", zap.Error(err))
```

main/main.go
```diff
@@ -227,136 +249,161 @@ func main() {
...
+			case config.OFF_CHAIN_REGISTRY:
+				m, err = offchainregistry.NewMessageHandlerFactory(
+					logger,
+					cfg,
+				)
+			default:
+				m, err = nil, fmt.Errorf("invalid message format %s", format)
+			}
+			if err != nil {
+				logger.Error("Failed to create message manager", zap.Error(err))
```
minghinmatthewlam, line 310:
```suggestion
				logger.Error("Failed to create message handler factory", zap.Error(err))
```

main/main.go
```diff
@@ -227,136 +249,161 @@ func main() {
...
-		applicationRelayers,
-		minHeight,
-		ethClient,
-	)
-	if err != nil {
-		return fmt.Errorf("failed to create listener instance: %w", err)
+) (map[ids.ID]map[common.Address]messages.MessageHandlerFactory, error) {
+	messageHandlerFactories := make(map[ids.ID]map[common.Address]messages.MessageHandlerFactory)
+	for _, sourceBlockchain := range globalConfig.SourceBlockchains {
+		messageHandlerFactoriesForSource := make(map[common.Address]messages.MessageHandlerFactory)
```
minghinmatthewlam, line 285: nit: can we rename to `messageProtocolHandlerFactory` or simiilar

main/main.go
```diff
@@ -162,55 +167,72 @@ func main() {
...
+		db,
+		ticker,
+		network,
+		messageCreator,
+		&cfg,
+		sourceClients,
+		destinationClients,
+	)
+	if err != nil {
+		logger.Fatal("Failed to create Application Relayers", zap.Error(err))
```
minghinmatthewlam, line 215:
```suggestion
		logger.Fatal("Failed to create application relayers", zap.Error(err))
``

File: relayer/message_coordinator.go [hunk 11]
```diff
@@ -0,0 +1,287 @@
...
+	"github.com/ava-labs/awm-relayer/database"
+	"github.com/ava-labs/awm-relayer/messages"
+	relayerTypes "github.com/ava-labs/awm-relayer/types"
+	"github.com/ava-labs/subnet-evm/core/types"
+	"github.com/ava-labs/subnet-evm/ethclient"
+	"github.com/ethereum/go-ethereum/common"
+	"go.uber.org/zap"
+)
+
+var globalMessageCoordinator *MessageCoordinator
```
minghinmatthewlam, line 22: I think we've generally avoided global vars, @cam-schultz is that correct? My personal preference has been to avoid them.

If we do decide to go with them though, should we have some checks that these are singleton instances?

main/main.go
```diff
@@ -132,10 +140,7 @@ func main() {
 		&cfg,
 	)
 	if err != nil {
-		logger.Error(
-			"Failed to create app request network",
-			zap.Error(err),
-		)
+		logger.Fatal("Failed to create app request network", zap.Error(err))
 		panic(err)
 	}
```
minghinmatthewlam, line 146: not part of this PR, but what hapepns when the function callback returns an error, I'm not sure where it gets handled, and how it's handled.

geoff-vball: Which function callback do you mean?

main/main.go
geoff-vball: Rename which?

relayer/relay_message_api_handler.go
```diff
@@ -0,0 +1,51 @@
...
+	"math/big"
+	"net/http"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+const RelayMessageApiPath = "/relay"
+
+type RelayMessageRequest struct {
```
minghinmatthewlam, line 14: should we have a `RelayMessageResponse` type as well?

relayer/message_coordinator.go
geoff-vball: Michael, Cam, and I landed on having a global object for these shared resources. With the current design, `ApplicationRelayer`s are inherently global, and we need some way to access them. We can't access the right one without the `MessageHandlerFactories`.

We _could_ make this is singleton, but I don't really see a benefit of enforcing only one instantiation of it (though I don't think we _should_ instantiate more than 1)

relayer/relay_message_api_handler.go
geoff-vball: In order to process a message in the relayer we need: `NetworkID`, `SourceAddress`, `SourceChainID`, and `Payload`. Do we want to specify all of these for this endpoint?

geoff-vball: The idea is to make the block number optional later, but it requires implementing an iterative lookback for on-chain messages.

main/main.go
minghinmatthewlam: `messageHandlerFactoriesForSource`

main/main.go
minghinmatthewlam: the health checker's check function https://github.com/ava-labs/awm-relayer/pull/327/files/6dc2441bb03dc49a9e1be1f5c768c913dda73254#diff-327181d0a8c5e6b164561d7910f4eeffd41442d55b2a2788fda2aa2692f17ec0R163

relayer/message_coordinator.go
minghinmatthewlam: was just thinking to have a builtin check that makes sure `globalMessageCoordinator` isn't overrode with another instance

main/main.go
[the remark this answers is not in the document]
cam-schultz: Sorry, my original comment was pretty unspecific. Now that we have two distinct endpoints, it might make sense to put them and their handlers into their own package, rather than the current organization with one handler in package `main`, and another in package `relayer`. I don't feel super strongly though, especially if it is a significant rework or breaks the existing dependency graph.

bernard-avalabs (commented):
Generally makes sense to me. I've left a few minor comments.

relayer/relay_message_api_handler.go
cam-schultz: I see two options:
1. Provide a separate endpoint that fully specifies the off-chain Warp message, including the fields you mentioned
2. Implement control flow to fetch the off-chain Warp message by it's ID using the Warp API. The downside here is that the public APIs disable the Warp API.

relayer/message_coordinator.go
cam-schultz: It's a semantic difference, but we have a bunch of variables that live at `package main` scope and are passed around, as opposed to static global vars that live at application scope. We confusingly use the the term "global" for the former. 

They're functionally the same, but in order to remain consistent, we should return the `MessageCoordinator` instance to `main` and leave it to `main` to ensure that only one instance exists. We can then pass the `MessageCoordinator` reference to all of the downstream objects that interact with it, similar to how we treat the `RelayerDatabase` and `AppRequestNetwork` global singletons.

main/main.go
geoff-vball: It returns a `503 (Service Unavailable)` HTTP response. https://github.com/alexliesenfeld/health/blob/main/examples/showcase/main.go

main/main.go
```diff
@@ -227,136 +249,161 @@ func main() {
...
-					zap.String("blockchainID", sourceBlockchain.BlockchainID),
-					zap.Error(err),
-				)
-				return err
-			}
-			logger.Info(
-				"Created application relayers",
-				zap.String("blockchainID", sourceBlockchain.BlockchainID),
-			)
+			return relayer.ProcessManualWarpMessages(manualWarpMessages[sourceBlockchain.GetBlockchainID()])
```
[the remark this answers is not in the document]
geoff-vball: Sorry, I already removed the manual warp messages from the config. Do you think we should keep them in there for a few releases with a deprecation note? I can't imagine many people were using them already given the difficulty in having to add them to the nodes as well.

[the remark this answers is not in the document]
cam-schultz: Yeah, we should deprecate features gracefully at this point, even though I agree that likely few people are using this particular feature. Apologies for not making that clear up front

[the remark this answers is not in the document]
geoff-vball: So a couple thoughts here. We don't initialize the logger for the application until after we finish parsing the config. So it would be a little hard to emit logs from here. Returning an error and failing validation seems a little too extreme here, is there some middle ground?

Do we still want to process the manual warp messages if they're included in the config? It should be easy enough to point the users to the new API.

main/main.go
```diff
@@ -191,180 +159,209 @@ func main() {
...
+		sourceClients,
+		destinationClients,
+	)
+	if err != nil {
+		logger.Fatal("Failed to create application relayers", zap.Error(err))
+		panic(err)
 	}
+	messageCoordinator := relayer.NewMessageCoordinator(logger, messageHandlerFactories, applicationRelayers, sourceClients)
+
+	// Each goroutine will have an atomic bool that it can set to false if it ever disconnects from its subscription.
```
cam-schultz, line 203:
```suggestion
	// Each Listener goroutine will have an atomic bool that it can set to false to indicate an unrecoverable error
```

main/main.go
```diff
@@ -191,180 +159,209 @@ func main() {
...
-		minHeight,
-		ethClient,
-	)
-	if err != nil {
-		return fmt.Errorf("failed to create listener instance: %w", err)
+) (map[ids.ID]map[common.Address]messages.MessageHandlerFactory, error) {
+	messageHandlerFactories := make(map[ids.ID]map[common.Address]messages.MessageHandlerFactory)
+	for _, sourceBlockchain := range globalConfig.SourceBlockchains {
+		messageHandlerFactoriesForSource := make(map[common.Address]messages.MessageHandlerFactory)
+		// Create message managers for each supported message protocol
```
cam-schultz, line 244:
```suggestion
		// Create message handler factories for each supported message protocol
```

messages/message_handler.go
```diff
@@ -27,7 +27,7 @@ type MessageHandler interface {
 
 	// SendMessage sends the signed message to the destination chain. The payload parsed according to
 	// the VM rules is also passed in, since MessageManager does not assume any particular VM
```
cam-schultz, line 29: Can you add to the interface comment what the returned common.Hash is intended to represent?

relayer/application_relayer.go
```diff
@@ -182,19 +184,17 @@ func (r *ApplicationRelayer) ProcessHeight(height uint64, handlers []messages.Me
 }
 
 // Relays a message to the destination chain. Does not checkpoint the height.
```
cam-schultz, line 186: Can you add to the funciton comment what the new return value represents?

tests/basic_relay.go
```diff
@@ -114,8 +114,8 @@ func BasicRelay(network interfaces.LocalNetwork) {
 	relayerIDA := database.CalculateRelayerID(subnetAInfo.BlockchainID, subnetBInfo.BlockchainID, database.AllAllowedAddress, database.AllAllowedAddress)
 	relayerIDB := database.CalculateRelayerID(subnetBInfo.BlockchainID, subnetAInfo.BlockchainID, database.AllAllowedAddress, database.AllAllowedAddress)
 	// Modify the JSON database to force the relayer to re-process old blocks
-	jsonDB.Put(relayerIDA, database.LatestProcessedBlockKey, []byte("0"))
-	jsonDB.Put(relayerIDB, database.LatestProcessedBlockKey, []byte("0"))
+	_ = jsonDB.Put(relayerIDA, database.LatestProcessedBlockKey, []byte("0"))
+	_ = jsonDB.Put(relayerIDB, database.LatestProcessedBlockKey, []byte("0"))
```
cam-schultz, line 118: Let's check the returned errors here, else leave it as is.

tests/relay_message_api.go
cam-schultz, line 1: Do we check that the message was successfully relayed in this test? If not, let's add that in.

tests/utils/utils.go
```diff
@@ -204,6 +204,7 @@ func CreateDefaultRelayerConfig(
 		MetricsPort:            9090,
 		SourceBlockchains:      sources,
 		DestinationBlockchains: destinations,
+		APIPort:                1234,
```
cam-schultz, line 207: Any reason not to use the default port 8080? That's what will be used if this field is not explicitly set.

relayer/message_coordinator.go
```diff
@@ -0,0 +1,254 @@
...
+			zap.Error(err),
+			zap.String("warpMessageID", warpMessage.UnsignedMessage.ID().String()),
+		)
+		return common.Hash{}, err
+	}
+
+	return appRelayer.ProcessMessage(handler)
+}
+
+func (mc *MessageCoordinator) ProcessMessage(blockchainID ids.ID, messageID common.Hash, blockNum *big.Int) (common.Hash, error) {
```
cam-schultz, line 187: I'd suggest renaming this to `ProcessMessageID` to distinguish it from `ApplicationRelayer.ProcessMessage`, which operates on the full message, and doesn't fetch it from the chain.

cam-schultz (commented):
Just a handful of minor comments, otherwise LGTM.

Let's also make sure to document all API routes (new and existing) in the README, and confirm with stakeholders that removing the manual warp message config won't break any existing functionaliity.

michaelkaplan13 (commented):
Could you add an example relay message request & responses to the README?

api/relay_message.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/ava-labs/awm-relayer/relayer"
+	"github.com/ava-labs/awm-relayer/types"
+	relayerTypes "github.com/ava-labs/awm-relayer/types"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+const (
+	RelayApiPath        = "/relay"
```
michaelkaplan13, line 17:
```suggestion
	RelayAPIPath        = "/relay"
```

types/types.go
```diff
@@ -107,3 +108,20 @@ func UnpackWarpMessage(unsignedMsgBytes []byte) (*avalancheWarp.UnsignedMessage,
...
+	})
+	if err != nil {
+		return nil, fmt.Errorf("could not fetch logs: %w", err)
+	}
+	if len(logs) != 1 {
+		return nil, fmt.Errorf("found more than 1 log: %d", len(logs))
+	}
+
+	return NewWarpMessageInfo(logs[0])
+}
```
michaelkaplan13, line 127: This feels a little odd to be in the `types` package. I think either `utils` or `relayer` would probably be a better fit for it.

tests/utils/utils.go
geoff-vball: I just needed to use something. The defaults don't get set in the tests. I've changed it to 8080

tests/basic_relay.go
geoff-vball: This sent me down a small rabbit hole... These were actually returning an error because the config wasn't validated, the `SupportedDestination` fields weren't populated, so `database.NewJSONFileStorage` wasn't populated with the correct relayer IDs.

I think it might make sense to add a `validated` field to the config struct, so that certain functions can check if the config has been validated before operating on it.

api/relay_message.go
```diff
@@ -0,0 +1,105 @@
...
+	MessageID string `json:"message-id"`
+	// Required. Integer representation of the block number
+	BlockNum string `json:"block-num"`
+}
+
+// Defines a manual warp message to be sent from the relayer through the API.
+type ManualWarpMessageRequest struct {
+	UnsignedMessageBytes []byte
+	SourceAddress        common.Address
+}
```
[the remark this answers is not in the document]
geoff-vball: This can be used for off-chain messages as well as on-chain messages if you want to provide the message manually.

relayer/message_coordinator.go
```diff
@@ -0,0 +1,254 @@
...
+	if !ok {
+		mc.logger.Error("Source client not found", zap.String("blockchainID", blockchainID.String()))
+		return common.Hash{}, fmt.Errorf("source client not set for blockchain: %s", blockchainID.String())
+	}
+
+	warpMessage, err := relayerTypes.FetchWarpMessageFromID(ethClient, messageID, blockNum)
+	if err != nil {
+		mc.logger.Error("Failed to fetch warp from blockchain", zap.String("blockchainID", blockchainID.String()), zap.Error(err))
+		return common.Hash{}, fmt.Errorf("could not fetch warp message from ID: %w", err)
+	}
```
cam-schultz: To simplify the `MessageCoorinator` exported interface, how about we lift this logic out to the caller, and combine `ProcessMessage` and `ProcessManualWarpMessage`?

geoff-vball: Good call :)

tests/basic_relay.go
cam-schultz: Rather than adding a field that's unused in the main application, I'd prefer to instead make sure `Config.Validate` is idempotent, and call it here in the test. On inspection, I think it should already be idempotent.

relayer/message_coordinator.go
cam-schultz: Yeah, we'd need to decide if the interface should accept only Warp Messages, or also IDs. Only accepting Warp Messages would yield a smaller interface, but would require more work on the caller's part as you described. I'm good to leave this as is for now.

relayer/message_coordinator.go
```diff
@@ -252,3 +240,20 @@ func (mc *MessageCoordinator) ProcessBlock(blockHeader *types.Header, ethClient
 		go appRelayer.ProcessHeight(block.BlockNumber, handlers, errChan)
 	}
 }
+
+func FetchWarpMessageFromID(ethClient ethclient.Client, warpID common.Hash, blockNum *big.Int) (*relayerTypes.WarpMessageInfo, error) {
```
cam-schultz, line 244: Naming nit: `FetchWarpMessage` seems sufficient to describe the function. Having the Warp ID in both the function name and as a parameter seems redundant.

README.md
```diff
@@ -316,6 +316,54 @@ The relayer consists of the following components:
...
 </div>
 
+### API
+
+#### `/relay`
+- Used to manually relay a warp message. The body of the request must contain the following JSON:
+```json
+{
+ "blockchain-id": "<cb58-encoding of blockchain ID>",
+ "message-id": "<Hex encoding of warp message ID>",
```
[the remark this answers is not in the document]
cam-schultz: We should also use `ids.ID` for `warpID` variables and arguments internally.

README.md
```diff
@@ -316,6 +316,54 @@ The relayer consists of the following components:
...
+ "block-num": <Integer representation of the block number that the message was sent in>
```
feuGeneA, line 327: I guess you left off the quotes around the value because it's an integer instead of a string, but the markdown render is in some weird highlight like its an error or something, pretty distracting.

README.md
```diff
@@ -316,6 +316,54 @@ The relayer consists of the following components:
...
+#### `/health`
+- Takes no arguments. Returns a `200` status code if all Application Relayers are healthy. Returns a `503` status if any of the Application Relayers have experienced an unrecoverable error. Here is an example return body:
+```json
+{
+  "status": "down",
+  "details": {
+    "relayers-all": {
+      "status": "down",
+      "timestamp": "2024-06-01T05:06:07.685522Z",
+      "error": "<List of cb-58 encoded IDs for unhealth relayers>"
```
feuGeneA, line 361:
```suggestion
      "error": "<List of cb-58 encoded IDs for unhealthy relayers>"
```
what kind of list is this? not a json array? a CSV?

api/relay_message.go
```diff
@@ -0,0 +1,128 @@
...
+			return
+		}
+
+		resp, err := json.Marshal(
+			RelayMessageResponse{
+				TransactionHash: txHash.Hex(),
+			},
+		)
+		if err != nil {
+			http.Error(w, "error writing response: "+err.Error(), http.StatusInternalServerError)
```
feuGeneA, line 81:
```suggestion
			http.Error(w, "error marshaling response: "+err.Error(), http.StatusInternalServerError)
```

api/relay_message.go
```diff
@@ -0,0 +1,128 @@
...
+			return
+		}
+
+		_, _ = w.Write(resp)
```
feuGeneA, line 85: Is there a reason not to do error checking here?

api/relay_message.go
```diff
@@ -0,0 +1,128 @@
...
```
feuGeneA, line 126: Is there a reason not to do error checking here?

api/relay_message.go
```diff
@@ -0,0 +1,128 @@
...
+			return
+		}
+
+		resp, err := json.Marshal(
+			RelayMessageResponse{
+				TransactionHash: txHash.Hex(),
+			},
+		)
+		if err != nil {
+			http.Error(w, "error writing response: "+err.Error(), http.StatusInternalServerError)
```
feuGeneA, line 122:
```suggestion
			http.Error(w, "error marshalling response: "+err.Error(), http.StatusInternalServerError)
```

api/relay_message.go
geoff-vball: So, if we error writing the response, there's really nothing we can tell the user, because we'd have to do so by writing something in the response... But I've plumbed the logger into our endpoint callers so we can log any errors here.

api/relay_message.go
```diff
@@ -21,7 +22,7 @@ const (
 type RelayMessageRequest struct {
 	// Required. cb58 encoding of the source blockchain ID for the message
 	BlockchainID string `json:"blockchain-id"`
-	// Required. Hex encoding of the warp message ID
+	// Required. cb58 encoding of the warp message ID
 	MessageID string `json:"message-id"`
 	// Required. Integer representation of the block number that the message was sent in
 	BlockNum string `json:"block-num"`
```
cam-schultz, line 28: Is there a reason we don't use an int here?

main/main.go
```diff
@@ -202,8 +201,8 @@ func main() {
 
 	// Each Listener goroutine will have an atomic bool that it can set to false to indicate an unrecoverable error
 	api.HandleHealthCheck(relayerHealth)
```
cam-schultz, line 203: Can we pass the logger to the health check handler as well, and emit a fatal log when triggered? I think those logs would be useful to have directly from the application, and it would make that handler interface consistent with the other two.

cam-schultz (commented):
Two more final comments, then I think we're good to go

api/relay_message.go
geoff-vball: Nope, changed :)

relayer/message_coordinator.go
```diff
@@ -0,0 +1,259 @@
...
+// processes multiple messages (using their corresponding MessageHandlers) in a single shot.
+func (mc *MessageCoordinator) getAppRelayerMessageHandler(
+	warpMessageInfo *relayerTypes.WarpMessageInfo,
+) (
+	*ApplicationRelayer,
+	messages.MessageHandler,
+	error,
+) {
+	// Check that the warp message is from a supported message protocol contract address.
+	messageHandlerFactory, supportedMessageProtocol := mc.messageHandlerFactories[warpMessageInfo.UnsignedMessage.SourceChainID][warpMessageInfo.SourceAddress]
```
michaelkaplan13, line 62: Nit: This line looks very long...I'm not too worried about it specifically but surprised we're not getting a linter error/warning in CI

michaelkaplan13 (commented):
Last few nits, but LGTM otherwise

relayer/message_coordinator.go
cam-schultz: It looks like this can be enabled with the `lll` linter: https://golangci-lint.run/usage/linters/#lll

geoff-vball: There are a good number of lines that don't fit in the line length, so I'll make a separate ticket for this.

cam-schultz (commented):
One last uber mega nit that I'm just noticing now - can we capitalize the first letters of the log message strings for consistency? I think the only files that need to be modified are `health_check.go` and `relay_message.go`

cam-schultz (approved):
LGTM! 🙏

bernard-avalabs (approved):
LGTM

PR #330, opened 2024-06-14
Add signature collection logs

cam-schultz:
## Why this should be merged
Adds additional logs in the signature collection code to aid debugging.

## How is this documented

### DIFF
--- a/relayer/application_relayer.go
+++ b/relayer/application_relayer.go
@@ -317,11 +317,17 @@ func (r *ApplicationRelayer) createSignedMessage(unsignedMessage *avalancheWarp.
 
 // createSignedMessageAppRequest collects signatures from nodes by directly querying them via AppRequest, then aggregates the signatures, and constructs the signed warp message.
 func (r *ApplicationRelayer) createSignedMessageAppRequest(unsignedMessage *avalancheWarp.UnsignedMessage, requestID uint32) (*avalancheWarp.Message, error) {
-	r.logger.Info("Fetching aggregate signature from the source chain validators via AppRequest")
+	r.logger.Info(
+		"Fetching aggregate signature from the source chain validators via AppRequest",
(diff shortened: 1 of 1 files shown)


relayer/application_relayer.go
```diff
@@ -317,11 +317,17 @@ func (r *ApplicationRelayer) createSignedMessage(unsignedMessage *avalancheWarp.
 
 // createSignedMessageAppRequest collects signatures from nodes by directly querying them via AppRequest, then aggregates the signatures, and constructs the signed warp message.
 func (r *ApplicationRelayer) createSignedMessageAppRequest(unsignedMessage *avalancheWarp.UnsignedMessage, requestID uint32) (*avalancheWarp.Message, error) {
-	r.logger.Info("Fetching aggregate signature from the source chain validators via AppRequest")
+	r.logger.Info(
+		"Fetching aggregate signature from the source chain validators via AppRequest",
+		zap.String("warpMessageID", unsignedMessage.ID().String()),
+		zap.String("sourceBlockchainID", r.sourceBlockchain.GetBlockchainID().String()),
+		zap.String("destinationBlockchainID", r.relayerID.DestinationBlockchainID.String()),
```
michaelkaplan13, line 324: Not necessary in this PR, but could be cleaner to have some logging helpers for repeated variables (such as "warpMessageID") to avoid the string literals in so many places.

michaelkaplan13 (approved):

relayer/application_relayer.go
cam-schultz: I've thought about this some, but haven't found any satisfying solution. Helpers would reduce repitition in the code, but aren't the most extensible or generic. A decorator-type extension to `logger.Logger` that always logs certain fields and still takes additional arguments would be ideal.

PR #344, opened 2024-06-24
`ShouldSendMessage` decider

feuGeneA:
## Why this should be merged
for #363 

## How this works
the relayer is informed of a decider service via config values that point to its host and port, and it communicates with that service process via grpc to make the decision.

## How this was tested
a simple decider implementation is included and incorporated into the tests


main/main.go
```diff
@@ -232,8 +245,101 @@ func main() {
...
+				)
+				panic(err)
+			}
+			logger.Info("Decider running")
+			<-ctx.Done()
+			logger.Info("global context cancelled, cancelling decider")
+			<-stopScanning
+			cancel()
+			<-cmdCtx.Done()
+			logger.Info("waiting for decider to finish")
```
feuGeneA, line 312: this is a mess of experimentation, and probably incorrect, but it's because I've yet to actually exercise a graceful shutdown of the process, because I think awm-relayer itself isn't being shut down gracefully. this mess should clean itself up easily once we actually start running through this code.

main/main.go
```diff
@@ -216,6 +225,33 @@ func main() {
 
 	// Create listeners for each of the subnets configured as a source
```
bernard-avalabs, line 226: Looks like this comment is duplicated on line 254. Was that intended?

main/main.go
```diff
@@ -216,6 +225,33 @@ func main() {
...
+		)
+		if err != nil {
+			logger.Fatal(
+				"Failed to instantiate decider client",
+				zap.Error(err),
+			)
+			panic(err)
+		}
+		runtime.SetFinalizer(grpcClient, func(c *grpc.ClientConn) { c.Close() })
+		grpcClient.WaitForStateChange(ctx, connectivity.Ready)
```
bernard-avalabs, line 251: I believe WaitForStateChange is currently marked as experimental:

https://pkg.go.dev/google.golang.org/grpc#ClientConn.WaitForStateChange

Not sure if this is an issue, just wanted to point it out.

messages/teleporter/message_handler.go
```diff
@@ -167,7 +183,47 @@ func (m *messageHandler) ShouldSendMessage(destinationClient vms.DestinationClie
 		return false, nil
 	}
 
-	return true, nil
+	var decision bool = true
```
bernard-avalabs, line 186: Should this be initialized to false? It looks like the error cases are just returning decision without modifying it, and I assume we should return false in those cases.

proto/README.md
```diff
@@ -0,0 +1,11 @@
+Protobuf linting and generation for this project is managed by
+[buf](https://github.com/bufbuild/buf).
+
+Please find installation instructions on
```
bernard-avalabs, line 4:
```suggestion
Please find installation instructions at
```

bernard-avalabs (commented):
Generally makes sense to me. I left a few comments.

tests/e2e_test.go
```diff
@@ -63,11 +73,31 @@ var _ = ginkgo.BeforeSuite(func() {
...
+	signalChan := make(chan os.Signal, 2)
+	signal.Notify(signalChan, os.Interrupt, syscall.SIGTERM)
+	go func() {
+		<-signalChan
+		cancelDecider()
+	}()
+	decider = exec.CommandContext(ctx, "./tests/cmd/decider/decider")
+	decider.Start()
+	log.Info("Started decider service")
+
```
geoff-vball, line 90: I don't think this is necessary. If ginkgo exits for whatever reason, everything that it was running should be cleaned up - we don't currently look for OS signals for the other processes we run.

What we _do_ need is something that watches the decider process, and if it exists abnormally, we should panic. Example here https://github.com/ava-labs/awm-relayer/blob/main/tests/utils/utils.go#L91

tests/e2e_test.go
```diff
@@ -63,11 +73,31 @@ var _ = ginkgo.BeforeSuite(func() {
...
 })
 
 var _ = ginkgo.AfterSuite(func() {
-	localNetworkInstance.TearDownNetwork()
+	if localNetworkInstance != nil {
+		localNetworkInstance.TearDownNetwork()
+	}
+	if decider != nil {
+		cancelDecider()
+	}
```
geoff-vball, line 100: We don't actually need this change either, afaik. Cleanup should only need to remove any files written that we want removed.

messages/teleporter/message_handler.go
```diff
@@ -167,7 +183,47 @@ func (m *messageHandler) ShouldSendMessage(destinationClient vms.DestinationClie
...
+	deciderClientValue := reflect.ValueOf(m.deciderClient)
+	if deciderClientValue.IsValid() && !deciderClientValue.IsNil() {
```
geoff-vball, line 188: Can we not just check `if m.deciderClient != nil`?

messages/teleporter/message_handler.go
cam-schultz: The intended control flow here is to separate the existing `ShouldSendMessage` decision logic from the `Decider` logic that is dispatched to the service. If the `Decider` fails, then we should return the prior decision from the existing logic.

I think this could be made clearer by adding a new helper, and calling it like so:

```
ShouldSendMessage() {
  // Existing decision logic
  decision, err :=  m.dispatchToDecider()
  if err != nil
    // log the error, but do not return it
    log.Warn("Decider returned error")
    return true, nil  
  }
  return decision
}
```

messages/teleporter/message_handler.go
```diff
@@ -167,7 +183,47 @@ func (m *messageHandler) ShouldSendMessage(destinationClient vms.DestinationClie
...
+		if err != nil {
+			m.logger.Error(
+				"Error decoding message ID",
+				zap.String("warpMsgIDStr", warpMsgIDStr),
+				zap.Error(err),
+			)
+			return decision, err
+		}
+
+		// TODO: add a timeout to the context
```
cam-schultz, line 201: Let's include timeouts in this PR

proto/README.md
```diff
@@ -0,0 +1,11 @@
...
+[https://docs.buf.build/installation/](https://docs.buf.build/installation/).
+
+Any changes made to proto definition can be updated by running
```
cam-schultz, line 7:
```suggestion
When changes are made to the proto definition, the generated source code can be updated by running
```

proto/decider/v1/decider.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
+
+package decider.v1;
```
cam-schultz, line 3: Is the intent behind `v1` to map `awm-relayer` versions to `Decider` versions? If so, I'd prefer if we followed the pattern used by Avalanchego / subnet-evm where protocol version [compatibility](https://github.com/ava-labs/subnet-evm?tab=readme-ov-file#avalanchego-compatibility) is provided out-of-band. Otherwise, the codebase would continue to grow with each new version.

config/config.go
```diff
@@ -58,6 +58,8 @@ type Config struct {
 	SourceBlockchains      []*SourceBlockchain      `mapstructure:"source-blockchains" json:"source-blockchains"`
 	DestinationBlockchains []*DestinationBlockchain `mapstructure:"destination-blockchains" json:"destination-blockchains"`
 	ProcessMissedBlocks    bool                     `mapstructure:"process-missed-blocks" json:"process-missed-blocks"`
+	DeciderHost            string                   `mapstructure:"decider-host" json:"decider-host"`
+	DeciderPort            *uint16                  `mapstructure:"decider-port" json:"decider-port"`
```
michaelkaplan13, line 62: Just calling this out because I wasn't aware how it worked exactly myself:

AvalancheGo communicates with VM plugins via gprc, but instead of assuming the VMs are running and providing their host/port in the configuration, it starts the configured VMs itself on the first port available on `127.0.0.1` [here](https://github.com/ava-labs/avalanchego/blob/master/vms/rpcchainvm/vm_client.go#L183).

That type of pattern may be less error prone for relayers to set up. i.e: the relayer config would specify the decider plugin(s) (if any) to run for a given source chain, and the application would handle initializing/configuring them itself.

Not suggesting we do this now, but could be worth considering going forward.

tests/e2e_test.go
feuGeneA: I took this part out and then the decider process was still running after the e2e tests completed successfully. I think we need it.

messages/teleporter/message_handler.go
feuGeneA: It's an interface, not a regular pointer. I googled and found this seemed to be the "right" way. Certainly open to suggestions though.

tests/e2e_test.go
feuGeneA: Are you sure it's not necessary? A little while ago I tried interrupting a test with Ctrl+C, and from the stdout it didn't look like it ran the `AfterSuite` hook, and the `decider` process was still running. And just now I tried again and saw that it _did_ execute that hook, and the decider process was _not_ running. So I think the behavior is inconsistent.

If the decider is left running/orphaned/defunct after the tests have finished, then it will fail to start in the next test run, because the port it uses is always the same and that port will be in use if it's already running from a previous test run. And if you're working on some changes totally unrelated to the decider, it could be pretty confusing and unclear what's going on.

This uncertainty and fear of confusion is why I decided to put this code into place. I'm happy to defer to the team's opinion though.

The idea of watching the decider for abnormal exit is a great idea, thanks, I'll do that.

config/config.go
iansuvak: I think that what Michael described is the generally preferred pattern for colocated services, and I agree that less config is better and less error prone. 

A counter example for why we would want to keep it, is that this allows for using remote deciders.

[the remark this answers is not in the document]
feuGeneA: The way I have it coded right now, you can skip specifying a host, and it will assume `localhost`, which felt nice for the case when the decider is running locally. It's nice, but not a huge deal. I'm definitely open to other opinions.

main/main.go
feuGeneA: I moved it, but I think the original crept back in as a merge conflict. Thanks for pointing it out. Addressed in https://github.com/ava-labs/awm-relayer/pull/344/commits/5c6422e8977a26bfbae4c62abc2678225547e3ae

main/main.go
feuGeneA: thanks for pointing that out, I removed it in https://github.com/ava-labs/awm-relayer/pull/344/commits/584556eccf9ea1c2fa23db5adffdb15b5d210fd4#diff-327181d0a8c5e6b164561d7910f4eeffd41442d55b2a2788fda2aa2692f17ec0

messages/teleporter/message_handler.go
feuGeneA: I extracted the logic into a new helper, and adapted my changes to `ShouldSendMessage` to fit the existing flow logic (if __ return false; if ___ return false; ...; return true), in https://github.com/ava-labs/awm-relayer/pull/344/commits/584556eccf9ea1c2fa23db5adffdb15b5d210fd4

main/main.go
```diff
@@ -216,6 +225,33 @@ func main() {
 
 	// Create listeners for each of the subnets configured as a source
 	errGroup, ctx := errgroup.WithContext(context.Background())
+
+	if cfg.DeciderPort != nil {
```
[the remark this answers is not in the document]
feuGeneA: addressed in https://github.com/ava-labs/awm-relayer/pull/344/commits/562cf7c383f81324f67ede3d812116c74ca93af3

main/main.go
```diff
@@ -33,9 +35,16 @@ import (
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/connectivity"
+	"google.golang.org/grpc/credentials/insecure"
 )
 
-var version = "v0.0.0-dev"
+var (
+	version = "v0.0.0-dev"
+
+	grpcClient *grpc.ClientConn // for connecting to the decider service
```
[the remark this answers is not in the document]
feuGeneA: addressed in https://github.com/ava-labs/awm-relayer/pull/344/commits/b01b14bba2fb7977fb2a9498467bc7ad0653416b

main/main.go
```diff
@@ -216,6 +225,33 @@ func main() {
...
+
+	if cfg.DeciderPort != nil {
+		port := strconv.FormatUint(uint64(*cfg.DeciderPort), 10)
+
+		host := cfg.DeciderHost
+		if len(host) == 0 {
+			host = "localhost"
+		}
+
+		grpcClient, err = grpc.NewClient(
```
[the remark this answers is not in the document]
feuGeneA: addressed in https://github.com/ava-labs/awm-relayer/pull/344/commits/b01b14bba2fb7977fb2a9498467bc7ad0653416b

messages/teleporter/message_handler.go
```diff
@@ -65,10 +72,18 @@ func NewMessageHandlerFactory(
 		return nil, err
 	}
 
+	var deciderClient pbDecider.DeciderServiceClient
+	if grpcClient == nil {
+		deciderClient = nil
+	} else {
+		deciderClient = pbDecider.NewDeciderServiceClient(grpcClient)
+	}
```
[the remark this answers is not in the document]
feuGeneA: addressed in https://github.com/ava-labs/awm-relayer/pull/344/commits/b01b14bba2fb7977fb2a9498467bc7ad0653416b

config/config.go
feuGeneA: Cam and I also discussed this. For this iteration I was going for the simplest thing possible, and sub-process management proved non-trivial, so I went with this host/port approach instead, since we said that ultimately we want to be able to support both local and remote deciders.

Going forward, I'm open to doing the sub-process/plugin approach; we _may_ be able to leverage the sub-process modules available in avalanchego, but if not then the complexity added to the code base for this would be a lot, at least as compared to the host/port approach.

Also, at the moment I'm leaning towards just the host/port approach so that there's one single config interface, rather than one for a local decider (the plugin path) and a separate/different one for a remote decider (the host/port).

README.md
```diff
@@ -101,7 +101,7 @@ awm-relayer --help                                      Display awm-relayer usag
 
 ### Building
 
-Before building, be sure to install Go, which is required even if you're just building the Docker image.
+Before building, be sure to install Go, which is required even if you're just building the Docker image. You'll also need to install [buf](github.com/bufbuild/buf/).
```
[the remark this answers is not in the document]
feuGeneA: In order to build, yes, because the relayer binary is built to be aware of the protobuf interfaces so that that same binary can be used later if you decide you do want to include those config options.

.github/workflows/e2e.yml
```diff
@@ -49,5 +49,10 @@ jobs:
         with:
           submodules: recursive
 
+      - name: Install buf
+        uses: bufbuild/buf-setup-action@v1.31.0
+        with:
+	  github_token: ${{ github.token }}
```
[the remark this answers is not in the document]
feuGeneA: Apparently not. I removed this in https://github.com/ava-labs/awm-relayer/pull/344/commits/ed3f4e76a7b280d7e33510aa3f96af183050112c and everything still ran fine. I had just copied this from avalanchego. Thanks for noticing it.

bernard-avalabs (commented):
LGTM other than a few remaining configuration/test-related comments from others that need to be addressed.

messages/teleporter/message_handler.go
feuGeneA: addressed in https://github.com/ava-labs/awm-relayer/pull/344/commits/f89bd35a55efdf0a470a13d42f44de78921490b1

proto/decider/v1/decider.proto
feuGeneA: the intent was just to follow the normal/standard behavior/conventions of buf+proto, at least as our starting point here. I stripped all of the custom lint rules out of the buf config file when copying it from avalanchego, and one of those was about needing to include the version in the service name.

in https://github.com/ava-labs/awm-relayer/pull/344/commits/3631b7bcf178e1b8e342434541fb84bd0b192ff5 I restored that linter rule and changed things to not include the version in the service/package names.

messages/teleporter/message_handler.go
```diff
@@ -183,47 +183,63 @@ func (m *messageHandler) ShouldSendMessage(destinationClient vms.DestinationClie
...
+			zap.String("warpMessageID", m.unsignedMessage.ID().String()),
+			zap.String("teleporterMessageID", teleporterMessageID.String()),
+			zap.String("destinationBlockchainID", destinationBlockchainID.String()),
+		)
+	}
+
+	return true, nil
+}
+
+func (m *messageHandler) deciderRejectedMessage() (bool, error) {
```
cam-schultz, line 206: I think this would be easier to read if the method returned the decision `bool` directly, and was renamed to `deciderAcceptedMessage`

messages/teleporter/message_handler.go
```diff
@@ -183,47 +183,63 @@ func (m *messageHandler) ShouldSendMessage(destinationClient vms.DestinationClie
...
+	if deciderRejectedMsg {
+		m.logger.Info(
+			"Decider rejected message",
+			zap.String("warpMessageID", m.unsignedMessage.ID().String()),
+			zap.String("teleporterMessageID", teleporterMessageID.String()),
+			zap.String("destinationBlockchainID", destinationBlockchainID.String()),
+		)
+	}
+
+	return true, nil
```
cam-schultz, line 203: If we rename the method to `deciderAcceptedMessage`, then we can emit a log if the returned bool is false, and directly return the variable.

main/main.go
```diff
@@ -469,6 +455,44 @@ func createApplicationRelayersForSourceChain(
...
+
+	uri := strings.Join(
+		[]string{host, strconv.FormatUint(uint64(*port), 10)},
+		":",
+	)
+
+	_, err := url.ParseRequestURI(uri)
+	if err != nil {
+		return nil, fmt.Errorf("Invalid URI: %w", err)
+	}
```
cam-schultz, line 475: I know that functionally this results in the same outcome (the application exiting), but can you move or replicate this config validation code to live in `package config` for consistency?

main/main.go
```diff
@@ -469,6 +455,44 @@ func createApplicationRelayersForSourceChain(
 	return applicationRelayers, minHeight, nil
 }
 
+func createDeciderClient(host string, port *uint16) (*grpc.ClientConn, error) {
```
cam-schultz, line 458: Can you add a comment explaining the requirements for `host` and `port`? Namely what combinations result in a `nil` client, and which result in an error.

README.md
```diff
@@ -291,6 +291,12 @@ The relayer is configured via a JSON file, the path to which is passed in via th
 
   - The AWS region in which the KMS key is located. Required if `kms-key-id` is provided.
 
+`"decider-host": string`
+
+`"decider-port": unsigned integer`
+
+- The network location of a service implementing the gRPC service defined by `proto/decider`, which will be queried for each message to determine whether that message should be relayed. If a port is specified but a host is not, the host is assumed to be `localhost`.
+
```
cam-schultz, line 299: Can you add to this that if host is specified then port is required?

messages/teleporter/message_handler.go
```diff
@@ -183,47 +183,63 @@ func (m *messageHandler) ShouldSendMessage(destinationClient vms.DestinationClie
...
 
-	var decision bool = true
+	deciderRejectedMsg, err := m.deciderRejectedMessage()
+	if err != nil {
+		m.logger.Warn(
+			"Error delegating to decider",
+			zap.String("warpMessageID", m.unsignedMessage.ID().String()),
+			zap.String("teleporterMessageID", teleporterMessageID.String()),
+		)
+	}
```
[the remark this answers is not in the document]
feuGeneA: I considered your suggestions but took a step back and decided on something slightly different, in https://github.com/ava-labs/awm-relayer/pull/344/commits/df26aa2b428fc59ec6f8af7e30a2b3d184148c41. Let me know what you think.

messages/teleporter/message_handler.go
feuGeneA: see https://github.com/ava-labs/awm-relayer/pull/344#discussion_r1690052917

messages/teleporter/message_handler.go
feuGeneA: see https://github.com/ava-labs/awm-relayer/pull/344#discussion_r1690052917

main/main.go
feuGeneA: ah I forgot about that existing validation logic, thanks for pointing this out. addressed in https://github.com/ava-labs/awm-relayer/pull/344/commits/a9c374c2377ed4b33ad86dafab7610c6881ffa40

main/main.go
feuGeneA: addressed in https://github.com/ava-labs/awm-relayer/pull/344/commits/6a9c27686d03fe38b1e03933335186ab23cc91f6

README.md
feuGeneA: addressed in https://github.com/ava-labs/awm-relayer/pull/344/commits/c37fd4176cec9ebac7b103ee654f8c49d6bd42bd

messages/teleporter/message_handler.go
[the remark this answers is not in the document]
cam-schultz: We should reserve returning pointers for large structs where we want to avoid deep copies. Returning pointers for built in types makes code less readable and more error prone, with no real performance gain.

README.md
```diff
@@ -295,7 +295,7 @@ The relayer is configured via a JSON file, the path to which is passed in via th
```
cam-schultz, line 295: It's a bit odd that there's no description for `decider-host`. Even a token description would be good. The description of the semantics when this field is omitted should be included here.

README.md
```diff
@@ -295,7 +295,7 @@ The relayer is configured via a JSON file, the path to which is passed in via th
 
 `"decider-port": unsigned integer`
 
-- The network location of a service implementing the gRPC service defined by `proto/decider`, which will be queried for each message to determine whether that message should be relayed. If a port is specified but a host is not, the host is assumed to be `localhost`.
+- The network location of a service implementing the gRPC service defined by `proto/decider`, which will be queried for each message to determine whether that message should be relayed. If a port is specified but a host is not, the host is assumed to be `localhost`. If a host is specified then a port is required.
```
cam-schultz, line 298: I've been thinking about this a bit, and IMO the semantics of a default `decider-host` set to `localhost` is unnecessarily complex. I think it's much easer to reason about if both fields are required to be populated, or both fields are required to be omitted. Thinking about somebody reading a config file, it seems like it would be confusing to see `decider-port` specified without a corresponding `decider-host`.

config/config.go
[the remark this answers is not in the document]
feuGeneA: addressed in https://github.com/ava-labs/awm-relayer/pull/344/commits/a9c374c2377ed4b33ad86dafab7610c6881ffa40

README.md
feuGeneA: Thanks for insisting. I changed it to use just a URL like you original suggested, in https://github.com/ava-labs/awm-relayer/pull/344/commits/7eb5112b149ca2ed3bf321f1a3a5d7d081c787d7

README.md
feuGeneA: I got rid of `decider-host` altogether. See https://github.com/ava-labs/awm-relayer/pull/344#discussion_r1690526033

cam-schultz (commented):
It's nice to see that moving from separate host+port config options to a unified URL was a strict lines of code decrease in all affected files.

One last thing: I think this comment may have fallen through the cracks. My bad for not directly including it in my last review! https://github.com/ava-labs/awm-relayer/pull/344#discussion_r1690248736

scripts/protobuf_codegen.sh
```diff
@@ -0,0 +1,56 @@
...
+fi
+
+## install "protoc-gen-go"
+PROTOC_GEN_GO_VERSION='v1.33.0'
+go install -v google.golang.org/protobuf/cmd/protoc-gen-go@${PROTOC_GEN_GO_VERSION}
+if [[ $(protoc-gen-go --version | cut -f2 -d' ') != "${PROTOC_GEN_GO_VERSION}" ]]; then
+  # e.g., protoc-gen-go v1.28.1
+  echo "could not find protoc-gen-go ${PROTOC_GEN_GO_VERSION}, is it installed + in PATH?"
+  exit 255
+fi
```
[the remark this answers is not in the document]
feuGeneA: addressed in https://github.com/ava-labs/awm-relayer/pull/344/commits/81b30bc212c5708d9dba60a78343f630d3b0fd57

messages/teleporter/message_handler.go
```diff
@@ -167,7 +183,47 @@ func (m *messageHandler) ShouldSendMessage(destinationClient vms.DestinationClie
...
+		if err != nil {
+			m.logger.Error(
+				"Error response from decider.",
+				zap.String("destinationBlockchainID", destinationBlockchainID.String()),
+				zap.String("teleporterMessageID", teleporterMessageID.String()),
+				zap.Any("warpMessageID", warpMsgIDStr),
+				zap.Error(err),
+			)
+			return decision, err
+		}
```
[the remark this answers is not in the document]
feuGeneA: The code shifted around since this comment and I think the comment no longer applies. (the "Outdated" label is appropriate for once haha.) Let me know if that's not the case and if there's still some improvement that could be made around this.

messages/teleporter/message_handler.go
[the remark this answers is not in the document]
feuGeneA: I thought it was also appropriate to use pointers for optional values, like we often do in structs, like I had originally done with the port value in the config struct. It felt weird to me to return a "decision" value while also returning an error, but googling just now I see that the value is generally considered undefined when the error is non-nil, so I guess that weirdness is just normality. :) addressed in https://github.com/ava-labs/awm-relayer/pull/344/commits/7e02e1ac549f556c3d3d9c4327f245232551e7a4

messages/teleporter/message_handler.go
```diff
@@ -183,32 +183,34 @@ func (m *messageHandler) ShouldSendMessage(destinationClient vms.DestinationClie
...
-		return *decision, nil
+		return decision, nil
 	}
 
 	return true, nil
 }
 
-func (m *messageHandler) getDeciderDecision() (*bool, error) {
+// queries the decider service to determine whether this message should be
+// sent. if the decider client is nil, returns true.
```
cam-schultz, line 208:
```suggestion
// Queries the decider service to determine whether this message should be
// sent. If the decider client is nil, returns true.
```

messages/teleporter/message_handler.go
```diff
@@ -183,32 +183,34 @@ func (m *messageHandler) ShouldSendMessage(destinationClient vms.DestinationClie
 		return false, nil
 	}
 
-	if decision, err := m.getDeciderDecision(); err != nil {
+	if decision, err := m.getShouldSendMessageFromDecider(); err != nil {
```
cam-schultz, line 186: I'm still not a huge fan of this control flow. The `return true, nil` at the end of this function is only reachable if `m.getShouldSendMessageFromDecider` errors, so IMO the code would be more readable if we returned at that point:

```
decision, err := m.getShouldSendMessageFromDecider()
if err != nil {
    // If the decider errors, default to true
    m.logger.Warn(
        "Error delegating to decider",
        zap.String("warpMessageID", m.unsignedMessage.ID().String()),
        zap.String("teleporterMessageID", teleporterMessageID.String()),
    )
    return true, nil
}

if !decision {
    m.logger.Info(
        "Decider rejected message",
        zap.String("warpMessageID", m.unsignedMessage.ID().String()),
        zap.String("teleporterMessageID", teleporterMessageID.String()),
        zap.String("destinationBlockchainID", destinationBlockchainID.String()),
    )
}

return decision, nil
```

This also removes the nested `if`. Lmk if you have other thoughts here.

messages/teleporter/message_handler.go
```diff
@@ -65,10 +72,18 @@ func NewMessageHandlerFactory(
 		return nil, err
 	}
 
+	var deciderClient pbDecider.DeciderServiceClient
+	if grpcClient == nil {
+		deciderClient = nil
```
geoff-vball, line 77: Setting an interface value to nil that will subsequently be called gives me the heebie-jeebies. Could we implement a struct `emptyDeciderServiceClient` that just returns true when `ShouldSendMessage` is called?

messages/teleporter/message_handler.go
```diff
@@ -167,9 +183,68 @@ func (m *messageHandler) ShouldSendMessage(destinationClient vms.DestinationClie
...
+	ctx, cancelCtx := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancelCtx()
+	response, err := m.deciderClient.ShouldSendMessage(
+		ctx,
+		&pbDecider.ShouldSendMessageRequest{
+			NetworkId:           m.unsignedMessage.NetworkID,
+			SourceChainId:       m.unsignedMessage.SourceChainID[:],
+			Payload:             m.unsignedMessage.Payload,
+			BytesRepresentation: m.unsignedMessage.Bytes(),
+			Id:                  warpMsgID,
```
geoff-vball, line 237:
```suggestion
	warpMsgID := m.unsignedMessage.ID()

	ctx, cancelCtx := context.WithTimeout(context.Background(), 30*time.Second)
	defer cancelCtx()
	response, err := m.deciderClient.ShouldSendMessage(
		ctx,
		&pbDecider.ShouldSendMessageRequest{
			NetworkId:           m.unsignedMessage.NetworkID,
			SourceChainId:       m.unsignedMessage.SourceChainID[:],
			Payload:             m.unsignedMessage.Payload,
			BytesRepresentation: m.unsignedMessage.Bytes(),
			Id:                  warpMsgID[:],
```

main/main.go
```diff
@@ -174,7 +177,20 @@ func main() {
 
 	relayerHealth := createHealthTrackers(&cfg)
 
-	messageHandlerFactories, err := createMessageHandlerFactories(logger, &cfg)
+	deciderClient, err := createDeciderClient(cfg.DeciderURL)
```
geoff-vball, line 180: This function is actually creating a `grpcClient`

main/main.go
```diff
@@ -240,6 +256,7 @@ func main() {
 func createMessageHandlerFactories(
 	logger logging.Logger,
 	globalConfig *config.Config,
+	deciderClient *grpc.ClientConn,
```
geoff-vball, line 259:
```suggestion
	grpcClient *grpc.ClientConn,
```

main/main.go
```diff
@@ -258,6 +275,7 @@ func createMessageHandlerFactories(
 					logger,
 					address,
 					cfg,
+					deciderClient,
```
geoff-vball, line 278:
```suggestion
					grpcClient,
```

main/main.go
```diff
@@ -432,6 +450,32 @@ func createApplicationRelayersForSourceChain(
 	return applicationRelayers, minHeight, nil
 }
 
+// create a client for the "should send message" decider service.
+// if url is unspecified, returns a nil client pointer
+func createDeciderClient(url string) (*grpc.ClientConn, error) {
```
geoff-vball, line 455:
```suggestion
func createGRPCClient(url string) (*grpc.ClientConn, error) {
```

messages/teleporter/message_handler.go
geoff-vball: Yeah, this is the flow I prefer when a function returns an `err` and a `bool`

main/main.go
feuGeneA: I actually used to call it a `grpcClient`, but Cam had me change it to `deciderClient`. I think Cam is right because this client instance is connected to a specific URL, and that URL will almost certainly be serving only the decider API (at least, as far as this context/scope needs to be aware of). I think the only situation in which it would make sense to generalize to `grpcClient` would be in the context of a client instance that will interact with a server using multiple different gRPC `service` interfaces.

messages/teleporter/message_handler.go
feuGeneA: thanks for pushing to make this great. addressed in https://github.com/ava-labs/awm-relayer/pull/344/commits/8ed81129aaeb1329a0eb2ef86e84542806ab392a

messages/teleporter/message_handler.go
feuGeneA: still new to Golang here, so thanks for bearing with me. a little research suggests that `nil` is the zero (uninitialized) value for interfaces, so this shouldn't matter, right? in any case, since this assignment is redundant anyways, I took it out in https://github.com/ava-labs/awm-relayer/pull/344/commits/682fead651913a9142ff49c59c52844b3a8b66b3.

is the "empty struct" thing a common idiom? I'm happy to defer to convention, but I'd appreciate a little reading material to help me understand the reasoning behind the idiom.

main/main.go
geoff-vball: So my issue here is that the struct `DeciderServiceClient` is created from this object in `NewMessageHandlerFactory`. We could call this one `deciderConnection`, but we should really distinguish them.

messages/teleporter/message_handler.go
```diff
@@ -183,47 +180,36 @@ func (m *messageHandler) ShouldSendMessage(destinationClient vms.DestinationClie
...
-// queries the decider service to determine whether this message should be
-// sent. if the decider client is nil, returns true.
+// Queries the decider service to determine whether this message should be
+// sent. If the decider client is nil, returns true.
 func (m *messageHandler) getShouldSendMessageFromDecider() (bool, error) {
 	deciderClientValue := reflect.ValueOf(m.deciderClient)
 
 	if !deciderClientValue.IsValid() || deciderClientValue.IsNil() {
 		return true, nil
 	}
```
cam-schultz, line 210: Since we're no longer directly assigning nil to `deciderClient`, we should be able to ditch reflection here: https://stackoverflow.com/a/13476765
```suggestion
	if m.deciderClient == nil {
		return true, nil
	}
```

messages/teleporter/message_handler.go
```diff
@@ -183,47 +180,36 @@ func (m *messageHandler) ShouldSendMessage(destinationClient vms.DestinationClie
 		return false, nil
 	}
```
cam-schultz, line 182:
```suggestion
// Dispatch to the external decider service. If the service is unavailable or returns
// an error, then use the decision that has already been made, i.e. return true
```

Let's add a comment to better separate the two "halves" of this function.

messages/teleporter/message_handler.go
feuGeneA: I took your suggestion about the empty struct thing in https://github.com/ava-labs/awm-relayer/pull/344/commits/f8662fdaf33fa9d196655df3ef485144d089e69d

messages/teleporter/message_handler.go
feuGeneA: I implemented [Geoff's suggestion to have an "empty client" implementation](https://github.com/ava-labs/awm-relayer/pull/344#discussion_r1692063683), so now the interface is always valid, so there's no need to do any check in this method. see https://github.com/ava-labs/awm-relayer/pull/344/commits/f8662fdaf33fa9d196655df3ef485144d089e69d

main/main.go
geoff-vball: Thank you, this is much clearer now :)

tests/e2e_test.go
geoff-vball: Yeah, I think we should move to running our e2e tests in a docker container (if feasible), that way we get the cleanup for free, and wouldn't need to check for OS signals, but until then, this is helpful.

geoff-vball (approved):
LGTM :)

cam-schultz (approved):
LGTM

PR #345, opened 2024-06-25
Warp API support

cam-schultz:
## Why this should be merged
Fixe #119. Exposes config options to fetch Warp message aggregate signatures via the Warp API rather than app request.

## How this works
Adds `SourceBlockchain.WarpAPIEndpoint` config option. Internally, `ApplicationRelayers` are composed of `[rpc.Client](https://github.com/ava-labs/subnet-evm/blob/master/rpc/client.go)` rather than `[warp.Client](https://github.com/ava-labs/subnet-evm/blob/master/warp/client.go)`. This is because `warp.Client` does not presently support query params or http headers, and expects the URI to be in a specific form.

## How this was tested
Added Warp API e2e test.

## How is this documented
Updated README

### DIFF
--- a/main/main.go
+++ b/main/main.go
@@ -122,13 +122,14 @@ func main() {
 
 	// The app request network generates P2P networking logs that are verbose at the info level.
 	// Unless the log level is debug or lower, set the network log level to error to avoid spamming the logs.
+	// We do not collect metrics for the network.
 	networkLogLevel := logging.Error
 	if logLevel <= logging.Debug {
 		networkLogLevel = logLevel
 	}
 	network, err := peers.NewNetwork(
 		networkLogLevel,
-		registerer,
+		prometheus.DefaultRegisterer,
 		&cfg,
 	)
 	if err != nil {
@@ -181,7 +182,14 @@ func main() {
 	}
 
 	// Initialize message creator passed down to relayers for creating app requests.
-	messageCreator, err := message.NewCreator(logger, registerer, "message_creator", constants.DefaultNetworkCompressionType, constants.DefaultNetworkMaximumInboundTimeout)
+	// We do not collect metrics for the message creator.
+	messageCreator, err := message.NewCreator(
+		logger,
+		prometheus.DefaultRegisterer,
+		"message_creator",
+		constants.DefaultNetworkCompressionType,
+		constants.DefaultNetworkMaximumInboundTimeout,
+	)
 	if err != nil {
 		logger.Error(
 			"Failed to create message creator",
--- a/main/main.go
+++ b/main/main.go
@@ -243,7 +243,7 @@ func main() {
 		// errgroup will cancel the context when the first goroutine returns an error
 		errGroup.Go(func() error {
 			// Dial the eth client
-			ethClient, err := utils.DialWithConfig(
+			ethClient, err := utils.NewEthClientWithConfig(
 				context.Background(),
 				sourceBlockchain.RPCEndpoint.BaseURL,
 				sourceBlockchain.RPCEndpoint.HTTPHeaders,
--- a/relayer/application_relayer.go
+++ b/relayer/application_relayer.go
@@ -70,6 +70,7 @@ type ApplicationRelayer struct {
 	checkpointManager *checkpoint.CheckpointManager
 	currentRequestID  uint32
 	lock              *sync.RWMutex
+	warpClient        warpBackend.Client
 }
 
 func NewApplicationRelayer(
@@ -108,6 +109,18 @@ func NewApplicationRelayer(
 	checkpointManager := checkpoint.NewCheckpointManager(logger, db, sub, relayerID, startingHeight)
 	checkpointManager.Run()
 
+	var warpClient warpBackend.Client
+	if !sourceBlockchain.UseAppRequestNetwork() {
+		warpClient, err = warpBackend.NewClient(sourceBlockchain.WarpAPIEndpoint.BaseURL, sourceBlockchain.GetBlockchainID().String())
+		if err != nil {
+			logger.Error(
+				"Failed to create Warp API client",
+				zap.Error(err),
+			)
+			return nil, err
+		}
+	}
+
 	ar := ApplicationRelayer{
 		logger:            logger,
 		metrics:           metrics,
@@ -121,6 +134,7 @@ func NewApplicationRelayer(
 		checkpointManager: checkpointManager,
 		currentRequestID:  rand.Uint32(), // TODONOW: pass via ctor
 		lock:              &sync.RWMutex{},
+		warpClient:        warpClient,
 	}
 
 	return &ar, nil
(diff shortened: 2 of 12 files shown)


tests/warp_api.go
```diff
@@ -0,0 +1,95 @@
...
+	relayerConfig := testUtils.CreateDefaultRelayerConfig(
+		[]interfaces.SubnetTestInfo{subnetAInfo, subnetBInfo},
+		[]interfaces.SubnetTestInfo{subnetAInfo, subnetBInfo},
+		teleporterContractAddress,
+		fundedAddress,
+		relayerKey,
+	)
+	// Enable the Warp API for all source blockchains
+	for _, subnet := range relayerConfig.SourceBlockchains {
+		subnet.WarpAPIEndpoint = subnet.RPCEndpoint
```
michaelkaplan13, line 51: Is there any (easy) way to check that the API endpoint is actually used by this test? Not necessary right now, but if the configuration changed in the future this test may silently pass if it falls back to using the P2P requests.

relayer/application_relayer.go
```diff
@@ -209,7 +229,7 @@ func (r *ApplicationRelayer) relayMessage(
 	startCreateSignedMessageTime := time.Now()
 	// Query nodes on the origin chain for signatures, and construct the signed warp message.
 	var signedMessage *avalancheWarp.Message
-	if useAppRequestNetwork {
+	if r.sourceBlockchain.UseAppRequestNetwork() {
```
michaelkaplan13, line 232: Nit: Should we check `if r.warpClient != nil` here instead? It looks correct as is, but the two independent checks on `r.sourceBlockchain.UseAppRequestNetwork()` feel error prone for prevent a nil pointer dereference.

relayer/application_relayer.go
```diff
@@ -70,6 +71,7 @@ type ApplicationRelayer struct {
 	checkpointManager *checkpoint.CheckpointManager
 	currentRequestID  uint32
 	lock              *sync.RWMutex
+	warpClient        *rpc.Client
```
michaelkaplan13, line 74: Nit: Can we call this `sourceWarpSignatureClient` to be super self-descriptive?

michaelkaplan13 (commented):
Two small nits, but LGTM. Thanks for putting this together so quickly.

Totally fine with punting on the question in the E2E regarding checking that the API is actually used for now; worth thinking about though IMO

relayer/application_relayer.go
geoff-vball: Can we add a comment that this might be nil?

relayer/application_relayer.go
```diff
@@ -229,7 +229,9 @@ func (r *ApplicationRelayer) relayMessage(
 	startCreateSignedMessageTime := time.Now()
 	// Query nodes on the origin chain for signatures, and construct the signed warp message.
 	var signedMessage *avalancheWarp.Message
-	if r.sourceBlockchain.UseAppRequestNetwork() {
+
+	// sourceWarpSignatureClient is nil iff the source blockchain is configured to fetch signatures via AppRequest
```
geoff-vball, line 233: Can we also add this comment to the `ApplicationRelayer` struct?

tests/warp_api.go
cam-schultz: Hmm that's a good question that I don't immediately know the best answer to. We could add distinct metrics and monitor those from the test. Those metrics would be useful in a deployment as well. I'll look into this.

cam-schultz: Added those metrics in and now check them in the test.

geoff-vball (approved):

relayer/application_relayer_metrics.go
```diff
@@ -53,9 +55,35 @@ func NewApplicationRelayerMetrics(registerer prometheus.Registerer) (*Applicatio
 	}
 	registerer.MustRegister(failedRelayMessageCount)
 
+	fetchSignatureAppRequestCount := prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "fetch_signature_app_request_count",
+			Help: "Number messages signed via AppRequest",
```
michaelkaplan13, line 61:
```suggestion
			Help: "Number messages aggregate signatures constructed via AppRequest",
```

relayer/application_relayer_metrics.go
```diff
@@ -53,9 +55,35 @@ func NewApplicationRelayerMetrics(registerer prometheus.Registerer) (*Applicatio
...
+	)
+	if fetchSignatureAppRequestCount == nil {
+		return nil, ErrFailedToCreateApplicationRelayerMetrics
+	}
+	registerer.MustRegister(fetchSignatureAppRequestCount)
+
+	fetchSignatureRPCCount := prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "fetch_signature_rpc_count",
+			Help: "Number messages signed via Warp API",
```
michaelkaplan13, line 73:
```suggestion
			Help: "Number aggregate signatures fetched via Warp API",
```

relayer/application_relayer.go
```diff
@@ -58,18 +59,19 @@ var (
...
+	messageCreator            message.Creator
+	sourceBlockchain          config.SourceBlockchain
+	signingSubnetID           ids.ID
+	destinationClient         vms.DestinationClient
+	relayerID                 database.RelayerID
+	warpQuorum                config.WarpQuorum
+	checkpointManager         *checkpoint.CheckpointManager
+	currentRequestID          uint32
+	lock                      *sync.RWMutex
+	sourceWarpSignatureClient *rpc.Client // nil if the source blockchain is configured to sign messages via AppRequest
```
michaelkaplan13, line 74:
```suggestion
	sourceWarpSignatureClient *rpc.Client // nil if configured to fetch signatures via AppRequest for the source blockchain
```

tests/warp_api.go
```diff
@@ -0,0 +1,131 @@
...
+	//
+	log.Info("Verifying the messages were signed using the Warp API")
+	resp, err := http.Get(fmt.Sprintf("http://localhost:%d/metrics", relayerConfig.MetricsPort))
+	Expect(err).Should(BeNil())
+
+	body, err := io.ReadAll(resp.Body)
+	Expect(err).Should(BeNil())
+	defer resp.Body.Close()
+
+	metricName := "app_fetch_signature_rpc_count"
```
michaelkaplan13, line 109: Super nit: may want to move to a const

main/main.go
```diff
@@ -122,13 +122,14 @@ func main() {
...
 	// Unless the log level is debug or lower, set the network log level to error to avoid spamming the logs.
+	// We do not collect metrics for the network.
 	networkLogLevel := logging.Error
 	if logLevel <= logging.Debug {
 		networkLogLevel = logLevel
 	}
 	network, err := peers.NewNetwork(
 		networkLogLevel,
-		registerer,
+		prometheus.DefaultRegisterer,
```
cam-schultz, line 132: While adding in the new get signature metrics, I noticed that we also collect a TON of unrelated network and app requests metrics. This disables these extra metrics from being collected.

michaelkaplan13 (approved):
🙏

geoff-vball (approved):

Issue #352, opened 2024-07-02
Enable line length linter


geoff-vball:
**Context and scope**
We currently don't check for line length with the linter in this repo. This can be done by adding `lll` to `.golangci.yml`.

PR #354, opened 2024-07-03
Add lll linter

geoff-vball:
## Why this should be merged
Closes https://github.com/ava-labs/awm-relayer/issues/352

## How this works
Enables the `lll` linter and conforms (almost) all line lengths to 120 characters.

There are some lines where I couldn't come up with a good was to get the line under 120 characters, so used `nolint:lll`. If you have suggestions for any of these lines, please give them :)


database/utils.go
```diff
@@ -18,7 +18,8 @@ func IsKeyNotFoundError(err error) bool {
 
 // Determines the height to process from. There are three cases:
 // 1) The database contains the latest processed block data for the chain
-//   - In this case, we return the maximum of the latest processed block and the configured processHistoricalBlocksFromHeight
+//   - In this case, we return the maximum of the latest processed block and the
+//     configured processHistoricalBlocksFromHeight
```
iansuvak, line 22:
```suggestion
//     configured processHistoricalBlocksFromHeight.
```

nit but you caught the missing period in `config/source_blockchain.go` so might as well here. Although it looks like I'm not in the `CODEOWNERS`file for this repo yet.

relayer/message_coordinator.go
```diff
@@ -76,6 +77,7 @@ func (mc *MessageCoordinator) getAppRelayerMessageHandler(
 	}
 
 	// Fetch the message delivery data
+	//nolint:lll
 	sourceBlockchainID, originSenderAddress, destinationBlockchainID, destinationAddress, err := messageHandler.GetMessageRoutingInfo()
```
cam-schultz, line 81: Not for this PR, but these are the exact fields that compose a `RelayerID`. It might make sense to directly return that struct instead of the individual fields.

config/source_blockchain.go
```diff
@@ -16,15 +16,15 @@ import (
 // Specifies the height from which to start processing historical blocks.
 type SourceBlockchain struct {
 	SubnetID                          string                           `mapstructure:"subnet-id" json:"subnet-id"`
-	BlockchainID                      string                           `mapstructure:"blockchain-id" json:"blockchain-id"`
+	BlockchainID                      string                           `mapstructure:"blockchain-id" json:"blockchain-id"` //nolint:lll
```
cam-schultz, line 19: Not for this PR, but we should look into if we still need both mpastructure and json tags. IIRC, the json tag is only used in tests.

config/source_blockchain.go
```diff
@@ -104,18 +104,25 @@ func (s *SourceBlockchain) Validate(destinationBlockchainIDs *set.Set[string]) e
 			return fmt.Errorf("invalid blockchainID in configuration. error: %w", err)
 		}
 		if !destinationBlockchainIDs.Contains(dest.BlockchainID) {
-			return fmt.Errorf("configured source subnet %s has a supported destination blockchain ID %s that is not configured as a destination blockchain",
+			return fmt.Errorf(
+				"configured source subnet %s has a supported destination blockchain ID %s that is not configured as a destination blockchain", //nolint:lll
```
cam-schultz, line 108: AFAIK, the only other way to preserve these long strings is through concatenation, where we'd put each section on its own line. Not the cleanest, so I think the `nolint` tag is appropriate in these instances.

cam-schultz (approved):
LGTM. Left some food for thought on how we might tackle the lines that require the `nolint` tag.

PR #357, opened 2024-07-05
Golang 1.21.12

michaelkaplan13:
## Why this should be merged

Bump to latest 1.21 Golang release. Removes unnecessary explicit version used by actions/setup-go in favor of pulling directly from go.mod.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,10 +19,6 @@ jobs:
           path: awm-relayer
           submodules: recursive
 
-      - name: Set Go version
-        run: |
-          source ./awm-relayer/scripts/versions.sh
-
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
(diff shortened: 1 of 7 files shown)


.github/workflows/release.yml
```diff
@@ -19,10 +19,6 @@ jobs:
           path: awm-relayer
           submodules: recursive
 
-      - name: Set Go version
-        run: |
-          source ./awm-relayer/scripts/versions.sh
```
cam-schultz, line 24: We still need to set `GO_VERSION` to provide as an arg to the Dockerfile build.

cam-schultz (dismissed):

cam-schultz (approved):
LGTM. We should do a test run of the release job to make sure it still works. We can do so by pushing an rc tag

iansuvak (approved):

Issue #355, opened 2024-07-03
Consider supporting hex and cb58 ids in API interfaces


cam-schultz:
**Context and scope**
https://github.com/ava-labs/awm-relayer/pull/345 added a dedicated API package. Methods byte fields at present accept arguments in cb58 or hex encoding, but not both. For fields such as TxID, this is not an issue since they are always represented in hex. Fields such as blockchainID are less consistent - they are often written out in cb58 format, but these IDs are also commonly extracted from hex encoded sources, such as raw Tx data. Further, there doesn't exist great tooling to convert hex -> cb58 and vice-versa. 

To better support this, we should consider supporting both encoding types for certain fields in the relayer API. We can limit this to extending cb58 fields to also accept hex. There shouldn't be any collision between encodings for fixed size data, since the length of the string for a fixed number of bytes will differ depending on the encoding scheme. However, as an extra precaution (and to support up front unbounded length byte arrays) we should require hex fields be prepended with `0x`.

PR #360, opened 2024-07-08
Decode hex or cb58

geoff-vball:
## Why this should be merged
Closes https://github.com/ava-labs/awm-relayer/issues/355

## How this works
If the string is prepended by '0x', attempts to decode from hex. Otherwise, tries to decode as cb58. The checksum in cb58 should prevent accidental collisions.

## How this was tested
Unit tests added

## How is this documented
Added hex as option in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,11 +199,11 @@ The relayer is configured via a JSON file, the path to which is passed in via th
 
   `"source-blockchain-id": string`
 
-  - cb58-encoded blockchain ID of the source blockchain.
+  - Hex or cb58-encoded blockchain ID of the source blockchain.
 
   `"destination-blockchain-id": string`
 
-  - cb58-encoded blockchain ID of the destination blockchain.
+  - Hex or cb58-encoded blockchain ID of the destination blockchain.
 
   `"source-address": string`
 
--- a/README.md
+++ b/README.md
@@ -199,11 +199,11 @@ The relayer is configured via a JSON file, the path to which is passed in via th
 
   `"source-blockchain-id": string`
 
-  - Hex or cb58-encoded blockchain ID of the source blockchain.
+  - cb58 encoded or "0x" prefixed Hex encoded blockchain ID of the source blockchain.
 
   `"destination-blockchain-id": string`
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -199,11 +199,11 @@ The relayer is configured via a JSON file, the path to which is passed in via th
 
   `"source-blockchain-id": string`
 
-  - cb58-encoded blockchain ID of the source blockchain.
+  - Hex or cb58-encoded blockchain ID of the source blockchain.
```
bernard-avalabs, line 202: Should this mention that there is a "0x" prefix for hex values?

bernard-avalabs (dismissed):
LGTM. Perhaps the README should mention that hex-encoded strings need to include a 0x prefix.

bernard-avalabs (approved):

minghinmatthewlam (commented):
mostly looks good for APIs, but do we want to support both cb58 and hex for the main config as well for blockchain IDS? Thinking it'd be strange to support cb58/hex for relay requests, but not keep it consistent for the main config

utils/utils.go
```diff
@@ -123,3 +124,15 @@ func StripFromString(input, substring string) string {
 
 	return strippedString
 }
+
+// Converts a '0x'-prefixed hex string or cb58-encoded string to an ID.
```
cam-schultz, line 128: Can you add a comment here that input length validation is handled by the `ids` package?

cam-schultz (commented):
One tiny comment, otherwise LGTM. Thanks for implementing this

michaelkaplan13 (dismissed):

utils/utils.go
cam-schultz: Looks like this comment was resolved, but I don't see a corresponding commit.

geoff-vball: Sorry, I pushed but I think I typed my yubikey pin wrong 😓

cam-schultz (approved):

michaelkaplan13 (approved):

PR #368, opened 2024-07-15
POC Signature aggregation API

### LINKED ISSUE #361: [POC] -  Warp Signature Aggregation API
**Context and scope**
Currently the AWM relayer is an all in one service, fetching the signatures from the subnet validators, deciding whether it should relay the message and actually doing it's own relaying. We can provide additional functionality by defining a separate service that will take a message to be signed as an input and attempt to aggregate signatures from the relevant subnet validators. It can then return a valid signed message to the caller that they themselves can deliver to the target chain. 

The scope of this issue is just an initial implementation following the existing examples in this repo and the subnet-evm repo. This should be a separate package/build artifact from the existing awm-relayer. 

What should the API design be? Should we return just the signature or the whole signed message?
Should we do base64 encoded JSON or a gRPC service with an optional gRPC gateway in the front?

**Open questions**
Should the code for this service live in this repo long term? If so should we rename it?

iansuvak:
## Why this should be merged
This will close #361 and represents a minimal mergeable implementation of the Warp Signature aggregation API. It maintains all functionality of the existing relayer implementation and adds the new binary running the signature aggregation portion. 

## How this works
- Moves the logic requesting signatures that was previously in the `application_relayer.go` to it's own file in `signature-aggregator/aggregator.go` which is imported by the relayer. 
- It re-uses the existing `peers` package including `AppRequestNetwork` but slightly modifies it to be able to accept the simplified config from signature-aggregator
- It doesn't maintain any persistent state or follow the chains and instead expects the user to supply the raw bytes to be signed. It uses the legacy AppRequest interface to request signatures but will switch to using the ACP-118 defined once it's implemented in coreth. subnet-evm implementation is already done. 
- The required configuration is minimal by design and doesn't support manually defining source/destination chains/subnets. 

### Code organization
- Currently there is anew `signature-aggregator` directory containing all files specific to the new API including it's config and network implementation. 
- There is a new `lib` folder containing some common interfaces and structs used by both configs and networks. Some of the network code in particular should be deduplicated. 
- Minor changes were made to the `awm-relayer` network implementation to be able to share interfaces for signature aggregation in the future. 

## How this was tested
Existing e2e tests pass and a new e2e test was added. Note that currently it will only work for messages originating on the P-Chain until avalanche go [1.11.11](https://github.com/ava-labs/avalanchego/milestone/33) (and matching coreth/subnet-evm) is released as a required networking fix to bypass trackesSubnets check has only recently been merged to `master` branch. 

## How is this documented
README.md with overview and example manual testing flow was added.


michaelkaplan13:
> The interface aims to be destination chain agnostic and doesn't ask for destination blockchain. Although we might have to add it as a required field when the source is pChain?

I don't think so. For messages from P-chain -> Subnet, the `SigningSubnetID` will be set as the subnet itself, but we shouldn't need any "destination chain" either way 👍

signature-aggregator/config/source_blockchain.go
```diff
@@ -0,0 +1,54 @@
...
+	SubnetID     string            `mapstructure:"subnet-id" json:"subnet-id"`
+	BlockchainID string            `mapstructure:"blockchain-id" json:"blockchain-id"` //nolint:lll
+	VM           string            `mapstructure:"vm" json:"vm"`
+	RPCEndpoint  baseCfg.APIConfig `mapstructure:"rpc-endpoint" json:"rpc-endpoint"` //nolint:lll
+	WSEndpoint   baseCfg.APIConfig `mapstructure:"ws-endpoint" json:"ws-endpoint"`   //nolint:lll
+
+	// convenience fields to access parsed data after initialization
+	subnetID     ids.ID
+	blockchainID ids.ID
+}
```
michaelkaplan13, line 24: Once ACP-118 is in place, will we need to configure `SourceBlockchains` at all?

I think we can hopefully get away without it entirely. Each request will specify the `sourceChainID` (used in the `AppRequest` [here](https://github.com/ava-labs/awm-relayer/blob/signature-aggregation-api/lib/aggregator.go#L140)), and the `signingSubnetID`, which determines which nodes the request should be sent to [here](https://github.com/ava-labs/awm-relayer/blob/signature-aggregation-api/lib/aggregator.go#L92). That should be sufficient for sending the proper request to the correct set of nodes, no matter the source blockchain.

If we're able to go this direction, I'd recommend that we start now by assuming all source subnets use subnet-evm's codec for signature requests. It won't work for the C-Chain in the immediate interim period, but then we can simply drop in the ACP-118 proto in place of the subnet-evm code for formatting the `AppRequest`s.

scripts/build.sh
```diff
@@ -42,7 +42,7 @@ fi
 if [[ $# -eq 1 ]]; then
     binary_path=$1
 elif [[ $# -eq 0 ]]; then
-    binary_path="build/awm-relayer"
+    binary_path=$relayer_path
```
michaelkaplan13, line 45: Did you mean to change this?

signature-aggregator/config/source_blockchain.go
iansuvak: Great point! Thank you. I assume that to do this we would need to be able to map a `sourceChainID` to `signingSubnetID` somehow. If not we can just make `signingSubnetID` a required field in the interface?

michaelkaplan13: > I assume that to do this we would need to be able to map a sourceChainID to signingSubnetID somehow. If not we can just make signingSubnetID a required field in the interface?

There's no mapping between `sourceChainID` and `signingSubnetID` (for instance, if the source chain is the P-chain, then the signing chain could be any subnet), but I think it'd be reasonable to either have `signingSubnetID` be a required param or to have the `signingSubnetID` default to the subnet containing the `sourceChainID` if it is not explicitly set.

scripts/build.sh
iansuvak: I didn't mean to commit this as part of the PR, but now that it snuck in. I was curious why do we set the variable in `constants.sh` file and then never use it.

feuGeneA (commented):
I feel like the API entry points (eg the request types and resource paths) should be verbs, eg "aggregate signatures".

Also, we should be consistent in saying "_by_ message ID" and "_by_ raw message". We're saying `ByID` in that request, but not the other request, and I think we can clarify it in the path.

feuGeneA (commented):
I wonder if we should integrate both utilities (relayer, signature-aggregator) into the repo's folder structure, instead of separating them out. So it would be in `cmd/relayer` and `cmd/signature-aggregator`, as described at https://github.com/golang-standards/project-layout/tree/master/cmd

signature-aggregator/README.md
```diff
@@ -0,0 +1,64 @@
+# Signature Aggregator
+
+This directory contains a lightweight stand-alone API for requesting signatures for a Warp message from subnet validators.
+It is also used by the relayer for gathering signatures when configured to use AppRequest instead of the WarpSignature RPC client.
+
+## Building
+
+To build the application run `signature-aggregator/scripts/build.sh` which will output the binary to `signature-aggregator/build/signature-aggregator` path by default.
```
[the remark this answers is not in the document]
iansuvak: Either way we should have all scripts on the top level `scripts` folder instead of nested by application

peers/app_request_network.go
```diff
@@ -42,7 +45,8 @@ type AppRequestNetwork struct {
 func NewNetwork(
 	logLevel logging.Level,
 	registerer prometheus.Registerer,
-	cfg *config.Config,
+	trackedSubnets set.Set[ids.ID],
+	cfg Config,
```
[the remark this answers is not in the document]
iansuvak: Yeah, Config here is an interface instead of a concrete type in order for both Signature Aggregator and the relayer to use it. The value that does get passed in is still technically a pointer that implements the smaller, new network.Config interface

signature-aggregator/aggregator/aggregator.go
```diff
@@ -0,0 +1,500 @@
...
+	}
+
+	var sigResponse msg.SignatureResponse
+	if _, err := msg.Codec.Unmarshal(appResponse.AppBytes, &sigResponse); err != nil {
+		s.logger.Error(
+			"Error unmarshaling signature response",
+			zap.Error(err),
+		)
+	}
+	signature := sigResponse.Signature
```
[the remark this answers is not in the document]
iansuvak: Yeah the check below confirms that we actually have a non-empty response and the relevant error will be logged.

config/config.go
```diff
@@ -246,3 +246,13 @@ func (c *Config) GetWarpQuorum(blockchainID ids.ID) (WarpQuorum, error) {
 	}
 	return WarpQuorum{}, errFailedToGetWarpQuorum
 }
+
+// Config implempents the peers.Config interface
```
minghinmatthewlam, line 250: nit: typos

config/config.go
```diff
@@ -246,3 +246,13 @@ func (c *Config) GetWarpQuorum(blockchainID ids.ID) (WarpQuorum, error) {
...
+func (c *Config) GetPChainAPI() *APIConfig {
+	return c.PChainAPI
+}
+
+// Config implempents the peers.Config interface
```
minghinmatthewlam, line 255: should we add a check that this type implements the peers.Config interface

main/main.go
```diff
@@ -133,11 +135,19 @@ func main() {
...
+		trackedSubnets.Add(sourceBlockchain.GetSubnetID())
+	}
+
 	network, err := peers.NewNetwork(
 		networkLogLevel,
 		prometheus.DefaultRegisterer,
+		trackedSubnets,
 		&cfg,
 	)
+	network.InitializeConnectionsAndCheckStake(&cfg)
```
minghinmatthewlam, line 149: does the config nee to be passed by reference

main/main.go
```diff
@@ -133,11 +135,19 @@ func main() {
 	if logLevel <= logging.Debug {
 		networkLogLevel = logLevel
 	}
+	var trackedSubnets set.Set[ids.ID]
```
minghinmatthewlam, line 138: can you add a comment why this is needs to be passed, insetad of previously in app request network

config/config.go
iansuvak: The compilation will fail if it doesn't since we pass this config into AppRequestsNetwork which requires peers.Config interface

peers/app_request_network.go
```diff
@@ -94,38 +94,47 @@ func NewNetwork(
...
-		Handler:         handler,
+		network:         testNetwork,
+		handler:         handler,
 		infoAPI:         infoAPI,
 		logger:          logger,
 		lock:            new(sync.Mutex),
 		validatorClient: validatorClient,
 	}
+	go logger.RecoverAndPanic(func() {
+		testNetwork.Dispatch()
```
minghinmatthewlam, line 108: do we need to watch the error returned from Dispatch?

peers/app_request_network.go
```diff
@@ -94,38 +94,47 @@ func NewNetwork(
...
 	// We return an error if we are unable to connect to sufficient stake on any of the subnets.
 	// Sufficient stake is determined by the Warp quora of the configured supported destinations,
 	// or if the subnet supports all destinations, by the quora of all configured destinations.
 	for _, sourceBlockchain := range cfg.SourceBlockchains {
 		if sourceBlockchain.GetSubnetID() == constants.PrimaryNetworkID {
-			if err := arNetwork.connectToPrimaryNetworkPeers(cfg, sourceBlockchain); err != nil {
-				return nil, err
+			if err := n.connectToPrimaryNetworkPeers(cfg, sourceBlockchain); err != nil {
+				return fmt.Errorf(
+					"Failed to connect to primary network peers: %w",
```
minghinmatthewlam, line 124: when returning error with `fmt.Errorf` we use lower case error mesasge, when printing/logging a new error we capitalize

main/main.go
iansuvak: Not strictly no, since it gets used as interface but from what I've seen we normally pass configs as references and I've done that in the past as well. Not a big deal in general but in applications where configs get passed around a lot doing this can save on allocs and garbage collection.

peers/app_request_network.go
```diff
@@ -245,6 +254,25 @@ func (n *AppRequestNetwork) ConnectToCanonicalValidators(subnetID ids.ID) (*Conn
 	}, nil
 }
 
+func (n *AppRequestNetwork) Send(
+	msg message.OutboundMessage,
+	nodeIDs set.Set[ids.NodeID],
+	subnetID ids.ID,
+	allower subnets.Allower,
```
minghinmatthewlam, line 261: do we need to set `allower` here? Or can we just set as the noop allower

relayer/application_relayer.go
```diff
@@ -246,8 +206,13 @@ func (r *ApplicationRelayer) relayMessage(
 
 	// sourceWarpSignatureClient is nil iff the source blockchain is configured to fetch signatures via AppRequest
 	if r.sourceWarpSignatureClient == nil {
+		// TODO: do we actually want to pass the pointer here or adapt the interface?
```
minghinmatthewlam, line 209: is this TODO decided on, and if not, should we handle it before the merge

peers/app_request_network.go
iansuvak: I'm open to either. In general I like to keep these wrapper interface methods as close as possible in the signature to the original but this is already broken here since we use the legacy VDR set instead of the new SendConfig. I'm fine with either:
:one: assuming our usage, since we only call send from here and do the conversion to SendConfig here and explicitly set the noopAllower here as well
:two: keep the interface of Send same as network.Send and do the conversion and allower setting in the application layer. 

I will defer on decision here to whatever the majority preference is

relayer/application_relayer.go
iansuvak: Sorry I removed this on a different branch. I will add it here as well. There is no longer interface since the both applications use the same aggregator and AppRequestNetwork.

main/main.go
```diff
@@ -428,14 +444,14 @@ func createApplicationRelayersForSourceChain(
 			logger,
 			metrics,
 			network,
-			messageCreator,
```
michaelkaplan13, line 431: Looks like `messageCreator` can be removed as a param for `createApplicationRelayersForSourceChain` and `createApplicationRelayers`

peers/config.go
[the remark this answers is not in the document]
cam-schultz: Alternatively, renaming to `AggregatorConfig` and moving it to `signature-aggregator/aggregator` would make more sense from a code organization perspective.

signature-aggregator/aggregator/aggregator.go
```diff
@@ -0,0 +1,500 @@
...
+		subnetIDsByBlockchainID: map[ids.ID]ids.ID{},
+		logger:                  logger,
+		messageCreator:          messageCreator,
+		currentRequestID:        atomic.Uint32{},
+	}
+	sa.currentRequestID.Store(rand.Uint32())
+	return &sa
+}
+
+func (s *SignatureAggregator) AggregateSignaturesAppRequest(
```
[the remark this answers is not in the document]
cam-schultz: Or even more simply `AggregateSignature`, since this type only does so via app request.

signature-aggregator/aggregator/aggregator.go
```diff
@@ -0,0 +1,507 @@
...
+	unsignedMessage *avalancheWarp.UnsignedMessage,
+	inputSigningSubnet ids.ID,
+	quorumPercentage uint64,
+) (*avalancheWarp.Message, error) {
+	requestID := s.currentRequestID.Add(1)
+
+	var signingSubnet ids.ID
+	var err error
+	// If signingSubnet is not set  we default to the subnet of the source blockchain
```
cam-schultz, line 85:
```suggestion
	// If signingSubnet is not set we default to the subnet of the source blockchain
```

signature-aggregator/aggregator/aggregator.go
```diff
@@ -0,0 +1,507 @@
...
+	errNotEnoughConnectedStake = errors.New("failed to connect to a threshold of stake")
+)
+
+type SignatureAggregator struct {
+	network                 *peers.AppRequestNetwork
+	subnetIDsByBlockchainID map[ids.ID]ids.ID
+	logger                  logging.Logger
+	messageCreator          message.Creator
+	currentRequestID        atomic.Uint32
+	mu                      sync.RWMutex
```
michaelkaplan13, line 57: Major nit: could we name this `requestIDLock`? I believe elsewhere in our codebases we always name `sync.Mutex`s something ending in "lock"

signature-aggregator/aggregator/aggregator.go
```diff
@@ -0,0 +1,507 @@
...
+	// Regardless of the response's relevance, call it's finished handler once this function returns
+	defer response.OnFinishedHandling()
+
+	// Check if this is an expected response.
+	m := response.Message()
+	rcvReqID, ok := message.GetRequestID(m)
+	if !ok {
+		// This should never occur, since inbound message validity is already checked by the inbound handler
+		s.logger.Error("Could not get requestID from message")
+		return nil, false, nil
```
cam-schultz, line 337: This may have already been here, but we should return an error in this case.

signature-aggregator/aggregator/aggregator.go
```diff
@@ -0,0 +1,507 @@
...
+				zap.String("sourceBlockchainID", unsignedMessage.SourceChainID.String()),
+				zap.String("warpMessageID", unsignedMessage.ID().String()),
+				zap.Error(err),
+			)
+			return nil, true, fmt.Errorf("%s: %w", msg, err)
+		}
+		return signedMsg, true, nil
+	}
+	// Not enough signatures, continue processing messages
+	return nil, true, nil
```
cam-schultz, line 413: We can combine these by declaring `signedMsg = nil` above this `if` block.

signature-aggregator/config/flags.go
cam-schultz, line 1: Some referenced to `awm-relayer` need to be renamed

signature-aggregator/api/api.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/logging"
+	"github.com/ava-labs/awm-relayer/signature-aggregator/aggregator"
+	"github.com/ava-labs/awm-relayer/types"
+	"github.com/ava-labs/awm-relayer/utils"
+	"go.uber.org/zap"
+)
+
+const (
+	RawMessageAPIPath = "/aggregate-signatures/by-raw-message"
```
cam-schultz, line 21: I'm a fan of single word paths, as long as they're descriptive enough within the context of the API. Here, I think `aggregate/message` or `signatures/message` would be appropriate. It's a matter of opinion of course, so lmk what you think.

signature-aggregator/api/api.go
```diff
@@ -0,0 +1,155 @@
...
+	defaultQuorumNum  = 67
+)
+
+// Defines a request interface for signature aggregation for a raw unsigned message.
+// Currently a copy of the `ManualWarpMessageRequest` struct in relay_message.go
```
cam-schultz, line 26: Can you add a TODO to combine these? We can probably have `awm-relayer` consume this package directly for manual Warp messages. We may want to add filtering for only supported subnets in that case though.

signature-aggregator/api/api.go
```diff
@@ -0,0 +1,155 @@
...
+type AggregateSignaturesByRawMsgRequest struct {
+	// Required. hex-encoded message, optionally prefixed with "0x".
+	UnsignedMessage string `json:"unsigned-message"`
+	// Optional hex or cb58 encoded signing subnet ID. If omitted will default to the subnetID of the source BlockChain
+	SigningSubnetID string `json:"signing-subnet-id"`
+	// Optional. Integer from 0 to 100 representing the percentage of the quorum that is required to sign the message
+	// defaults to 67 if omitted.
+	QuorumNum uint64 `json:"quorum-num"`
```
cam-schultz, line 34: I left a similar comment elsewhere, but "numerator" only makes sense with a corresponding "denominator". Rather than require the user to know that the implicit denominator is 100, a name like `QuorumPercentage` would be clearer.

signature-aggregator/README.md
```diff
@@ -0,0 +1,64 @@
+# Signature Aggregator
+
+This directory contains a lightweight stand-alone API for requesting signatures for a Warp message from subnet validators.
```
cam-schultz, line 3:
```suggestion
This directory contains a lightweight stand-alone API for requesting signatures for a Warp message from Subnet validators.
```

signature-aggregator/README.md
```diff
@@ -0,0 +1,64 @@
...
+It is also used by the relayer for gathering signatures when configured to use AppRequest instead of the WarpSignature RPC client.
```
cam-schultz, line 4:
```suggestion
It is also used by `awm-relayer` for gathering signatures when configured to use AppRequest instead of the WarpSignature RPC client.
```

Let's be specific

signature-aggregator/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Building
+
+To build the application run `scripts/build_signature_aggregator.sh` which will output the binary to `signature-aggregator/build/signature-aggregator` path by default.
+
+## Running
+
+To run the binary you must supply a config file via `./signature-aggregator --config-file`
+Currently required configurations are a small subset of the [relayer configuration](https://github.com/ava-labs/awm-relayer?tab=readme-ov-file#configuration).
```
cam-schultz, line 13:
```suggestion
Currently required configurations are a small subset of the [`awm-relayer` configuration](https://github.com/ava-labs/awm-relayer?tab=readme-ov-file#configuration).
```

signature-aggregator/README.md
```diff
@@ -0,0 +1,64 @@
...
+    "method": "eth_getTransactionReceipt",
+    "params": [
+        "<hex_encoded_transaction>"
+    ],
+    "id": 1
+}'
+```
+3. Search these logs for the `SendWarpMessage` Log event emitted from the Warp precompile address (`0x0200000000000000000000000000000000000005`)
+   The topic of the message will be `0x56600c567728a800c0aa927500f831cb451df66a7af570eb4df4dfbf4674887d` which is the output of`cast keccak "SendWarpMessage(address,bytes32,bytes)"`
+3. Use the data field of the log message found in step 2 and send it to the locally running service via curl.
```
cam-schultz, line 57:
```suggestion
4. Use the data field of the log message found in step 2 and send it to the locally running service via curl.
```

signature-aggregator/README.md
```diff
@@ -0,0 +1,64 @@
...
+```json
+{
+    "data": "",               // (string) hex-encoded unsigned message bytes to be signed
+    "justification": "",      // (string) hex-encoded bytes to supply to the validators as justification
+ .  "signing-subnet-id": "",  // (string) hex or cb58 encoded signing subnet ID. Defaults to source blockchain's subnet from data if omitted.
+    "quorum-num": 67          // (int) quorum percentage required to sign the message. Defaults to 67 if omitted
+}
+```
+
+## Sample workflow
```
cam-schultz, line 35: Can you add a note here that this example workflow will only work for recent messages due to validator churn?

cam-schultz (commented):
Generally looks pretty good. I left a handful of comments, but most of them are minor change requests, or asks to document future work.

signature-aggregator/api/api.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+type AggregateSignaturesResponse struct {
```
michaelkaplan13, line 37:
```suggestion
type AggregateSignatureResponse struct {
```

signature-aggregator/api/api.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+func HandleAggregateSignaturesByRawMsgRequest(
+	logger logging.Logger,
+	signatureAggregator *aggregator.SignatureAggregator,
+) {
+	http.Handle(RawMessageAPIPath, signatureAggregationAPIHandler(logger, signatureAggregator))
+}
+
+func writeJsonError(
```
michaelkaplan13, line 49: Nit: To follow golang conventions, this should be `writeJSONError`

signature-aggregator/api/api.go
```diff
@@ -0,0 +1,155 @@
...
+	w.Header().Set("Content-Type", "application/json")
+
+	w.Write(resp)
+	if err != nil {
+		logger.Error("Error writing error response", zap.Error(err))
+	}
+}
+
+func signatureAggregationAPIHandler(logger logging.Logger, aggregator *aggregator.SignatureAggregator) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
michaelkaplan13, line 70: For readability, can we define the handler function as a standalong func, and then pass it to `HandlerFunc` by reference?

signature-aggregator/config/config.go
```diff
@@ -0,0 +1,59 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package config
```
michaelkaplan13, line 4: Should this `config` package (or just its files) go under `api`? IIUC, they only apply to the aggregation API itself. That structure may better reflect that.

tests/utils/utils.go
```diff
@@ -102,6 +104,67 @@ func BuildAndRunRelayerExecutable(ctx context.Context, relayerConfigPath string)
 	}
 }
 
+func BuildAndRunSignatureAggregatorExecutable(ctx context.Context, configPath string) context.CancelFunc {
+	// Build the signature-aggregator binary
+	cmd := exec.Command("./scripts/build_signature_aggregator.sh")
```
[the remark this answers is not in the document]
iansuvak: We had it this way originally but moved it back after a discussion. I am ambivalent towards which one we end up doing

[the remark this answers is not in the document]
cam-schultz: We can defer it to the followup repo reorganization tickets.

signature-aggregator/config/config.go
iansuvak: We will have metrics port + cache memory limit here soon. If it was just metrics than I'd say move it but Cache memory limit is not API specific

scripts/build.sh
```diff
@@ -1,56 +1,12 @@
...
-if [[ $# -eq 1 ]]; then
-    binary_path=$1
-elif [[ $# -eq 0 ]]; then
-    binary_path="build/awm-relayer"
-else
-    echo "Invalid arguments to build awm-relayer. Requires zero (default location) or one argument to specify binary location."
-    exit 1
-fi
+source "$root"/scripts/versions.sh
+source "$root"/scripts/constants.sh
```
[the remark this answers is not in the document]
iansuvak: We shouldn't need to. I can remove these here but also think that there is some more potential cleanup in our bash scripts that we can follow up with  later

signature-aggregator/api/api.go
iansuvak: I agree that I'd like to update this. This is really a vestige of the old way where initial implementation supported both full message bytes as well as a `message id` so referencing `by-raw-message` was to distinguish from `by-id`. But we've since removed that. 

Since IIRC ACP-118 requires either the justification or message field to be non-zero I'd like to remove message entirely from the path. Any ideas about the most logical path here in that case `signatures/aggregate` ?

peers/app_request_network.go
```diff
@@ -42,7 +46,8 @@ type AppRequestNetwork struct {
 func NewNetwork(
 	logLevel logging.Level,
 	registerer prometheus.Registerer,
-	cfg *config.Config,
+	trackedSubnets set.Set[ids.ID],
```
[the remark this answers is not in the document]
iansuvak: Yeah we definitely need to keep it for now and I want to save bumping different versions of all our dependencies for a later PR. I am sort of considering keeping it for now as well though since there are other differences/places where `trackedSubnets` are used in the networking code. 

Specifically uptime tracking via ping messages. Not having `trackedSubnets` set makes peer nodes spew more logs about uptime tracking. This uptime tracking via ping messages  should go away soon as well but it hasn't yet.

peers/app_request_network.go
iansuvak: I think that `RecoverAndPanic` should cover our needs here but could be wrong

signature-aggregator/api/api.go
iansuvak: I'm just removing that comment for now since it's lying. Right now manual warp messages require specifying source address which is not part of this interface. I'm not convinced that we need that though, but either way that would be a breaking change on the relayer side.

config/config.go
minghinmatthewlam: makes sense and is a nit, but can we still add it? We have it in other places of the repo, and believe it's a standard go practice

main/main.go
minghinmatthewlam: can we remove the pass by ref here, want to avoid it being unintentionally altered

signature-aggregator/README.md
```diff
@@ -0,0 +1,64 @@
+# Signature Aggregator
+
+This directory contains a lightweight stand-alone API for requesting signatures for a Warp message from Subnet validators.
+It is also used by `awm-relayer` for gathering signatures when configured to use AppRequest instead of the WarpSignature RPC client.
```
minghinmatthewlam, line 4: nit: is WarpSignature meant to be one word

signature-aggregator/README.md
```diff
@@ -0,0 +1,64 @@
...
+}
+```
+
+## Sample workflow
+If you want to manually test a locally running service pointed to the Fuji testnet you can do so with the following steps.
+
+The basic request consists of sending just the `data` field containing the hex-encoded bytes of the full unsigned Warp message that the validators would be willing to sign. Here are the steps to obtain a sample valid unsigned Warp message bytes from the Fuji network.
+
+1. Find a valid on-chain `Receive Cross Chain Message` Transaction
+   This can be done by looking at the `Teleporter` [contract tracker](https://subnets-test.avax.network/c-chain/address/0x253b2784c75e510dD0fF1da844684a1aC0aa5fcf) on the Fuji-C
```
minghinmatthewlam, line 41: can link to Teleporter explorer

signature-aggregator/aggregator/aggregator.go
```diff
@@ -0,0 +1,507 @@
...
+}
+
+func (s *SignatureAggregator) GetSubnetID(blockchainID ids.ID) (ids.ID, error) {
+	s.mu.RLock()
+	subnetID, ok := s.subnetIDsByBlockchainID[blockchainID]
+	s.mu.RUnlock()
+	if ok {
+		return subnetID, nil
+	}
+	s.logger.Info("Signing subnet not found, requesting from PChain", zap.String("chainID", blockchainID.String()))
```
minghinmatthewlam, line 298:
```suggestion
	s.logger.Info("Signing subnet not found, requesting from PChain", zap.String("blockchainID", blockchainID.String()))
```

signature-aggregator/aggregator/aggregator.go
```diff
@@ -0,0 +1,507 @@
...
+	return subnetID, nil
+}
+
+func (s *SignatureAggregator) SetSubnetID(blockchainID ids.ID, subnetID ids.ID) {
+	s.mu.Lock()
+	s.subnetIDsByBlockchainID[blockchainID] = subnetID
+	s.mu.Unlock()
+}
+
+// Attempts to create a signed warp message from the accumulated responses.
```
minghinmatthewlam, line 313:
```suggestion
// Attempts to create a signed Warp message from the accumulated responses.
```

signature-aggregator/api/api.go
minghinmatthewlam: good with `signatures/aggregate`

signature-aggregator/api/api.go
```diff
@@ -0,0 +1,154 @@
...
+	"go.uber.org/zap"
+)
+
+const (
+	RawMessageAPIPath = "/aggregate-signatures/by-raw-message"
+	defaultQuorumNum  = 67
+)
+
+// Defines a request interface for signature aggregation for a raw unsigned message.
+type AggregateSignaturesByRawMsgRequest struct {
```
minghinmatthewlam, line 26: can this be just `AggregateSignatureRequest`

signature-aggregator/api/api.go
```diff
@@ -0,0 +1,154 @@
...
+	// Required. hex-encoded message, optionally prefixed with "0x".
+	UnsignedMessage string `json:"unsigned-message"`
+	// Optional hex or cb58 encoded signing subnet ID. If omitted will default to the subnetID of the source BlockChain
```
minghinmatthewlam, line 29:
```suggestion
	// Optional hex or cb58 encoded signing subnet ID. If omitted will default to the subnetID of the source blockchain
```

signature-aggregator/aggregator/aggregator.go
iansuvak: Actually I removed the requestIDLock since it's just uint32 and made it atomic instead. The lock above protects `subnetsByBlockchainID` but the point stands

utils/utils.go
```diff
@@ -54,6 +54,15 @@ func CheckStakeWeightExceedsThreshold(
 	return scaledTotalWeight.Cmp(scaledSigWeight) != 1
 }
 
+// Wrapper for CheckStakeWeightExceedThreshold with a quorumDen of 100.
+func CheckStakeWeightPercentageExceedsThreshold(
+	accumulatedSignatureWeight *big.Int,
+	totalWeight uint64,
+	quorumNumerator uint64,
```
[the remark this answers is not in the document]
iansuvak: Quroum denominator is assumed to be 100. You want the function to just be `stakeWeightPercentage`? It doesn't just calculate percentage it returns true if the percentage is above the numerator threshold which is the thing that we care about. 

Is the suggestion to rename it `stakeWeightPercentage` and have it return the percentage and then check in the calling code that it's >= numerator? Or still do it here?

signature-aggregator/api/api.go
feuGeneA: I wrote the path that's in place here right now, but I'm liking these discussion directions more.

Some more food for thought:
* Should the path be a noun or a verb? I would think a noun. 
* Should it be `signature` (singular) or `signatures` (plural)? Seems like plural makes sense if we're using "aggregate" as a verb ("aggregate the signatures for me") but singular makes sense if we're using "aggregate" as a noun ("get me the aggregate signature").

My current best suggestion is just `/aggregate`. That leaves it open ended so that we don't have to answer any of those questions 😅 but if we want to add another word or two to the path then we should consider noun/verb/singular/plural.

feuGeneA: Friendly reminder to update the path references in the README as well :)

utils/utils.go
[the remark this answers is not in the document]
cam-schultz: Sorry, my initial comment wasn't worded very clearly. I suggest we rename `quorumNumerator` to `stakeWeightPercentage` or `stakeWeightPercentageThreshold` to remove any assumptions about the implicit denominator.

scripts/build.sh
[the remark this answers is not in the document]
cam-schultz: Definitely. I do think a first pass to remove redundancies should be done in this PR.

peers/app_request_network.go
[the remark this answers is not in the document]
cam-schultz: Sounds good. A more general comment, there's a lot of miscellania that's rightly being deferred in this PR. Do you mind enumerating them in a followup ticket so that they don't get lost in the chaff?

signature-aggregator/api/api.go
cam-schultz: If there's no intention to add different subpaths, I agree with @feuGeneA that a single path level is sufficient. I like `aggregate`, though if that's not descriptive enough, the original `aggregate-signatures` (without `by-raw-message`) is good too.

signature-aggregator/api/api.go
cam-schultz: Cool, sounds good

signature-aggregator/api/api.go
iansuvak: +1 for me for `aggregate-signatures`

signature-aggregator/aggregator/aggregator.go
```diff
@@ -0,0 +1,507 @@
...
+// Returns a non-nil Warp message if [accumulatedSignatureWeight] exceeds the signature verification threshold.
+// Returns false in the second return parameter if the app response is not relevant to the current signature
```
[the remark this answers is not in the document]
iansuvak: if is correct here I believe

utils/utils.go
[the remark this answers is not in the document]
iansuvak: Thanks for clarifying. Yeah I agree and will make the change

config/config.go
iansuvak: Actually can't do it here since it would introduce a circular dependency. I'm happy to enter an issue for cleaning up some of the config handling between both this and `awm-relayer` but I'd like to punt on it until a future PR

main/main.go
iansuvak: FWIW configs are being passed by reference everywhere in this codebase. As mentioned in previous comment we can clean this up but I'd like to punt on this until later

peers/app_request_network.go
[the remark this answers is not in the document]
iansuvak: Yup I will enter those issues and link them in the description for easy reference before we are ok to merge this one.

signature-aggregator/README.md
iansuvak: Sort of since it was referring to a particular client but that's not it's proper name so split it up for now

signature-aggregator/README.md
iansuvak: I looked through Teleporter explorer originally but it's more clicks to find `Receive Cross Chain Message` transactions via Teleporter explorer and you would still end up having to click to a destination Teleporter contract so this is a shortcut

signature-aggregator/api/api.go
iansuvak: I'm not actually sure what do you mean here. The `HandlerFunc` needs it's signature to be `w http.ResponseWriter, r *http.Request`and the anonymous function inside that we are returning needs access to `logger` and `aggregator` as well.  Is there an example doing what you mean?

signature-aggregator/aggregator/aggregator.go
iansuvak: This shouldn't happen but if it did for a single request, returning an error here instead of nil would short-circuit the entire aggregation. Do we still want to make the change?

signature-aggregator/aggregator/aggregator.go
iansuvak: For me it's more readable to keep the control flow as is with the comment. Makes it clear that the happy case terminates inside the if statement loop and that if you are outside you didn't get enough signatures. It's also not reducing LOC since it would be trading adding `signedMsg=nil` above for the return value. 

If you strongly prefer it I will make the change though

tests/utils/utils.go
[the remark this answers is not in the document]
iansuvak: Added to #409

tests/utils/utils.go
```diff
@@ -102,6 +104,67 @@ func BuildAndRunRelayerExecutable(ctx context.Context, relayerConfigPath string)
...
+		err := signatureAggregatorCmd.Wait()
+		// Context cancellation is the only expected way for the process to exit, otherwise panic
+		if !errors.Is(signatureAggregatorCtx.Err(), context.Canceled) {
+			panic(fmt.Errorf("signature-aggregator exited abnormally: %w", err))
+		}
+	}()
+	return func() {
+		signatureAggregatorCancelFunc()
+		<-signatureAggregatorCtx.Done()
+	}
```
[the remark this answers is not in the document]
iansuvak: Added to #409

signature-aggregator/api/api.go
michaelkaplan13: Sorry I was confusing this with how we've set it up in other places using a `struct` to hold the `logger`/`aggregator` as members, and named handler `func`s. I don't think it's worth doing that here now though.

signature-aggregator/api/api.go
```diff
@@ -0,0 +1,154 @@
...
+	SigningSubnetID string `json:"signing-subnet-id"`
+	// Optional. Integer from 0 to 100 representing the percentage of the quorum that is required to sign the message
+	// defaults to 67 if omitted.
+	QuorumPercentage uint64 `json:"quorum-percentage"`
+}
+
+type AggregateSignatureResponse struct {
+	// hex encoding of the signature
+	SignedMessage string `json:"signed-message"`
+}
```
cam-schultz, line 39: We should consider returning the P-Chain height corresponding to the signature's canonical validator bit vector. Without it, there's an implicit "expiration" on responses from the API. With it, the caller would be better able to implement checks against the current state of the P-Chain before sending a tx to deliver the Warp message.

iansuvak: I'm happy to add this to follow up. Since we are already changing interfaces there for ACP-118. But I do think that the implicit P-chain height is current height and the general workflow is request it and send it immediately. Failure in that case is extremely unlikely but a simple retry of the full flow should be an easy fix.

signature-aggregator/README.md
```diff
@@ -22,19 +22,21 @@ Sample config that can be used for local testing is `signature-aggregator/sample
...
  .  "signing-subnet-id": "",  // (string) hex or cb58 encoded signing subnet ID. Defaults to source blockchain's subnet from data if omitted.
-    "quorum-num": 67          // (int) quorum percentage required to sign the message. Defaults to 67 if omitted
+    "quorum-percentage": 67  // (int) quorum percentage required to sign the message. Defaults to 67 if omitted
 }
 ```
 
 ## Sample workflow
 If you want to manually test a locally running service pointed to the Fuji testnet you can do so with the following steps.
 
+Note that this might fail for older messages if there has been enough validator churn, and less then threhold weight of stake of validators have seen the message when it originated. In this case try picking a more recent message.
```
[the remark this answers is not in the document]
cam-schultz: Expanding this to 1) explain that the Warp protocol does not support resigning, but 2) protocols on top of Warp such as Teleporter do (and link to the [relevant docs](https://github.com/ava-labs/teleporter/pull/487/files#diff-633988a43ea67f810d01288f07e014281067fe8129b594e93e550fa49f07e580R77))

cam-schultz (commented):
A couple of final nits, otherwise LGTM

config/config.go
minghinmatthewlam: good with deferring, can we create an issue to track and ref here

main/main.go
minghinmatthewlam: same as above, good on punting but can you create an issue to track

signature-aggregator/aggregator/aggregator.go
```diff
@@ -0,0 +1,506 @@
...
+		)
+		return nil, true, nil
+	}
+
+	// As soon as the signatures exceed the stake weight threshold we try to aggregate and send the transaction.
+	if utils.CheckStakeWeightPercentageExceedsThreshold(
+		accumulatedSignatureWeight,
+		connectedValidators.TotalValidatorWeight,
+		quorumPercentage,
+	) {
```
geoff-vball, line 379: Can we invert this if clause and return early?

signature-aggregator/aggregator/aggregator.go
geoff-vball: Yeah, I think I prefer explicitly returning nil

signature-aggregator/aggregator/aggregator.go
```diff
@@ -0,0 +1,506 @@
...
+	}
+
+	if !bls.Verify(pubKey, sig, unsignedMessage.Bytes()) {
+		s.logger.Debug(
+			"Failed verification for signature",
+		)
+		return blsSignatureBuf{}, false
+	}
+
+	return signature, true
```
geoff-vball, line 475: Would it make more sense to unmarshal the signature here instead of in `aggregateSignatures`?

signature-aggregator/README.md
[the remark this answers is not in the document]
iansuvak: I just left a comment on your PR but I will link to the doc section once they are merged in main.  I don't really want to over-elaborate here beyond linking since this is just a sample testing flow to make sure that the service works as expected.

signature-aggregator/api/api.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/ava-labs/avalanchego/utils/logging"
+	"github.com/ava-labs/awm-relayer/signature-aggregator/aggregator"
+	"github.com/ava-labs/awm-relayer/types"
+	"github.com/ava-labs/awm-relayer/utils"
+	"go.uber.org/zap"
+)
+
+const (
+	APIPath          = "/aggregate-signatures"
+	QuorumPercentage = 67
```
geoff-vball, line 22: Could we name this `DefaultQuorumPercentage`?

signature-aggregator/main/main.go
```diff
@@ -0,0 +1,119 @@
...
+	"go.uber.org/zap"
+)
+
+var version = "v0.0.0-dev"
+
+func main() {
+	fs := config.BuildFlagSet()
+	if err := fs.Parse(os.Args[1:]); err != nil {
+		fmt.Println("Failed to parse flags: %w", err)
+		os.Exit(1)
```
geoff-vball, line 30: Is there a reason this is `os.Exit(1)` instead of `panic`?

geoff-vball: We might want to `DisplayUsageText()` in this case, and also in any other panic case here that could be helped with the usage text.

signature-aggregator/aggregator/aggregator.go
```diff
@@ -0,0 +1,506 @@
...
+					zap.String("nodeID", nodeID.String()),
+					zap.Error(err),
+				)
+				responsesExpected--
+			}
+		}
+
+		responseCount := 0
+		if responsesExpected > 0 {
+			for response := range responseChan {
```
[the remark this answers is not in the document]
iansuvak: We do via a timeout mechanism which sends a dummy AppError response

minghinmatthewlam (commented):
mostly LGTM my end, just leftover comments Geoff pointed out

signature-aggregator/aggregator/aggregator.go
iansuvak: Potentially but we have a follow up ticket to re-evaluate this logic: Referenced this in #388

main/main.go
iansuvak: Updated: https://github.com/ava-labs/awm-relayer/issues/409

geoff-vball (approved):
🚀

cam-schultz (approved):

PR #370, opened 2024-07-17
Add API monitoring latency (info, validator)

### LINKED ISSUE #187: Add API monitoring latency
**Context and scope**
We should measure the latency of node API calls to help us determine scaling strategies. We should measure latency of calls to subnet nodes as well as to the P-chain API node.

We currently only measure the end-to-end latency of creating a signed message.

won-js:
## Why this should be merged
Resolve #187 

## How this works
Added latency measurement logic for the info and validator APIs.
I created a new Metrics type
I have also made the necessary modifications.

- registerer should change to prometheus.DefaultRegisterer [here](https://github.com/ava-labs/awm-relayer/blob/main/peers/app_request_network.go#L58C52-L58C62)
- we should pass in registerer to the constructor [here](https://github.com/ava-labs/awm-relayer/blob/main/main/main.go#L135)

## How is this documented


peers/app_request_netrwork_metrics.go
```diff
@@ -0,0 +1,52 @@
...
+	infoAPIBaseURL   string
+	pChainAPIBaseURL string
+
+	infoAPICallLatencyMS   *prometheus.GaugeVec
+	pChainAPICallLatencyMS *prometheus.GaugeVec
+}
+
+func NewAppRequestNetworkMetrics(cfg *config.Config, registerer prometheus.Registerer) (*AppRequestNetworkMetrics, error) {
+	infoAPICallLatencyMS := prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
```
najeal, line 23: Shall it be an histogram ?

won-js: That's a good idea
i've made the changes.

peers/app_request_netrwork_metrics.go
```diff
@@ -0,0 +1,54 @@
...
+	infoAPICallLatencyMS   *prometheus.HistogramVec
+	pChainAPICallLatencyMS *prometheus.HistogramVec
+}
+
+func NewAppRequestNetworkMetrics(cfg *config.Config, registerer prometheus.Registerer) (*AppRequestNetworkMetrics, error) {
+	infoAPICallLatencyMS := prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "info_api_call_latency_ms",
+			Help:    "Latency of calling info api in milliseconds",
+			Buckets: prometheus.LinearBuckets(10, 10, 10),
```
cam-schultz, line 26: Rather than linearly spaced buckets, exponentially distributed buckets would be more useful for identifying latency spikes. I think a good distribution to use would be to measure between ~0.1s and ~10s latency (a reasonable timeout), with bucket doubling the previous bucket's range. This can be done with [ExponentialBucketsRange](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#ExponentialBuckets).

cam-schultz (commented):
Thanks for putting this together! At a high level, this all looks good. I have a handful of minor comments and asks.

peers/app_request_network.go
```diff
@@ -191,12 +191,15 @@ func (n *AppRequestNetwork) ConnectPeers(nodeIDs set.Set[ids.NodeID]) set.Set[id
...
+		n.setInfoAPICallLatencyMS(float64(time.Since(startInfoAPICall).Milliseconds()))
+
 		n.logger.Error(
 			"Failed to get API Node ID",
 			zap.Error(err),
 		)
-		n.setInfoAPICallLatencyMS(float64(time.Since(startInfoAPICall).Milliseconds()))
 	} else if nodeIDs.Contains(apiNodeID) {
+		n.setInfoAPICallLatencyMS(float64(time.Since(startInfoAPICall).Milliseconds()))
+
```
cam-schultz, line 202: This control flow is a bit hard to follow. I'd recomment moving the call to `infoAPI.GetNodeID` above the `if` statement to clean it up:
```
startInfoAPICall = time.Now()
apiNodeID, _, err := n.infoAPI.GetNodeID(context.Background())
n.setInfoAPICallLatencyMS(float64(time.Since(startInfoAPICall).Milliseconds()))
if err != nil {
  ...
} else if nodeIDs.Contains(apiNodeID) {
  ...
} 
```

Similarly for the below call to `infoAPI.GetNodeIP`, let's move it out of the `if` statement so that we can put the latency measurement right next to the call:
```
startInfoAPICall = time.Now()
apiNodeIP, err := n.infoAPI.GetNodeIP(context.Background())
n.setInfoAPICallLatencyMS(float64(time.Since(startInfoAPICall).Milliseconds()))
if err != nil {
 ...
}
```

peers/app_request_network.go
```diff
@@ -55,7 +55,7 @@ func NewNetwork(
 		),
 	)
 
-	metrics, err := NewAppRequestNetworkMetrics(cfg, registerer)
+	metrics, err := newAppRequestNetworkMetrics(registerer)
```
cam-schultz, line 58: I can't leave a comment at the exact line, but let's also measure the latency of the call to `infoAPI.GetNetworkID` in the constructor.

peers/app_request_netrwork_metrics.go
cam-schultz, line 1: Can you please add a copyright notice to the top of this file?
```
// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
// See the file LICENSE for licensing terms.
```

peers/app_request_netrwork_metrics.go
```diff
@@ -0,0 +1,46 @@
+package peers
+
+import (
+	"errors"
+	"github.com/prometheus/client_golang/prometheus"
+)
```
cam-schultz, line 6:
```suggestion
import (
	"errors"
	
	"github.com/prometheus/client_golang/prometheus"
)
```
I believe we have to add a line separation stdlib from third party imports for linting to pass: https://stackoverflow.com/questions/59960394/file-is-not-goimports-ed-with-local-somepath

won-js: Thank you for the review.
I have rebased due to conflicts with the upstream code.
I have also added latency measurement code for the infoAPI in the constructor.

peers/app_request_network.go
```diff
@@ -145,8 +156,10 @@ func (n *AppRequestNetwork) ConnectPeers(nodeIDs set.Set[ids.NodeID]) set.Set[id
 	// through connectedPeers for already tracked peers, just iterate through the full list,
 	// re-adding connections to already tracked peers.
 
+	startInfoAPICall := time.Now()
 	// Get the list of peers
 	peers, err := n.infoAPI.Peers(context.Background())
+	n.metrics.setInfoAPICallLatencyMS(float64(time.Since(startInfoAPICall).Milliseconds()))
```
geoff-vball, line 162: Could we create a receiver function `func (n *AppRequestNetwork) setInfoAPICallLatencyMS(t time.Time)` to extract all this conversion?

geoff-vball:
Hey @won-js do you mind solving the merge conflicts in this PR and address the last comment @cam-schultz made?

geoff-vball (approved):
LGTM 👍

peers/app_request_network.go
cam-schultz: Looks like this comment is not resolved, but on second thought I think it's fine to skip measuring this call since it's only called once at startup.

cam-schultz (approved):

cam-schultz:
@won-js this looks good to go, just need to merge in `main`.

PR #373, opened 2024-07-19
adapt to teleporter v1.0.3

feuGeneA:
## How is this documented


geoff-vball (approved):

relayer/application_relayer.go
```diff
@@ -471,7 +472,14 @@ func (r *ApplicationRelayer) createSignedMessageAppRequest(
 		}
 		responseChan := r.network.Handler.RegisterRequestID(requestID, vdrSet.Len())
 
-		sentTo := r.network.Network.Send(outMsg, vdrSet, r.sourceBlockchain.GetSubnetID(), subnets.NoOpAllower)
+		sentTo := r.network.Network.Send(
+			outMsg,
+			// TODO: consider whether we need to specify the other
```
cam-schultz, line 477: My understanding is that the other fields are used to specify the number of additional nodes to sample. Since we are making requests to specific nodes explicitly, I don't think additional sampling is necessary.

peers/info_client.go
```diff
@@ -48,7 +48,8 @@ func (i *InfoAPI) GetNodeID(ctx context.Context) (ids.NodeID, *signer.ProofOfPos
 }
 
 func (i *InfoAPI) GetNodeIP(ctx context.Context) (string, error) {
-	return i.client.GetNodeIP(ctx, i.options...)
+	addrPort, err := i.client.GetNodeIP(ctx, i.options...)
+	return addrPort.Addr().String(), err
```
cam-schultz, line 52: Let's have `InfoAPI` continue to be a passthrough to `InfoClient`, and have `GetNodeIP` have the same return values as `InfoClient`'s method.

geoff-vball:
I made an issue for this here https://github.com/ava-labs/awm-relayer/issues/376

relayer/application_relayer.go
feuGeneA: thanks for the input, removed the TODO in https://github.com/ava-labs/awm-relayer/pull/373/commits/e0794af65bdfa5c322347a8370272690cd00dbc4

peers/info_client.go
feuGeneA: addressed in https://github.com/ava-labs/awm-relayer/pull/373/commits/641e5326b88389f0647a4d74781366343c8fb37a

peers/info_client.go
```diff
@@ -47,7 +48,7 @@ func (i *InfoAPI) GetNodeID(ctx context.Context) (ids.NodeID, *signer.ProofOfPos
 	return i.client.GetNodeID(ctx, i.options...)
 }
 
-func (i *InfoAPI) GetNodeIP(ctx context.Context) (string, error) {
+func (i *InfoAPI) GetNodeIP(ctx context.Context) (netip.AddrPort, error) {
```
geoff-vball, line 51: Are we able to change the name of this function?

cam-schultz: It's not part of an interface, but our `InfoAPI` type is written as an extension to `InfoClient` that embeds the rpc options in the object, rather than leaving it to the caller. As such, matching the `InfoClient` interface makes the code more readable.

What's the motivation for changing the function name?

geoff-vball: The function is named `GetNodeIP` but we're returning both the IP and the port

cam-schultz: My preference would be to keep parity with the underlying client interface, even if the name could be improved. That greatly helps code discovery when referencing existing usages as examples.

iansuvak (approved):
Excited to have this merged and speed up the e2e tests!

cam-schultz (approved):

geoff-vball (approved):

PR #387, opened 2024-07-26
e2e tests: use fewer nodes

feuGeneA:
## Why this should be merged
Because it reduced my test run time from 5m10s to 4m45s, an 8% reduction.

## How this was tested
By running the existing tests.


cam-schultz (approved):
LGTM. Some context, the condition I want to make sure we still hit is that we _don't_ reach the stake threshold on the first received signature, that way we exercise the full signature request control flow before returning the signature. 2 validators is sufficient for this (and is in fact the minimum required number)

geoff-vball (approved):

PR #393, opened 2024-07-30
Signature aggregation API fixes for small items

### LINKED ISSUE #381: Small Warp Signature API Standardization Items
**Context and scope**
#361 introduces the scaffolding for fetching an arbitrary aggregate signature via API request. There a handful of small clean up items that will bring its initial implementation more inline with our other API standards:
- [ ] Raw message should be hexadecimal encoded in request (with optional 0x prefix). We generally have not used base64 encoding anywhere.
- [ ] Response body fields should also be hexadecimal encoded
- [ ] Error responses can be descriptive strings, but shouldn't expose internal errors. Error responses should also be HTTP 200s, with JSON response body that includes the error message.
- [ ] Avoid pointer usage for `uint`s / `ids.ID`s where not strictly necessary
- [ ] Responses should have `application/JSON` header
- [ ] Consider having the response only include the aggregate signature and bit vector rather than the full signed message.

feuGeneA:
## Why this should be merged
Because it fixes #381 

## How this was tested
With the tests in the code base

### DIFF
--- a/signature-aggregator/api/api.go
+++ b/signature-aggregator/api/api.go
@@ -54,13 +54,35 @@ func HandleSignatureAggregationRawRequest(logger logging.Logger, signatureAggreg
 	http.Handle(RawMessageAPIPath, signatureAggregationAPIHandler(logger, signatureAggregator))
 }
 
+func writeJsonError(
+	logger logging.Logger,
+	w http.ResponseWriter,
+	errorMsg string,
+) {
+	resp, err := json.Marshal(struct{ error string }{error: errorMsg})
+	if err != nil {
+		logger.Error(
+			"Error marshalling JSON error response",
+			zap.Error(err),
+		)
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+
+	w.Write(resp)
+	if err != nil {
+		logger.Error("Error writing error response", zap.Error(err))
+	}
+}
+
 func signatureAggregationAPIHandler(logger logging.Logger, aggregator *aggregator.SignatureAggregator) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		var req SignatureAggregationRawRequest
 		err := json.NewDecoder(r.Body).Decode(&req)
 		if err != nil {
-			logger.Warn("Could not decode request body")
-			http.Error(w, err.Error(), http.StatusBadRequest)
+			msg := "Could not decode request body"
+			logger.Warn(msg, zap.Error(err))
+			writeJsonError(logger, w, msg)
 			return
 		}
 		var decodedMessage []byte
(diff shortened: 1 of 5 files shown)


signature-aggregator/api/api.go
```diff
@@ -34,75 +36,128 @@ type SignatureAggregationByIDRequest struct {
...
 			return
 		}
 		var quorumNum uint64
-		if req.QuorumNum == nil {
+		if req.QuorumNum == 0 {
 			quorumNum = defaultQuorumNum
-		} else if *req.QuorumNum >= 0 || *req.QuorumNum > 100 {
-			logger.Warn("Invalid quorum number", zap.Uint64("quorum-num", *req.QuorumNum))
-			http.Error(w, "invalid quorum number", http.StatusBadRequest)
+		} else if req.QuorumNum >= 0 || req.QuorumNum > 100 {
```
iansuvak, line 112: This was my initial mistake but this if statement is wrong and  will return true for any `QuorumNum` supplied since any uint is >=0. It was originally meant to be <=0 but this is uint64 so that shouldn't be possible. Additionally since it's defined as 0-100 we should be safe to use uint8 here?

geoff-vball, line 112:
```suggestion
		} else if req.QuorumNum <= 0 || req.QuorumNum > 100 {
```

signature-aggregator/api/api.go
```diff
@@ -34,75 +36,128 @@ type SignatureAggregationByIDRequest struct {
...
+			msg := "Invalid quorum number"
+			logger.Warn(msg, zap.Uint64("quorum-num", req.QuorumNum))
+			writeJsonError(logger, w, msg)
 			return
 		} else {
-			quorumNum = *req.QuorumNum
+			quorumNum = req.QuorumNum
 		}
```
geoff-vball, line 119: `else` isn't needed here if the `if` returns early

signature-aggregator/api/api.go
iansuvak: The first portion shouldn't be necessary since it's uint right now either

my suggestion would be
```suggestion
		} else if req.QuorumNum > 100 {
```

signature-aggregator/api/api.go
```diff
@@ -34,75 +36,128 @@ type SignatureAggregationByIDRequest struct {
...
-	// Required. Unsigned base64 encoded message bytes.
-	UnsignedMessageBytes []byte `json:"unsigned-message-bytes"`
+	// Required. hex-encoded message, optionally prefixed with "0x".
+	UnsignedMessage string `json:"unsigned-message"`
 	// Optional hex or cb58 encoded signing subnet ID. If omitted will default to the subnetID of the source BlockChain
 	SigningSubnetID string `json:"signing-subnet-id"`
 	// Optional. Integer from 0 to 100 representing the percentage of the quorum that is required to sign the message
 	// defaults to 67 if omitted.
-	QuorumNum *uint64 `json:"quorum-num"`
+	QuorumNum uint64 `json:"quorum-num"`
```
geoff-vball, line 45: Does the code still default to 67 if omitted now?

feuGeneA: Yes the default is applied in that quorum check if we were discussing elsewhere here.

signature-aggregator/api/api.go
feuGeneA: addressed in https://github.com/ava-labs/awm-relayer/pull/393/commits/da09584537decb4de155eb37c00ad64af61931a4

feuGeneA: addressed in https://github.com/ava-labs/awm-relayer/pull/393/commits/da09584537decb4de155eb37c00ad64af61931a4

signature-aggregator/api/api.go
feuGeneA: addressed in addressed in https://github.com/ava-labs/awm-relayer/pull/393/commits/da09584537decb4de155eb37c00ad64af61931a4

geoff-vball (approved):

PR #401, opened 2024-08-01
Update teleporter to v1.0.4

geoff-vball:
## Why this should be merged
Should fix test failures introduced in teleporter `v1.0.3`.


scripts/build.sh
```diff
@@ -39,8 +39,6 @@ if version_lt "$(go_version)" "$go_version_minimum"; then
     exit 1
 fi
 
-scripts/protobuf_codegen.sh
```
geoff-vball, line 42: This was causing some issues locally. We shouldn't need to install buf to run anything, since the files it generates are checked in.

iansuvak (dismissed):

cam-schultz (approved):

scripts/build.sh
feuGeneA: If we remove this then we should remove the README instruction to install buf, under `Building`.

Also though, I think the README should also mention/explain that we have a codgen stage, and point to the scripts we use for that.

iansuvak (approved):

README.md
```diff
@@ -410,3 +408,19 @@ The E2E tests use the `TeleporterMessenger` contract deployment transaction spec
 ```bash
 go generate ./...
 ```
+
+### Generate Protobuf Files
+
+[buf](github.com/bufbuild/buf/) is used to generate protobuf definitions for communication with the decider module. If you change any of the protobuf definitions you will have to regenerate the `.go` files. To generate these files, run the following command at the root of the project:
```
cam-schultz, line 414:
```suggestion
[buf](github.com/bufbuild/buf/) is used to generate protobuf definitions for communication with the [Decider service](https://github.com/ava-labs/awm-relayer/blob/main/proto/decider/decider.proto). If you change any of the protobuf definitions you will have to regenerate the `.go` files. To generate these files, run the following command at the root of the project:
```

README.md
```diff
@@ -410,3 +408,19 @@ The E2E tests use the `TeleporterMessenger` contract deployment transaction spec
 ```bash
 go generate ./...
 ```
+
+### Generate Protobuf Files
+
+[buf](github.com/bufbuild/buf/) is used to generate protobuf definitions for communication with the [Decider service](https://github.com/ava-labs/awm-relayer/blob/main/proto/decider/decider.proto). If you change any of the protobuf definitions you will have to regenerate the `.go` files. To generate these files, run the following command at the root of the project:
```
feuGeneA, line 414: Should we mention here that you need to manually pre-install `buf`? Or I wonder if we should modify the script so that it installs it for you, like we do with `abigen`.

cam-schultz: We should at minimum add this to the `Requirements` section in the top-level README

geoff-vball: Added this to the Requirements section. Installing the correct version of `buf` seems quite tricky and platform dependent, so I think sending people to the `buf` github page is the best course of action

cam-schultz (approved):

iansuvak (approved):

PR #404, opened 2024-08-01
workflows: add buf linter for proto files

### LINKED ISSUE #402: Add protobuf lint job
**Context and scope**
We should check in CI that any protobuf definition changes have been properly linted. Example: https://github.com/ava-labs/avalanchego/blob/master/.github/workflows/ci.yml#L129

najeal:
## Why this should be merged

Closes #402 
This adds a new job using bug linter on proto files

## How this works

It uses the `bufbuild` github actions to setup buf cli and run linter on `proto` directory.

## How this was tested

it has been tested locally using `act`.
-> with current proto files:  No lint errors were found.
-> introducing an error in decider.proto, it logs:
```
::error file=proto/decider/decider.proto,line=15,col=32::syntax error: unexpected ';', expecting int literal
::error::buf found 1 lint failures.
```

## How is this documented

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -12,6 +12,17 @@ on:
       - "**"
 
 jobs:
+  buf-lint:
+    name: Protobuf Lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: bufbuild/buf-setup-action@v1.31.0
+        with:
+          github_token: ${{ github.token }}
+      - uses: bufbuild/buf-lint-action@v1
+        with:
+          input: "proto"
   golangci:
     runs-on: ubuntu-22.04
     timeout-minutes: 10
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -14,12 +14,10 @@ on:
 jobs:
   buf-lint:
     name: Protobuf Lint
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
-      - uses: bufbuild/buf-setup-action@v1.31.0
-        with:
-          github_token: ${{ github.token }}
+      - uses: bufbuild/buf-setup-action@v1.35.1
       - uses: bufbuild/buf-lint-action@v1
         with:
           input: "proto"
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -1,7 +1,7 @@
 # Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
 # See the file LICENSE for licensing terms.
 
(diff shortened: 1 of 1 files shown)


.github/workflows/linter.yml
```diff
@@ -12,6 +12,17 @@ on:
...
 
 jobs:
+  buf-lint:
+    name: Protobuf Lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: bufbuild/buf-setup-action@v1.31.0
+        with:
+          github_token: ${{ github.token }}
```
cam-schultz, line 22:
```suggestion
```
The github token is not needed

.github/workflows/linter.yml
```diff
@@ -12,6 +12,17 @@ on:
       - "**"
 
 jobs:
+  buf-lint:
+    name: Protobuf Lint
+    runs-on: ubuntu-latest
```
cam-schultz, line 17: Let's use 22.04 for consistency
```suggestion
    runs-on: ubuntu-22.04
```

.github/workflows/linter.yml
```diff
@@ -12,6 +12,17 @@ on:
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: bufbuild/buf-setup-action@v1.31.0
```
cam-schultz, line 20:
```suggestion
      - uses: bufbuild/buf-setup-action@v1.35.1
```

cam-schultz (commented):
Thanks for tackling this! A few small comments, but then should be good to go.

feuGeneA:
There's already a buf lint invocation within `scripts/lint.sh`, so with this change we're effectively doing it twice. I'm not completely opposed to that, if there's some added benefit to it being in a GitHub workflow.

In any case, I'd prefer to keep the one in `scripts/lint.sh` so that there's an easy entry point for local development, and I'd prefer to keep exercising that same entry point via CI, as a validation of the local development workflow.

cam-schultz:
Closing, as we already lint the proto files in lint.sh

cam-schultz:
> There's already a buf lint invocation within `scripts/lint.sh`, so with this change we're effectively doing it twice. I'm not completely opposed to that, if there's some added benefit to it being in a GitHub workflow.
> 
> In any case, I'd prefer to keep the one in `scripts/lint.sh` so that there's an easy entry point for local development, and I'd prefer to keep exercising that same entry point via CI, as a validation of the local development workflow.

I prefer the current approach so that a developer can lint the entire codebase locally outside of CI.

najeal:
One of the solution to avoid running two times the same linter:
We keep the script lint.sh for development environment and cleanup the go-lint job to run golang lint only.
This way we have two different jobs using standard github actions easily testable locally with act : proto-lint & go-lint.

PR #405, opened 2024-08-01
config: check tracked subnets does not exceed the limit

### LINKED ISSUE #399: Enforce a limit of 16 tracked subnets
**Context and scope**
Avalanchego only allows a peer to track 16 subnets, otherwise it will drop outbound requests. To prevent this scenario, we should validate that the configuration supplies at most 16 source subnets.

najeal:
## Why this should be merged

Closes #399 

## How this works

This checks that the provided subnets does not exceed the limit (16) and returns error during config validation when exceeding.

## How this was tested

unit test has been added

## How is this documented


cam-schultz (approved):
LGTM. Thanks for putting this together 🙏

iansuvak (approved):
👍

Issue #311, opened 2024-05-31
Relayer executable panic does not exit test


minghinmatthewlam:
**Describe the bug**
If the relayer executable panics in an e2e test the test does not exit and fail, instead it will hang.

**To Reproduce**
Can try to replace relayer main with a panic

**Expected behavior**
Should log the error and panic, then exit the test case with a failure.

**Logs**
This was the panic case hit that discovered the bug. The initial panic cause should now be fixed.
```
2024-05-31T18:57:35.0333859Z ERROR[05-31|18:57:35.033] panic: couldn't build config: failed to validate configuration: invalid account private key hex string 
2024-05-31T18:57:35.0335216Z ERROR[05-31|18:57:35.033]  
2024-05-31T18:57:35.0335684Z ERROR[05-31|18:57:35.033] goroutine 1 [running]: 
2024-05-31T18:57:35.0336138Z ERROR[05-31|18:57:35.033] main.main() 
2024-05-31T18:57:35.0336807Z ERROR[05-31|18:57:35.033] 	/home/runner/work/awm-relayer/awm-relayer/main/main.go:68 +0x1d0e 
```

**To best protect the Avalanche community security bugs should be reported in accordance to our [Security Policy](../../SECURITY.md)**

PR #411, opened 2024-08-05
Move relayer to subfolder

### LINKED ISSUE #409: Move relayer main package and related configs to it's own subfolder
**Context and scope**
After #368 is merged we should do some clean up and move relayer specific main, api and config packages and to the relayer subfolder. 

TODOs before merging
- [ ] Reorganize configs such that we can explicitly confirm that relayer config and signature aggregator config both implement `peers.Config` interfaces. 
- [ ] Extract duplicative e2e test functionality for running a binary into a helper function that both applications use. 
- [ ] Re-evaluate locations of scripts, build and testing
- [ ] Re-evaluate passing configs around by reference

**Open questions**
- Should we consider renaming the repo at this point? To what?

iansuvak:
## Why this should be merged

Partially fixes #409. This would also enable us to rename this repo into something more general. `icm-offchain-services` ?

## How this works
**Moved**
- `main` to `relayer/main`
- relayer specific configs to `relayer/config`
- relayer specific network initialization code from `peers.AppRequestNetwork` to it's own `main.go`
- Updates Readme and creates a stub top level README that links to individual applications

**Kept at top level**
- E2E testing
- scripts folder
- build folder (binaries are distinguished by name but output to the same folder)

**Not moved**
- database
- messages

The folders above are relayer specific right now but I haven't moved them yet. Should we? They seem auxillary to the core function and could be feasibly generalized in the future. 

## How this was tested

- CI

## How is this documented

Updated READMEs


scripts/build_relayer.sh
```diff
@@ -22,15 +22,22 @@ version_lt() {
...
+
+# Relayer directory
 RELAYER_PATH=$(
     cd "$(dirname "${BASH_SOURCE[0]}")"
+    cd ../relayer && pwd
+)
+
+# Base directory
+BASE_PATH=$(
+    cd $RELAYER_PATH
```
geoff-vball, line 34: This seems like a backwards way of doing this. We should define `BASE_PATH` as 
`cd "$(dirname "${BASH_SOURCE[0]}")" && pwd`
and then `RELAYER_PATH` as `"$BASE_PATH"/relayer`

geoff-vball (commented):
One nit otherwise LGTM

scripts/build_relayer.sh
iansuvak: Fixed in 07e43af8f585bacf6016d6b34a2a52a0d4945533

geoff-vball (approved):

README.md
```diff
@@ -1,432 +1,8 @@
-# AWM Relayer
+# Avalanche ICM Off-chain Services
 
-Reference relayer implementation for cross-chain Avalanche Warp Message delivery.
+This repository contains services related to Avalanche ICM (Interchain Messaging) systems along with shared code to support them.
```
michaelkaplan13, line 3:
```suggestion
This repository contains off-chain services that help support Avalanche Interchain Messaging (ICM).
```
Super nit on wording

README.md
```diff
@@ -1,432 +1,8 @@
...
-  To update the version of Teleporter used by the E2E tests, update these values with the latest contract deployment information. For more information on how to deploy the Teleporter contract, see the [Teleporter documentation](https://github.com/ava-labs/teleporter/tree/main/utils/contract-deployment).
-
-### Generate Mocks
-
-[Gomock](https://pkg.go.dev/go.uber.org/mock/gomock) is used to generate mocks for testing. To generate mocks, run the following command at the root of the project:
-
-```bash
-go generate ./...
-```
+1. [AWM Relayer](relayer/README.md)
```
michaelkaplan13, line 7: Should this be ICM relayer now?

May also be helpful to include a 1 sentence description of each service at the top level here.

relayer/main/main.go
```diff
@@ -534,3 +535,125 @@ func initializeMetrics() (prometheus.Gatherer, prometheus.Registerer, error) {
 	}
 	return gatherer, registry, nil
 }
+
+func initializeConnectionsAndCheckStake(
```
michaelkaplan13, line 539: `initializeConnectionsAndCheckStake`, `connectToNonPrimaryNetworkPeers`, and `connectToPrimaryNetworkPeers` may make more sense to live in a non-main package specific to the relayer, but not a huge deal either way.

michaelkaplan13 (commented):
👍 I this structure will much better suit us going forward.

relayer/main/main.go
iansuvak: Yeah I thought about it originally, but they are only called as part of initialization at app startup. I am happy to move them but any idea on the naming? I don't think we need a new package for it since it would define no structs, methods or interfaces. 

Maybe `network_utils.go` in the `relayer` package?

README.md
iansuvak: I'll include descriptions but do we want to hold off on changing the name here until we change the name of the build binary / rest of the documentation / and the whole repo?

michaelkaplan13: Ah yeah good call

relayer/main/main.go
michaelkaplan13: Sounds good to me 👍

relayer/network_utils.go
```diff
@@ -0,0 +1,139 @@
...
+				"Failed to connect to a threshold of stake",
+				zap.String("destinationBlockchainID", blockchainID.String()),
+				zap.Uint64("connectedWeight", connectedValidators.ConnectedWeight),
+				zap.Uint64("totalValidatorWeight", connectedValidators.TotalValidatorWeight),
+				zap.Any("warpQuorum", quorum),
+			)
+			return err
+		}
+	}
+	return nil
```
geoff-vball, line 114: In the case that we don't connect to a threshold of stake, we log an error here and then return `nil`, is that's what is intended?

I think the intended flow here should be to check for an error first, return the error if present, then check `ok`, which should maybe return a new error if we haven't connected to enough stake.

relayer/network_utils.go
```diff
@@ -0,0 +1,139 @@
...
+		)
+		return false, nil, err
+	}
+	return utils.CheckStakeWeightExceedsThreshold(
+		big.NewInt(0).SetUint64(connectedValidators.ConnectedWeight),
+		connectedValidators.TotalValidatorWeight,
+		quorum.QuorumNumerator,
+		quorum.QuorumDenominator,
+	), &quorum, nil
+}
```
geoff-vball, line 139: Going along with my above comment - I think this would actually be better served _not_ being a function - the tri-state return is kind of confusing. I think we should just inline this in the calling function. 

If we wanted, we could make a helper that takes `connectedValidators` and a `quorum` and returns a bool to do the calculation on lines 133-138

relayer/network_utils.go
iansuvak: Thank you for finding this. It indeed seems like a bug in the existing implementation, but this whole initialization step is not really necessary I only kept it for now to maintain existing behavior. 

For now I updated the comment to indicate actual behavior, since I do think that the existing bug i.e. more permissive behavior is actually what we want but if we align on another approach I'm happy to implement it.

relayer/network_utils.go
```diff
@@ -0,0 +1,143 @@
...
+			logger.Error(
+				"Failed to connect to canonical validators",
+				zap.String("subnetID", subnetID.String()),
+				zap.Error(err),
+			)
+			return err
+		}
+
+		if ok, quorum, err := checkForSufficientConnectedStake(logger, cfg, connectedValidators, blockchainID); !ok {
+			logger.Error(
```
cam-schultz, line 108: Thoughts on downgrading this log (and the above) to `Warn` since we're treating it as expected?

cam-schultz (approved):
LGTM. I left once comment about the level of one of the logs, but don't feel strongly either way.

relayer/network_utils.go
iansuvak: Addressed! Thanks!

relayer/network_utils.go
bernard-avalabs: I like the suggestion to break up the quorum check and the stake weight check into separate functions.

bernard-avalabs (commented):
Generally looks good to me. Just left some minor comments.

cam-schultz (commented):
The entry point path in `.goreleaser.yml` needs to be updated from `main/main.go` to `relayer/main/main.go`

https://github.com/ava-labs/awm-relayer/blob/main/.goreleaser.yml#L4

geoff-vball (approved):

cam-schultz (approved):

bernard-avalabs (approved):

relayer/network_utils.go
```diff
@@ -0,0 +1,143 @@
...
+// Sufficient stake is determined by the Warp quora of the configured supported destinations,
+// or if the subnet supports all destinations, by the quora of all configured destinations.
+func InitializeConnectionsAndCheckStake(
+	logger logging.Logger,
+	network *peers.AppRequestNetwork,
+	cfg *config.Config,
+) error {
+	for _, sourceBlockchain := range cfg.SourceBlockchains {
+		if sourceBlockchain.GetSubnetID() == constants.PrimaryNetworkID {
+			if err := connectToPrimaryNetworkPeers(logger, network, cfg, sourceBlockchain); err != nil {
```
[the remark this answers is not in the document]
iansuvak: Updated #311 to include addressing followups from you and @geoff-vball in a follow-up

PR #412, opened 2024-08-06
decoupling relayer and checkpointmanager implementation

najeal:
## Why this should be merged

This is decoupling relayer and checkpoint manager implementation.
It reduces the number of args in the relayer factory as they were just used to instantiate checkpoint manager.

## How this works

Now, the checkpoint manager is injected in the relayer factory as a dependency.

## How this was tested

It uses the existing tests.

## How is this documented


geoff-vball (approved):
This is great, thanks for the change! :)

cam-schultz (approved):

PR #415, opened 2024-08-07
cache validator signatures

### LINKED ISSUE #382: Cache individual and aggregated signatures
**Context and scope**
Currently the service stores individual signatures from the validators only for the duration of the request. We should add an read-through in-memory cache layer that is hit prior to requesting a signature from a node. the interface for interacting with the cache should be behind an interface that we can re-use if we decide to implement a persistence layer in the future. 

**Open questions**
- The max size of the cache should be configurable?

### LINKED ISSUE #389: Re-aggregation of stored signatures
**Context and scope**
Once #382 is implemented, we should add logic to re-use the individually cached signatures to reconstruct an aggregate signature with a larger `QuorumNum` and or larger/different validator set without re-querying still active validators who have already signed the message. 
**Open questions**

feuGeneA:
## Why this should be merged
fixes #382
fixes #389 

## How this was tested
see included test modifications


signature-aggregator/aggregator/cache/cache.go
```diff
@@ -0,0 +1,54 @@
...
+	return value, isCached
+}
+
+func (c *Cache) Set(warpMsgID ids.ID, value *warp.Message) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	if len(c.cache) < maxSize {
+		c.cache[warpMsgID] = *value
+	} else {
+		c.logger.Warn("Cache length is at max size; skipped Set")
```
iansuvak, line 52: Since keys don't expire here we should make this an LRU (or FIFO?) cache.

signature-aggregator/aggregator/cache/cache.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/logging"
+	"github.com/ava-labs/avalanchego/vms/platformvm/warp"
+	"go.uber.org/zap"
+)
+
+const (
+	initialCapacity = 1024
+	maxSize         = 1024 * 1024
```
iansuvak, line 14: MaxSize should be configurable via the application configs. We should have a sane default though

cam-schultz (commented):
My understanding was that we wanted to cache individual BLS signatures so that we could reconstruct the aggregate signature from the cache even if the validator set changes. By caching the full signed Warp message, we skip fetching signatures of already requested messages _only if the validator set hasn't changed_.  We also need to be careful when caching the full signed message, since it will no longer be valid after the validator set changes.

iansuvak:
> My understanding was that we wanted to cache individual BLS signatures so that we could reconstruct the aggregate signature from the cache even if the validator set changes. By caching the full signed Warp message, we skip fetching signatures of already requested messages _only if the validator set hasn't changed_. We also need to be careful when caching the full signed message, since it will no longer be valid after the validator set changes.

 Indeed this is the case. We want to cache individual signatures as well as full signature only (not the message). Before returning the full signature we can query for the current validator set to confirm if it's still correct

signature-aggregator/aggregator/cache/cache.go
feuGeneA: the current commit uses the same LRU cache as the compliance screener.

signature-aggregator/aggregator/cache/cache.go
feuGeneA: done. I'm not sure what would be sane to you though. I left this number in place. Happy to change it to whatever the team thinks.

signature-aggregator/aggregator/cache/cache.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+type Cache struct {
+	logger logging.Logger
+	lock   sync.RWMutex
+	cache  map[[ids.IDLen]byte]warp.Message
```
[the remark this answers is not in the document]
feuGeneA: done. I'm not sure what would be sane to you though. I left this number in place. Happy to change it to whatever the team thinks.

signature-aggregator/aggregator/cache/cache.go
```diff
@@ -0,0 +1,69 @@
...
+) {
+	cachedValue, ok := c.Get(msgID)
+	if !ok {
+		c.signatures.Add(cacheKey(msgID), make(cacheValue))
+		cachedValue, ok = c.Get(msgID)
+		if !ok {
+			c.logger.Error(
+				"failed to create cache entry; this should never happen",
+			)
+		}
```
geoff-vball, line 66:
```suggestion
		cachedValue = make(cacheValue)
		c.signatures.Add(cacheKey(msgID), make(cacheValue))
```

signature-aggregator/api/api.go
```diff
@@ -147,6 +147,7 @@ func signatureAggregationAPIHandler(
 			msg := "Failed to aggregate signatures"
 			logger.Warn(msg, zap.Error(err))
 			writeJSONError(logger, w, msg)
+			return
```
geoff-vball, line 150: I know this wasn't you, but should we also return after writing the JSON error on line 137?

config/config.go
```diff
@@ -34,6 +34,7 @@ const (
 	defaultAPIPort             = uint16(8080)
 	defaultMetricsPort         = uint16(9090)
 	defaultIntervalSeconds     = uint64(10)
+	defaultSignatureCacheSize  = uint(1024 * 1024)
```
geoff-vball, line 37: I think we may want a cache size much smaller than this. 1M entries could be gigabytes worth of data that we'd have in memory, and the problem we're solving here is mostly an optimization for validator churn, so most entries should be needed for a relatively short amount of time.

signature-aggregator/aggregator/cache/cache.go
```diff
@@ -0,0 +1,69 @@
...
+	"go.uber.org/zap"
+)
+
+type Cache struct {
+	logger     logging.Logger
+	signatures *lru.Cache[cacheKey, cacheValue]
+}
+
+type cacheKey ids.ID // warp message ID
+type cacheValue map[PublicKeyBytes]SignatureBytes
```
geoff-vball, line 20: Can we just use the regular types directly? It seems odd to have to wrap everything before passing them in

signature-aggregator/config/config.go
```diff
@@ -13,6 +13,8 @@ import (
 const (
 	defaultAPIPort     = uint16(8080)
 	defaultMetricsPort = uint16(8081)
+
+	DefaultSignatureCacheSize = uint(1024 * 1024)
```
geoff-vball, line 17: Is there somewhere we could move this so that we don't have to duplicate it between here and the top-level config?

signature-aggregator/aggregator/cache/cache.go
feuGeneA: For the public keys, we _can_ use the regular types directly in `cache.go` but when the interactions happen from `aggregator.go` they will need to be wrapped in _something_ either way, because `warp.Validator.PublicKeyBytes` is a slice, and a map key can only be a fixed length array, so even if we change to using regular types directly, we'll still have to wrap the arguments to `Get` and `Add` with `[bls.PublicKeyLen]byte(...)`.

I'm fine with going either way on it, but at the time I wrote this it felt "cleaner" to have a name for a type that needs to be repeated several times over. (Note this is just like what's done with the `type blsSignatureBuf [bls.SignatureLen]byte` in `aggregator.go`.)

signature-aggregator/config/config.go
feuGeneA: Yeah, I wasn't too happy about this. But it's what's already happening here with the other config values.

And to make matters worse, there's a separate set of hardcoded "default config" values in `tests/utils/utils.go` that are used by the tests because the tests don't go through the viper procedure.

So unfortunately this value is repeated _4_ times: validator viper default, validator test default, aggregator viper default, aggregator test default.

But again, we've already got that duplication across the API port and the metrics port, so I'd prefer to defer that to after this PR. I'll raise an issue for it.

signature-aggregator/api/api.go
feuGeneA: good catch. the one I added already was because I was hitting it in my testing, but I didn't think to look for more. thanks! addressed in https://github.com/ava-labs/awm-relayer/pull/415/commits/acb3e80a838bbf3c3305d93179e55dcad4df7a46

signature-aggregator/aggregator/cache/cache.go
feuGeneA: I tried this but it seems like the map that's constructed here (`cachedValue = make(cacheValue)`) is copied into the cache via `Add()`, so that the subsequent assignment to a map entry only affects the instance in the function scope, not the one that was added to the cache.

I got rid of the "should never happen" weirdness in https://github.com/ava-labs/awm-relayer/pull/415/commits/b8862e7524de554015c9a0224036e17eb0634036, in exchange for duplicating the signature assignment line. let me know what you think.

config/config.go
```diff
@@ -61,6 +62,7 @@ type Config struct {
 	DestinationBlockchains []*DestinationBlockchain `mapstructure:"destination-blockchains" json:"destination-blockchains"`
 	ProcessMissedBlocks    bool                     `mapstructure:"process-missed-blocks" json:"process-missed-blocks"`
 	DeciderURL             string                   `mapstructure:"decider-url" json:"decider-url"`
+	SignatureCacheSize     uint                     `mapstructure:"signature-cache-size" json:"signature-cache-size"`
```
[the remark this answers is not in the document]
feuGeneA: addressed in https://github.com/ava-labs/awm-relayer/pull/415/commits/c0a6a0a632694246b83001e4d9a5c9add4235b03

config/config.go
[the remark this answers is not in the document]
feuGeneA: addressed in https://github.com/ava-labs/awm-relayer/pull/415/commits/c0a6a0a632694246b83001e4d9a5c9add4235b03

iansuvak:
Looks like this includes functionality of #389 . I want to take a more thorough look before shipping it but looks good so far. 
I'm not familiar with the differences between that and the hashicorp implementation but did you consider using the LRU cache that's defined in [avalanchego](https://pkg.go.dev/github.com/ava-labs/avalanchego/cache)?

feuGeneA:
> did you consider using the LRU cache that's defined in [avalanchego](https://pkg.go.dev/github.com/ava-labs/avalanchego/cache)?

No, I didn't even realize that existed. I just took [Geoff's suggestion](https://github.com/ava-labs/awm-relayer/pull/415#discussion_r1707885616) and ran with it. Happy to convert it over though if that's what we think makes sense. It should be a pretty self-contained change.

signature-aggregator/aggregator/cache/cache.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (c *Cache) Get(msgID ids.ID) (cacheValue, bool) {
+	cachedValue, isCached := c.signatures.Get(cacheKey(msgID))
+
+	if isCached {
+		c.logger.Debug("cache hit", zap.Stringer("msgID", msgID))
+		return cachedValue, true
+	} else {
+		c.logger.Debug("cache miss", zap.Stringer("msgID", msgID))
+		return make(cacheValue), false
```
cam-schultz, line 49: We can just return `nil` instead of allocating.

signature-aggregator/aggregator/cache/cache.go
```diff
@@ -0,0 +1,65 @@
...
+	pubKey PublicKeyBytes,
+	signature SignatureBytes,
+) {
+	if signatures, ok := c.Get(msgID); ok {
+		signatures[pubKey] = signature
+	} else {
+		signatures := make(cacheValue)
+		signatures[pubKey] = signature
+		c.signatures.Add(cacheKey(msgID), signatures)
+	}
```
cam-schultz, line 64: This control flow is a bit awkward due to the multiple uses of `signatures` and the repeated assignment to the map. The following rewrite cleans it up a bit I think:
```
var (
    sigs cacheValue
    ok bool
)
if sigs, ok = c.Get(msgID); !ok {
    sigs = make(cacheValue)
}
sigs[pubKey] = signature
c.signatures.Add(cacheKey(msgID), sigs)
```

signature-aggregator/aggregator/cache/cache.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/crypto/bls"
+	"github.com/ava-labs/avalanchego/utils/logging"
+	lru "github.com/hashicorp/golang-lru/v2"
+	"github.com/pingcap/errors"
+	"go.uber.org/zap"
+)
+
+type Cache struct {
```
cam-schultz, line 14: There is a new goroutine for each [http.Serve](https://pkg.go.dev/net/http#Serve) request handled. We should add a lock to the cache to provide mutual exclusion.

cam-schultz (commented):
A handful of minor comments, but generally LGTM

signature-aggregator/aggregator/cache/cache.go
feuGeneA: The LRU implementation currently in use says this in [its docs](https://pkg.go.dev/github.com/hashicorp/golang-lru/v2): "All caches in this package take locks while operating, and are therefore thread-safe for consumers."

signature-aggregator/aggregator/cache/cache.go
feuGeneA: addressed in https://github.com/ava-labs/awm-relayer/pull/415/commits/d77300b756c2284c6cf4ffafaf46ca18fc02441e

signature-aggregator/aggregator/cache/cache.go
feuGeneA: I like this, thanks. addressed in https://github.com/ava-labs/awm-relayer/pull/415/commits/7196197cc44c0fb4b5d9bd4aad83a69b174e70c2

signature-aggregator/aggregator/aggregator.go
```diff
@@ -382,14 +422,37 @@ func (s *SignatureAggregator) handleResponse(
...
+		connectedValidators,
+		quorumPercentage,
+	); err != nil {
+		return nil, true, err
+	} else if signedMsg != nil {
+		return signedMsg, true, nil
+	}
+
+	// Not enough signatures, continue processing messages
+	return nil, true, nil
```
geoff-vball, line 438: I generally dislike the inlined `if` when the function returns more than just `err`.

signature-aggregator/aggregator/cache/cache.go
geoff-vball: I think we could pass in `thisIsASlice[:]' to a function that takes a fixed size array to get the conversion

signature-aggregator/aggregator/cache/cache.go
feuGeneA: addressed in https://github.com/ava-labs/awm-relayer/pull/415/commits/d77300b756c2284c6cf4ffafaf46ca18fc02441e

signature-aggregator/aggregator/cache/cache.go
feuGeneA: thanks, I like that, addressed in https://github.com/ava-labs/awm-relayer/pull/415/commits/7196197cc44c0fb4b5d9bd4aad83a69b174e70c2

signature-aggregator/aggregator/cache/cache.go
feuGeneA: I tried, without any luck:
```
gene-dev 08/13 22:21:21 (signature-aggregation-api-cache) 1 ~/dev/awm-relayer$ git diff signature-aggregator/                                                   
diff --git a/signature-aggregator/aggregator/aggregator.go b/signature-aggregator/aggregator/aggregator.go                                                      
index d500e82..73ee1da 100644                                                                                                                                   
--- a/signature-aggregator/aggregator/aggregator.go                                                                                                             
+++ b/signature-aggregator/aggregator/aggregator.go                                                                                                             
@@ -412,7 +412,7 @@ func (s *SignatureAggregator) handleResponse(                                                                                               
                signatureMap[vdrIndex] = signature                                                                                                              
                s.cache.Add(                                                                                                                                    
                        unsignedMessage.ID(),                                                                                                                   
-                       cache.PublicKeyBytes(validator.PublicKeyBytes),                                                                                         
+                       validator.PublicKeyBytes[:],                                                                                                            
                        cache.SignatureBytes(signature),                                                                                                        
                )                                                                                                                                               
                accumulatedSignatureWeight.Add(accumulatedSignatureWeight, new(big.Int).SetUint64(validator.Weight))                                            
diff --git a/signature-aggregator/aggregator/cache/cache.go b/signature-aggregator/aggregator/cache/cache.go                                                    
index e0fa57d..6d7a4e8 100644                                                                                                                                   
--- a/signature-aggregator/aggregator/cache/cache.go                                                                                                            
+++ b/signature-aggregator/aggregator/cache/cache.go                                                                                                            
@@ -52,7 +52,7 @@ func (c *Cache) Get(msgID ids.ID) (cacheValue, bool) {                                                                                        
                                                                                                                                                                
 func (c *Cache) Add(                                                                                                                                           
        msgID ids.ID,                                                                                                                                           
-       pubKey PublicKeyBytes,                                                                                                                                  
+       pubKey [bls.PublicKeyLen]byte,                                                                                                                          
        signature SignatureBytes,                                                                                                                               
 ) {                                                                                                                                                            
        var (                                                                                                                                                   
gene-dev 08/13 22:21:27 (signature-aggregation-api-cache) 0 ~/dev/awm-relayer$ scripts/build.sh                                                                 
Running protobuf fmt...                                                                                                                                         
Running protobuf lint check...                                                                                                                                  
Re-generating protobuf...                                                                                                                                       
Building AWM Relayer Version: v1.4.0-rc.2 at /home/gene/dev/awm-relayer/build/awm-relayer                                                                       
# github.com/ava-labs/awm-relayer/signature-aggregator/aggregator                                                                                               
signature-aggregator/aggregator/aggregator.go:415:28: cannot use validator.PublicKeyBytes[:] (value of type []byte) as [48]byte value in argument to s.cache.Add
Building Signature Aggregator Version: v1.4.0-rc.2 at /home/gene/dev/awm-relayer/signature-aggregator/build/signature-aggregator                                
# github.com/ava-labs/awm-relayer/signature-aggregator/aggregator                                                                                               
aggregator/aggregator.go:415:28: cannot use validator.PublicKeyBytes[:] (value of type []byte) as [48]byte value in argument to s.cache.Add                     
```

Feel free to give it a try and let me know what you prefer.

signature-aggregator/aggregator/aggregator.go
feuGeneA: Happy to defer to team opinion, but... in general I find the inlined `if` syntax weird and generally avoid it, but here I wrote it this way in order to limit the scope of `signedMsg` so that it's more clear to the reader that this instance of `signedMsg` _only_ refers to the one that's returned here, and has nothing to do with the one that's returned from `handleResponse` later in this function.

signature-aggregator/aggregator/aggregator.go
```diff
@@ -151,7 +151,13 @@ func (s *SignatureAggregator) CreateSignedMessage(
 				)
 			}
 		}
+		s.metrics.SignatureCacheHits.Add(float64(len(signatureMap)))
```
cam-schultz, line 154: What's the reasoning behind incrementing based on the number of signatures (i.e. the cache value) rather than the number of key hits? Signature-based metrics would make it more difficult to identify appropriate cache sizes based on usage.

signature-aggregator/metrics/metrics.go
```diff
@@ -62,6 +64,14 @@ var Opts = struct {
 		Name: "invalid_signature_responses",
 		Help: "Number of responses from validators that were not valid signatures",
 	},
+	SignatureCacheHits: prometheus.CounterOpts{
+		Name: "signature_cache_hits",
+		Help: "Number of aggregated signatures that were found in the cache",
```
cam-schultz, line 69:
```suggestion
		Help: "Number of signatures that were found in the cache",
```

signature-aggregator/metrics/metrics.go
```diff
@@ -62,6 +64,14 @@ var Opts = struct {
...
+	},
+	SignatureCacheMisses: prometheus.CounterOpts{
+		Name: "signature_cache_misses",
+		Help: "Number of aggregated signatures that were not found in the cache",
```
cam-schultz, line 73:
```suggestion
		Help: "Number of signatures that were not found in the cache",
```

signature-aggregator/aggregator/aggregator.go
iansuvak: Not my PR but not sure what do you mean by this. The `signatureMap` here is just the number of key hits from my understanding.

cam-schultz: The cache is keyed by `warpMessageID`, but we're adding metrics based on the size of the corresponding value. If the Warp message is signed by 1 or 1000 validators, it's still a cache hit or miss, but the measuring the number of signatures would tell a wildly different story than measuring the actual cache hit/miss rate.

feuGeneA: as discussed on standup, I left the metric in place on a per-signature basis. also, I changed the "cache miss" metric to reflect the number of signatures that need to be requested from the network, in https://github.com/ava-labs/awm-relayer/pull/415/commits/70b7722ea01cdd358da4568073638bdcfa67e610

signature-aggregator/aggregator/cache/cache.go
feuGeneA: addressed in https://github.com/ava-labs/awm-relayer/pull/415/commits/aa631b608d8c85e6c1a3e23815d0cf568f608184

geoff-vball (approved):

signature-aggregator/aggregator/aggregator.go
```diff
@@ -214,6 +209,9 @@ func (s *SignatureAggregator) CreateSignedMessage(
 	// Query the validators with retries. On each retry, query one node per unique BLS pubkey
 	for attempt := 1; attempt <= maxRelayerQueryAttempts; attempt++ {
 		responsesExpected := len(connectedValidators.ValidatorSet) - len(signatureMap)
+		if len(signatureMap) > 0 {
+			s.metrics.SignatureCacheMisses.Add(float64(responsesExpected))
+		}
```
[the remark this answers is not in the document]
feuGeneA: you're right! thank you! addressed in https://github.com/ava-labs/awm-relayer/pull/415/commits/0d6da49763e8776726469c1f1674dc36bb9f7415

cam-schultz (approved):

iansuvak (approved):

PR #421, opened 2024-08-08
refactor: database and checkpoint dependencies

### LINKED ISSUE #418: Refactor: RelayerID declared in database
**Context and scope**
RelayerID is currently declared in the database package. I think we can find a better place for the type declaration but I did not find a ideal place yet.

It could be moved in relayer package but it is not possible at the moment as it would create an import cycle. This would require to create some type abstraction to avoid direct dependency.
A temporary work around can be to create new package for the RelayerID but I am not a big fan.

najeal:
## Why this should be merged

It closes #418 
This PR is a proposition and open discussion for the refactoring of the database and the checkpoint packages.

## How this works

This declares the DB interface that checkpoint manager is expecting.
It hides database complexity and implementation details.
Checkpoint package has no more dependency on database package, database has dependency on checkpoint.

RelayerID is a "business object", the declaration has been moved into the relayer package.

## How this was tested

Existing tests have been adapted.

## How is this documented


database/kvstore.go
```diff
@@ -0,0 +1,57 @@
...
+	Put(relayerID common.Hash, key DataKey, value []byte) error
+}
+
+func NewKeyValueDatabase(logger logging.Logger, cfg *config.Config) (keyValueDatabase, error) {
+	relayerIDs := relayer.GetConfigRelayerIDs(cfg)
+	dbConnect := func() (keyValueDatabase, error) { return NewJSONFileStorage(logger, cfg.RedisURL, relayerIDs) }
+	usedDB := "Redis"
+	if cfg.RedisURL != "" {
+		dbConnect = func() (keyValueDatabase, error) { return NewRedisDatabase(logger, cfg.StorageLocation, relayerIDs) }
+		usedDB = "JSON"
```
geoff-vball, line 36: The `usedDB` values here are flipped.

geoff-vball (commented):
Hey, thanks for this! This generally looks really helpful, just one tiny bug but otherwise lgtm

geoff-vball (approved):
LGTM, but @cam-schultz should take a look at this as well :)

najeal:
This looks out dated, closing it

PR #422, opened 2024-08-09
Signature Aggregator ACP-118 updates

iansuvak:
~~This should pass the tests but not necessarily be merged into the PR until all of the required commits make it into default branches. Right now the main outstanding one would be the subnet-evm that references newer avalanchego commit that includes latest coreth changes.~~

The commits have been updated to reference default branches ones now. 

Includes a bugfix on the validation logic.

I'm open to suggestions on cleaner prefix packing and to avoid copying into `blsSignatureBuf` if anyone has ideas.


feuGeneA (commented):
LGTM! Happy to see it working!

geoff-vball:
I think you may have accidentally included an update to teleporter? The go.work.sum was added back

iansuvak:
> I think you may have accidentally included an update to the `teleporter` submodule? The go.work.sum was added back

Thanks, I had teleporter updated to a branch commit that was not necessary so I downgraded it v1.0.4 which is what it is on `main` branch I believe https://github.com/ava-labs/awm-relayer/tree/main/tests/contracts/lib . I'm happy to bump to any version that we need but it wouldn't be related to this and the parent PR which shouldn't require changes to teleporter dependencies

geoff-vball (approved):

signature-aggregator/aggregator/aggregator.go
```diff
@@ -459,8 +462,9 @@ func (s *SignatureAggregator) isValidSignatureResponse(
 		)
 		return blsSignatureBuf{}, false
 	}
-
-	return signature, true
+	blsSig := blsSignatureBuf{}
+	copy(blsSig[:], signature[:])
```
michaelkaplan13, line 466: Should we have a check higher up that the length of `signature` is the expected length?

signature-aggregator/api/api.go
```diff
@@ -123,12 +123,13 @@ func signatureAggregationAPIHandler(logger logging.Logger, aggregator *aggregato
 			return
 		}
 
-		if isEmptyOrZeroes(message.Bytes()) || isEmptyOrZeroes(justification) {
+		if isEmptyOrZeroes(message.Bytes()) && isEmptyOrZeroes(justification) {
```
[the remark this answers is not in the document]
iansuvak: Filed #446

signature-aggregator/aggregator/aggregator.go
iansuvak: Possibly, we do call `bls.SignatureFromBytes` and `bls.Verify` on it's result so presumably those wouldn't work if it wasn't correct lenght but an explicit check wouldn't hurt

iansuvak: Addressed in 811c97a

bernard-avalabs (commented):
LGTM

PR #426, opened 2024-08-12
Fix start height edge case

cam-schultz:
## Why this should be merged
Fixes a bug when handling the edge case in which the the DB stores a height, but the catch up mechanism is disabled. In that scenario, we were initializing the checkpoint manager in such a way that it never committed to the database.

## How this works
Suppose the DB stores block 100, the current height is 200, and on restart, we set `process-missed-blocks=false`. The relayer will correctly skip blocks 100-200 for processing, but will initialize the checkpoint manager with block 100. To prevent gaps, the checkpoint manager only commits blocks if they are exactly 1 greater than the max committed height. In this scenario, the next block height the checkpoint manager sees is 200, which is never committed.

This fixes this by:
1. Explicitly setting the starting height to the next block (current height + 1) if `process-missed-blocks=false`
2. Always starts the goroutine to process blocks on startup, so that there are no gaps. Even if `process-missed-blocks=false`, there's a race condition between fetching of the current height and the beginning of the Listener that could cause a gap. 

## How this was tested
CI

## How is this documented
N/A


main/main.go
```diff
@@ -431,20 +430,26 @@ func createApplicationRelayersForSourceChain(
...
-			sourceBlockchain.ProcessHistoricalBlocksFromHeight,
-			currentHeight,
-		)
-		if err != nil {
-			logger.Error(
-				"Failed to calculate starting block height",
-				zap.String("relayerID", relayerID.ID.String()),
-				zap.Error(err),
+		// Calculate the starting block height for the relayer only if we're processing historical blocks
+		// Otherwies, the first block we process is the next block after the current height
```
iansuvak, line 434:
```suggestion
		// Otherwise, the first block we process is the next block after the current height
```

main/main.go
```diff
@@ -431,20 +430,26 @@ func createApplicationRelayersForSourceChain(
...
+		height := currentHeight + 1
```
iansuvak, line 435: nit, free to drop: I  prefer having this in an else statement below the `if cfg.ProcessMissedBlocks` instead of setting it here and overwriting it.

tests/relay_message_api.go
```diff
@@ -52,6 +52,15 @@ func RelayMessageAPI(network interfaces.LocalNetwork) {
 	)
 	warpMessage := getWarpMessageFromLog(ctx, receipt, subnetAInfo)
 
+	// // Send an additional tx to make sure the relayer doesn't process the previous one on startup
```
iansuvak, line 55: Do we want to keep this commented?

iansuvak (commented):
Generally LGTM, few non-blocking comments that I am posting early.

I'll add an approval soon but want to read/think some more on the catchup mechanism before I do.

main/main.go
```diff
@@ -431,20 +430,26 @@ func createApplicationRelayersForSourceChain(
...
+		if cfg.ProcessMissedBlocks {
+			var err error
```
minghinmatthewlam, line 437: can we declare the error with `err := database...`

cam-schultz: Unfortunately not, because doing so would declare a new variable `height` in the inner scope, shadowing the one in the outer scope.

main/main.go
geoff-vball: Yeah, I agree the else statement would make it clearer

geoff-vball (commented):
LGTM after the comments are addressed

tests/relay_message_api.go
cam-schultz: Removed

main/main.go
cam-schultz: I went with a slightly different approach so that we can emit the `process-missed-blocks=false` log only once per source chain. Lmk if you agree.

main/main.go
```diff
@@ -427,27 +426,42 @@ func createApplicationRelayersForSourceChain(
...
-		if err != nil {
-			logger.Error(
-				"Failed to calculate starting block height",
-				zap.String("relayerID", relayerID.ID.String()),
-				zap.Error(err),
+		height = currentHeight + 1
+		minHeight = height
+	}
+	for _, relayerID := range database.GetSourceBlockchainRelayerIDs(&sourceBlockchain) {
+		// Calculate the catcp-up starting block height, and update the min height if necessary
```
iansuvak, line 442:
```suggestion
		// Calculate the catch-up starting block height, and update the min height if necessary
```

iansuvak (approved):
Minor typo but LGTM

PR #429, opened 2024-08-14
Update go

geoff-vball:
## Why this should be merged
Updates go to v1.22.6.

Removes 3 instances of copying the loop variable, which is no longer necessary in go1.22


cam-schultz (commented):
Can you also update this loop please? https://github.com/ava-labs/awm-relayer/blob/main/relayer/application_relayer.go#L151

cam-schultz (approved):

PR #430, opened 2024-08-14
Add explicit gas check

cam-schultz:
## Why this should be merged
Adds an explicit check against the gas limit specified in the Teleporter message.

It's possible for the required gas limit in a Teleporter message to exceed the block gas limit. This causes the transaction to fail as expected, but also causes the relayer to enter an unhealthy state, preventing it from updating the checkpoint in its database. This can cause compounding latency that will eventually prevent the relayer from running at all.

## How this works
Generally speaking, such validity checks are done in `ShouldSendMessage` and the `Decider` service. This is done before 
signature aggregation, to short circuit that step if the message is invalid. However, the number of signers impacts the gas estimation, so we can't use that in `ShouldSendMessage`. Instead, I added a hard cap of 12 million gas, which is the C-Chain's 15 million gas block limit minus other Warp message related overhead that is conservatively estimated. Specifically, the 12 million figure assumes:
- 10kb Warp message size
- 8kb Teleporter message size
- 1000 signing validators
- 5 Teleporter receipts (the maximum amount)

## How this was tested
CI, updated unit tests

## How is this documented
N/A

### DIFF
--- a/messages/teleporter/message_handler.go
+++ b/messages/teleporter/message_handler.go
@@ -29,6 +29,10 @@ import (
 	"google.golang.org/grpc"
 )
 
+// The maximum gas limit that can be specified for a Teleporter message
+// Based on the C-Chain 15_000_000 gas limit per block, with other Warp message gas overhead conservatively estimated.
+const maxTeleporterGasLimit = 12_000_000
+
 type factory struct {
 	messageConfig   Config
 	protocolAddress common.Address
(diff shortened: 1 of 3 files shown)


michaelkaplan13 (approved):
Makes sense to me, thanks for turning it around quickly 🙏 

We may want to consider making the max required gas limit configurable per destination chain (since subnet-evm chains can set their own block gas limits), but not necessary for this PR.

iansuvak (approved):
LGTM and I agree that `ShouldSendMessage` is the right place for it for now.

 If we come up with more cases like this we could group them together down the road into a separate helper method to distinguish between cases where the message is invalid due to relayer settings vs message validity like in this case.

cam-schultz:
> We may want to consider making the max required gas limit configurable per destination chain (since subnet-evm chains can set their own block gas limits), but not necessary for this PR.

Created an issue for this: https://github.com/ava-labs/awm-relayer/issues/432

PR #434, opened 2024-08-16
signature-aggregator: add ConnectedStakeWeightPercentage metric

feuGeneA:
as discussed on standup

## How is this documented


signature-aggregator/metrics/metrics.go
```diff
@@ -31,6 +31,7 @@ var Opts = struct {
 	InvalidSignatureResponses          prometheus.CounterOpts
 	SignatureCacheHits                 prometheus.CounterOpts
 	SignatureCacheMisses               prometheus.CounterOpts
+	CachedSignatureWeightPercentage    prometheus.GaugeOpts
```
iansuvak, line 34: I feel like a Prometheus Gauge is not a great fit for this metric for consumption. It doesn't describe health/state of application as a whole but of current request. I do think that we want this data available but maybe a log would be better fit. 

An idea for useful usage of gauges might be currently connected stake-weight percentage with subnets as labels.

tests/signature_aggregator_api.go
```diff
@@ -193,7 +203,9 @@ func sampleMetrics(port uint16) map[string]uint64 {
 				log.Debug("Found metric line", "line", line)
 				parts := strings.Fields(line)
 
-				// Fetch the metric count from the last field of the line
+				metricName = strings.Replace(parts[0], "U__signature_2d_aggregator_", "", 1)
```
iansuvak, line 206: JW how does this work with multiple subnets? Can we add a test case for that?

feuGeneA: note how I constructed the metric name using the subnet ID here: https://github.com/ava-labs/awm-relayer/pull/434/files#diff-973f1f8ae5b8f6acd388e8c2fdaa6e08f999b4211536e8e3258b5df24de79d04R135

in short, the metric name/key for the existing test case comes out as `connected_stake_weight_percentage{subnetID="11111111111111111111111111111111LpoYY"}`.

If I understand correctly, we can't send a message from any subnets until the ACP-118 stuff is merged. (Right?) So at this time I can't add a test case for a different subnet, but I think the existing test does a good job of showing how it works.

iansuvak (approved):

signature-aggregator/aggregator/aggregator.go
```diff
@@ -121,6 +121,14 @@ func (s *SignatureAggregator) CreateSignedMessage(
 		s.metrics.FailuresToGetValidatorSet.Inc()
 		return nil, fmt.Errorf("%s: %w", msg, err)
 	}
+	s.metrics.ConnectedStakeWeightPercentage.WithLabelValues(
+		signingSubnet.String(),
+	).Set(
+		float64(
+			connectedValidators.ConnectedWeight /
+				connectedValidators.TotalValidatorWeight,
```
geoff-vball, line 129: This is returning a ratio, not a percentage, afaict?

bernard-avalabs (commented):
Looks good to me other than Geoff's comment about ratio vs. percentage.

signature-aggregator/aggregator/aggregator.go
geoff-vball, line 129:
```suggestion
		float64(connectedValidators.ConnectedWeight) /
		float64(connectedValidators.TotalValidatorWeight),
```

This way we actually get float division, instead of doing integer division and converting to float

feuGeneA: yes. i thought "percentage as a decimal" was a common enough representation go keep it that way, but i'm happy to name it to `...Ratio` or something if you'd like.

geoff-vball: I think I'd rather multiply by 100, I think percentage is a more readable format

feuGeneA: addressed in https://github.com/ava-labs/awm-relayer/pull/434/commits/503992a32b8960bcef7ae6505f940f1d5f571ce2

geoff-vball (approved):

PR #435, opened 2024-08-16
dockerize signature aggregator

### LINKED ISSUE #425: Dockerize signature aggregator
**Context and scope**
- [ ] Add a Dockerfile
  - The application can be built in a few ways:
    - With a build stage that uses the golang-bullseye image - see https://github.com/ava-labs/teleporter-analytics/blob/main/Dockerfile
    - Using goreleaser- see https://github.com/ava-labs/awm-relayer/blob/main/.github/workflows/release.yml#L66
  - The Dockerfile should run the service using a lightweight image. golang-bullseye is overkill, debian:11-slim is a good choice. See again https://github.com/ava-labs/teleporter-analytics/blob/main/Dockerfile#L23C6-L23C20
  - Make sure the Dockerfile exposes the health check port 8080 (and any other ports (et API, metrics)), and does not set a user.
 - [ ] Add a CI job to build & push the docker image
    -  Add a release job to publish the image to dockerhub
      - Manual docker build and push example: https://github.com/ava-labs/teleporter-analytics/blob/main/.github/workflows/release.yml
goreleaseer example: https://github.com/ava-labs/awm-relayer/blob/main/.github/workflows/release.yml
- [ ] Create the target dockerhub repo in avaplatform. Create an infra ticket in Jira.
- [ ] Add the DOCKER_USERNAME and DOCKER_PASSWORD secrets to the github repo. Create an infra ticket in Jira.

feuGeneA:
## Why this should be merged
fixes #425 

## How was this tested
by creating `...-test` tags off of the commit here to trigger the Release Actions, the results of which can be seen [here](https://github.com/ava-labs/awm-relayer/actions/runs/10713415715/job/29705392202) and [here](https://github.com/ava-labs/awm-relayer/actions/runs/10713413789/job/29705386898).


relayer/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM debian:11-slim
```
feuGeneA, line 1: Compare to the old `repo_root/Dockerfile`. I guess I made too many changes for the PR to reflect that the file was just renamed and edited.

The edits I made were to change the base image and to drop the build `ARG`. The base image was formerly `golang`, but we're not actually building any Go code in the building of this image (rather, goreleaser is pre-building the binary, and then the Dockerfile just copies that binary into the image) so there's no need to have a Go build environment. Furthermore, there's no reason to pass in a `GO_VERSION` either.

geoff-vball (approved):

.github/workflows/release_signature_aggregator.yml
```diff
@@ -0,0 +1,75 @@
...
+      - name: Git checkout
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      # The GO_VERSION must be set explicitly to be used in the Dockerfile.
+      - name: Set Go version
+        run: |
+          source ./scripts/versions.sh
+          echo GO_VERSION=$GO_VERSION >> $GITHUB_ENV
```
geoff-vball, line 24: I don't think this bit is necessary anymore now that the go.mod file can contain the patch version and `actions/setup-go` below accepts it

signature-aggregator/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM debian:11-slim
+COPY signature-aggregator /usr/bin/signature-aggregator
+EXPOSE 8080
+EXPOSE 8081
```
iansuvak, line 4: Why are we exposing this metrics port here but not in the  relayer?

feuGeneA: Great question. I did it for the aggregator because it felt like the right thing to do, since there'd be no way to consume metrics without it, but I left things the way they were for the relayer in order to limit the scope of changes.

There's a similar question to be asked about why the relayer sets a `USER`. @cam-schultz [once advised me](https://github.com/ava-labs/avalanche-bridge-v2/issues/135#issue-2228504345) to avoid setting a `USER`.

iansuvak: I think we should be consistent and remove it from here since it looks like relayer handles without exposing it directly.

.github/workflows/release_signature_aggregator.yml
feuGeneA: great catch! thank you! addressed in https://github.com/ava-labs/awm-relayer/pull/435/commits/7c289ce0d9732a239b0cc883c29e9943c02d15d1, and validated in https://github.com/ava-labs/awm-relayer/actions/runs/10724002269 and https://github.com/ava-labs/awm-relayer/actions/runs/10724006401

signature-aggregator/Dockerfile
feuGeneA: after doing a little more research, it looks like `EXPOSE` is purely documentation, with a bit of automation convenience that takes advantage of that documentation.

when you run an image with docker run, there's a `-P` option that will automatically "publish" all of the ports that the image `EXPOSE`s, but you can manually choose to publish that port, or not, or publish different ports, and the image can also listen on other ports (eg if a config file said to use a different port), and the runtime can still publish ports that were never even `EXPOSE`d.

so, it seems completely unnecessary, but good documentation, and potentially a helpful convenience to some users. my vote would be to keep it, but i don't feel that strongly about it.

iansuvak: Yeah it all makes sense. What this is documenting is default ports that are probably going to be overwritten by the config that you pass in. But I don't feel strongly either actually so can drop

iansuvak (approved):
Have we tested that awm-relayer also builds correctly? If so LGTM and would like to build 0.1.0.RC for signature-aggregator

PR #436, opened 2024-08-16
E2E testing improvements

### LINKED ISSUE #380: Make e2e tests exit cleanly
**Context and scope**
Currently when e2e tests don't exit cleanly, due to some failures or due to being interrupted by user, the avalanchego processes spun up by `tmpnet` stay alive and prevent re-running tests.  We should make sure that tests clean up after themselves and are ready for another run regardless of status of previously aborted/failed tests.

iansuvak:
## Why this should be merged
This cuts down testing time locally by more than half, from 7+ minutes to ~3:30 and implements feedback from #368 about generalizing binary running. It also closes #380

This shouldn't be merged until #411 is merged since it's based off of it. 

## How this works

- Testing time improvements come from:
  - Building executables once instead of every restart
  - Looking for Initialization Complete log message instead of sleeping blindly on startups.
- New `runExecutable` function is defined that abstracts away common logic from signature-aggregator and relayer.
- Panics and SIGINT, SIGTERMs are handled via a handler to make cleanup code run even if tests didn't exit cleanly. 

## How this was tested
CI.

Issuing sigterms during running tests via Ctrl+C no longer leaves around tmpnet avalanchego processes.


tests/e2e_test.go
```diff
@@ -91,24 +116,14 @@ var _ = ginkgo.BeforeSuite(func() {
...
-	go func() {
-		<-signalChan
-		cancelDecider()
-	}()
 	decider = exec.CommandContext(ctx, "./tests/cmd/decider/decider")
 	decider.Start()
-	go func() { // panic if the decider exits abnormally
+	go func() {
 		err := decider.Wait()
 		// Context cancellation is the only expected way for the
```
cam-schultz, line 123:
```suggestion
		// Context cancellation is the only expected way for the process to exit
```

tests/e2e_test.go
```diff
@@ -91,24 +116,14 @@ var _ = ginkgo.BeforeSuite(func() {
...
-	signalChan := make(chan os.Signal, 2)
-	signal.Notify(signalChan, os.Interrupt, syscall.SIGTERM)
-	go func() {
-		<-signalChan
-		cancelDecider()
-	}()
 	decider = exec.CommandContext(ctx, "./tests/cmd/decider/decider")
 	decider.Start()
-	go func() { // panic if the decider exits abnormally
+	go func() {
```
cam-schultz, line 121: I know this was the existing behavior, but it's kind of odd that the decider persists across all tests, while the aggregator and relayer services are spun up and torn down by each test as needed. I think we should do the same for the decider.

As far as the E2E tests go, I don't think the existing relayer E2E tests need to run the decider service at all. We should define a test that explicitly runs the no-op test decider, though.

cam-schultz: (We can defer this to a separate change, I know it's not exactly in scope of the changes you've made)

cam-schultz (approved):
This all makes sense to me and I'm very stoked for the 100% test iteration time improvement

tests/e2e_test.go
```diff
@@ -35,6 +35,24 @@ var (
 )
 
 func TestE2E(t *testing.T) {
+	// In case of a panic we need to recover to ensure the Ginkgo cleanup is done
+	defer func() {
```
iansuvak, line 39: I don't think we actually need this. It looks like this would only catch top level panics, and those do seem to be handled by Ginkgo. 

Issues we had were with panics in goroutines but this doesn't catch those.

relayer/main/main.go
```diff
@@ -275,6 +275,7 @@ func main() {
 			)
 		})
 	}
+	logger.Info("Initialization complete")
```
iansuvak, line 278: This is not actually safe yet. There is a race condition in the API actually becoming available. We need to add a loop pinging it's own healthcheck before emitting the message. Alternatively we can make the testing code hit the test loop instead of scanning stdout.

I went with the stdout approach to begin with because we were already reading it, and health endpoint isn't yet implemented for the signature aggregator

tests/e2e_test.go
```diff
@@ -91,24 +116,14 @@ var _ = ginkgo.BeforeSuite(func() {
...
-	go func() { // panic if the decider exits abnormally
+	go func() {
 		err := decider.Wait()
-		// Context cancellation is the only expected way for the
-		// process to exit, otherwise panic
+		// Context cancellation is the only expected way for the process to exit
+		// otherwise log an error but don't panic to allow for easier cleanup
 		if !errors.Is(ctx.Err(), context.Canceled) {
-			panic(fmt.Errorf("decider exited abnormally: %w", err))
+			log.Error("Decider exited abnormally: ", "error", err)
```
geoff-vball, line 126: Can we make these `log.Fatal`? With the avalanchego logger, fatal doesn't exit the program.

geoff-vball (approved):
Love this! Added one suggestion, feel free to take it or leave it

tests/allowed_addresses.go
```diff
@@ -172,12 +172,12 @@ func AllowedAddresses(network interfaces.LocalNetwork) {
...
-	relayerCleanup := testUtils.BuildAndRunRelayerExecutable(ctx, relayerConfigPath1)
+	relayerCleanup, readyChan := testUtils.RunRelayerExecutable(ctx, relayerConfigPath1)
 	defer relayerCleanup()
 
-	// Sleep for some time to make sure relayer has started up and subscribed.
-	log.Info("Waiting for the relayers to start up")
-	time.Sleep(10 * time.Second)
+	// Wait for relayer to start up
+	log.Info("Waiting for the relayer to start up")
+	testUtils.WaitForChannelClose(readyChan, 15*time.Second)
```
feuGeneA, line 180: Is there a reason not to encapsulate this `Wait` logic within `RunRelayerExecutable()`?

tests/e2e_test.go
```diff
@@ -45,8 +63,15 @@ func TestE2E(t *testing.T) {
...
+	var ctx context.Context
+	ctx, cancelDecider = context.WithCancel(context.Background())
+
+	log.Info("Building all ICM off-chain service executables")
+	testUtils.BuildAllExecutables(ctx)
+
+	log.Info("Setting up local network")
 	localNetworkInstance = local.NewLocalNetwork(
-		"awm-relayer-e2e-test",
+		"icm-off-chain-services-e2e-test",
```
feuGeneA, line 74: 🙌

tests/e2e_test.go
```diff
@@ -45,8 +63,15 @@ func TestE2E(t *testing.T) {
 
 // Define the Relayer before and after suite functions.
 var _ = ginkgo.BeforeSuite(func() {
+	var ctx context.Context
+	ctx, cancelDecider = context.WithCancel(context.Background())
```
feuGeneA, line 67: Maybe rename this to just `cancel` now that it can be used to cancel other things too (eg the build process)

tests/e2e_test.go
```diff
@@ -120,16 +135,21 @@ var _ = ginkgo.BeforeSuite(func() {
 		localNetworkInstance.Dir(),
 		ginkgo.ReportEntryVisibilityFailureOrVerbose,
 	)
+
```
feuGeneA, line 138:
```suggestion
```

tests/allowed_addresses.go
iansuvak: We could do it, the main counterexample is if we want to start multiple instances or multiple services like we do in one place for now, this let's us parallelize it. Open to moving though

feuGeneA: I don't feel that strongly about it, but my vote would be to encapsulate it, since it would clean up a lot of the code, and since there's only one case where parallelization is used (though admittedly I could foresee more usages like that).

tests/allowed_addresses.go
```diff
@@ -172,12 +172,19 @@ func AllowedAddresses(network interfaces.LocalNetwork) {
 
 	// Test Relayer 1
 	log.Info("Testing Relayer 1: All sources -> All destinations")
-	relayerCleanup := testUtils.BuildAndRunRelayerExecutable(ctx, relayerConfigPath1)
+	relayerCleanup, readyChan := testUtils.RunRelayerExecutable(
+		ctx,
+		relayerConfigPath1,
+		fmt.Sprintf(
+			"http://localhost:%d/health",
+			relayerConfig1.APIPort,
```
iansuvak, line 180: Not a blocker but since we can assume that all of these are running on localhost I have a slight preference to just pass in the port and handle the URL construction on the other side. That would be assuming that we have the same path for all healthcheck endpoints but I don't think that's too constraining.

tests/e2e_test.go
iansuvak: Actually this file is using the geth logger apparently which is not instantiated but just called directly.

tests/e2e_test.go
iansuvak: Filed #452

tests/e2e_test.go
```diff
@@ -146,7 +138,7 @@ var _ = ginkgo.BeforeSuite(func() {
 
 func cleanup() {
 	if decider != nil {
-		cancelDecider()
+		cancelFn()
```
geoff-vball, line 141: We should lift this `cancelFn` out of this `if` block. It can be called whether or not we have a decider running.

iansuvak: Addressed in 3d7e3c0

bernard-avalabs (approved):

tests/allowed_addresses.go
cam-schultz: A more extensible option would be to pass the config object in its entirety, and grab the relevant fields inside the function.

relayer/main/main.go
feuGeneA: I've integrated the usage of the health endpoints rather than polling stdout, for both the signature-aggregator and the relayer.

tests/utils/utils.go
```diff
@@ -538,3 +490,70 @@ func DeployBatchCrossChainMessenger(
...
+			}
+			time.Sleep(time.Second * 1)
+		}
+	}()
+	return readyChan
+}
+
+// Helper function that waits for a signaling channel to be closed
+// or throws an error if the channel is not closed in time
+func WaitForChannelClose(ch <-chan struct{}, timeout time.Duration) {
```
cam-schultz, line 553: Can we pass a context instead, and leave it up to the caller to set a timeout on it? That's more canonical Go, and makes this function more flexible.

iansuvak: We can, the reason I implemented it this way was to allow for simultaneous starts of multiple services which is what we do in one place so far but I can see it being done in more places in the future.

iansuvak: Sorry I misunderstood what you meant. Yeah I will make the change

cam-schultz (approved):

tests/allowed_addresses.go
```diff
@@ -184,7 +184,8 @@ func AllowedAddresses(network interfaces.LocalNetwork) {
 
 	// Wait for relayer to start up
 	log.Info("Waiting for the relayer to start up")
-	testUtils.WaitForChannelClose(readyChan, 15*time.Second)
+	startupCtx, _ := context.WithTimeout(ctx, 15*time.Second)
```
geoff-vball: Doesn't this only give 15 seconds for us to get past line 312?

geoff-vball: I'm okay with setting one context for the whole test, but we should use a more generous timeout.

geoff-vball: Sorry, posted this on outdated code. It doesn't need to be done in this PR, but I think using `context.WithTimeout(ctx, 15*time.Second)` for every test instead of `context.Background()` and reusing the same context throughout the test would make sense.

iansuvak: > Sorry, posted this on outdated code. It doesn't need to be done in this PR, but I think using `context.WithTimeout(ctx, 15*time.Second)` for every test instead of `context.Background()` and reusing the same context throughout the test would make sense.

I'm not sure I'm parsing your meaning correctly. I am calling context.WithTimeout for each call of the startup. 

You prefer using a top level context for each test that gets passed throughout? That seems more likely to cause flakes and need adjustments later on but I'm ok with that later. Are you ok with me merging this as is?

geoff-vball: So we call `ctx := context.Background()` on line 15, and then create a new context here. I think we should just set one context per test, and give it a reasonable timeout of 1 minute or so. Right now reusing this context in 4 places means we have to get through here to line 312 in 15 seconds, which may or may not be tight.

I'm fine with merging this as-is though

PR #439, opened 2024-08-19
Install buf with go

geoff-vball:
## Why this should be merged
We currently have an issue where the `buf` version needs to be updated in multiple locations, and it is tricky for users to install the correct version.

This PR creates a single source of truth, and removed the need for the install step in the github workflows.


scripts/lint.sh
```diff
@@ -16,4 +16,5 @@ source $RELAYER_PATH/scripts/versions.sh
 go install -v github.com/golangci/golangci-lint/cmd/golangci-lint@${GOLANGCI_LINT_VERSION}
 golangci-lint run --config=$RELAYER_PATH/.golangci.yml --build-tags=testing ./... --timeout 5m
 
+go install github.com/bufbuild/buf/cmd/buf@${BUF_VERSION}
```
feuGeneA, line 19: Maybe add `-v` here like in the previous `go install`?

.github/workflows/e2e.yml
```diff
@@ -49,8 +49,5 @@ jobs:
         with:
           submodules: recursive
 
-      - name: Install buf
-        uses: bufbuild/buf-setup-action@v1.36.0
-
       - name: Run E2E Tests
```
feuGeneA, line 52: Just to confirm my understanding, this actually wasn't being used anymore here, since we removed the buf usage from the build script, right?

go.mod
```diff
@@ -86,19 +86,19 @@ require (
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
 	github.com/pelletier/go-toml/v2 v2.0.8 // indirect
 	github.com/rivo/uniseg v0.2.0 // indirect
-	github.com/rogpeppe/go-internal v1.10.0 // indirect
+	github.com/rogpeppe/go-internal v1.12.0 // indirect
 	github.com/russross/blackfriday/v2 v2.1.0 // indirect
```
feuGeneA, line 90: I'm confused why there are any `go.{mod,sum}` changes in this PR.

feuGeneA (commented):
thanks for this!! a few minor comments but overall excited to see the positive iteration here!

scripts/protobuf_codegen.sh
```diff
@@ -14,13 +14,6 @@ RELAYER_PATH=$(
 
 source $RELAYER_PATH/scripts/versions.sh
 
-## ensure the correct version of "buf" is installed
-BUF_VERSION='1.36.0'
-if [[ $(buf --version | cut -f2 -d' ') != "${BUF_VERSION}" ]]; then
-  echo "could not find buf ${BUF_VERSION}, is it installed + in PATH?"
-  exit 255
-fi
```
cam-schultz, line 22: We should still keep this no? Otherwise calling this script on its own has no guarantee that the code is generated using the expected version.

geoff-vball: I've just added an install step here as well, which is basically a no-op if it's already there.

go.mod
geoff-vball: Yeah, not sure what I did lol

iansuvak (approved):

PR #441, opened 2024-08-20
Use wait teleporter message utility

### LINKED ISSUE #437: Fix E2E test flake in `TriggerProcessMissedBlocks`
**Context and scope**
One of the CI E2E [test runs](https://github.com/ava-labs/awm-relayer/actions/runs/10425191297/job/28875538721?pr=411) for #411 failed with an unexpected error. Re-run succeeded and the failure seemed unrelated to the changes so it was merged anyway. 

We should find the cause of the test failure and fix it. The symptom is that the transaction that was meant to be relayed 

@cam-schultz  pointed out that [TriggerProcessMissedBlocks](https://github.com/ava-labs/awm-relayer/blob/46daa555659afc1f19847745c9b9f43f331e164c/tests/utils/utils.go#L535) depends on the assumption that relayer is the only thing that can trigger block production which might not be true with newer versions and the current tmpnet.  

Logs do show that the relayer sent the message but there is no corresponding [delivered message log](https://github.com/ava-labs/awm-relayer/blob/427b5ef04a7d70afef6ec6dfd9739aba3a2c843e/messages/teleporter/message_handler.go#L349)

cam-schultz:
## Why this should be merged
Fixes #437 
Upgrades to `teleporter@v1.0.5`

## How this works
- Uses the new `WaitTeleporterMessageDelivered` utility rather than waiting for a new block head
- Updates to the new local network creation flow implemented in `teleporter@v1.0.5`
- All hardcoded Teleporter deployment bytes are from `teleporter@v1.0.5`

## How this was tested
CI

## How is this documented
N/A


geoff-vball (approved):

geoff-vball (commented):
We should probably update the teleporter submodule to 1.0.5 as well, just to keep them in sync

iansuvak (approved):

michaelkaplan13 (approved):

PR #442, opened 2024-08-20
add health check endpoint to signature-aggregator

### LINKED ISSUE #424: Health check for signature aggregator
**Context and scope**
The signature aggregation service needs to have a health check endpoint for deployment into a managed service environment.

Example of a no-op health endpoint: https://github.com/ava-labs/teleporter-analytics/blob/main/main/main.go#L65-L80

feuGeneA:
## Why this should be merged

fixes #424 

## How this was tested

by putting a 30 second sleep into the signature-aggregator e2e test and then while it was sleeping did:

```
$ curl localhost:8082/health
{"status":"up","details":{"signature-aggregator":{"status":"up","timestamp":"2024-08-20T01:23:14.351599017Z"}}}
```


geoff-vball (approved):

feuGeneA (commented):
discussed on a call. we should just add a separate endpoint path to the existing `ListenAndServe` call, rather than serving on another port, so that the health check won't start saying "healthy" until the main `ListenAndServe` is actually up and running.

iansuvak (approved):

signature-aggregator/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,19 @@
...
+	"net/http"
+
+	"github.com/alexliesenfeld/health"
+)
+
+func HandleHealthCheckRequest() {
+	healthChecker := health.NewChecker(
+		health.WithCheck(health.Check{
+			Name:  "signature-aggregator",
+			Check: func(context.Context) error { return nil },
```
cam-schultz, line 14: Looks like this is a no-op for now. Can you create a ticket and/or add todo comments here to make this health check meaningful?

feuGeneA: I raised an issue here: https://github.com/ava-labs/awm-relayer/issues/447

But, note, this actually already _is_ meaningful, in that the `http.ListenAndServe()` call made in the aggregator includes both the health check path and also the signature request path, so when this health check starts responding with "healthy", that's an indication that _all_ of the paths served by that `ListenAndServe()` call are ready, so it tells us that the service is ready to handle signature requests.

Issue #448, opened 2024-08-20
Message Coordinator unit tests


cam-schultz:
**Context and scope**
The global message coordinator singleton routes all messages from all blocks to the appropriate application relayer. As such, it is a critical component, and should be thoroughly unit tested,

PR #449, opened 2024-08-20
Only write to listener's app relayers

cam-schultz:
## Why this should be merged
Currently, the message coordinator singleton dispatches to _every_ block from _every_ chain to _every_ application relayer. If there are messages in the block, they will still be processed as expected by the appropriate application relayer, and the other application relayers are very likely to simply no-op. However, if there are no messages in the blocks, then all application relayers will instead stage the height to be committed to the database. The committed heights will be out of order since they don't pertain to the expected chain, so no heights will be written. This breaks block checkpointing, and can cause OOM as the queue of pending blocks grows unbounded.

## How this works
The message coordinator only dispatches to application relayers whose source chain ID matches the block being processed.

## How this was tested
CI. Created https://github.com/ava-labs/awm-relayer/issues/448 to add unit tests for the message coordinator

## How is this documented
N/A

### DIFF
--- a/relayer/listener.go
+++ b/relayer/listener.go
@@ -192,7 +192,12 @@ func (lstnr *Listener) processLogs(ctx context.Context) error {
 				return fmt.Errorf("failed to catch up on historical blocks")
 			}
 		case blockHeader := <-lstnr.Subscriber.Headers():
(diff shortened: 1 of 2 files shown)


iansuvak (approved):
LGTM and thanks for creating the unit test issue, we definitely need it

Issue #443, opened 2024-08-20
Relayer listener cannot catch up on historical blocks after restart and exits


midnight-commit:
**Describe the bug**
My relayer crashes every now and then. Sometimes after running only a couple of hours, sometimes after days. The error is always the same, see logs below. 

**Logs**
{"level":"error","timestamp":"2024-08-14T16:40:20.525Z","logger":"awm-relayer","caller":"evm/subscriber.go:92","msg":"Failed to process block range","error":"websocket: close 1006 (abnormal closure): unexpected EOF"}
{"level":"error","timestamp":"2024-08-14T16:40:20.525Z","logger":"awm-relayer","caller":"relayer/listener.go:262","msg":"Received error from subscribed node","sourceBlockchainID":"yH8D7ThNJkxmtkuv2jgBa4P1Rn3Qpr4pPr7QYNfcdoS6k6HWp","error":"websocket: close 1006 (abnormal closure): unexpected EOF"}
{"level":"info","timestamp":"2024-08-14T16:40:21.300Z","logger":"awm-relayer","caller":"evm/subscriber.go:131","msg":"Successfully subscribed","blockchainID":"yH8D7ThNJkxmtkuv2jgBa4P1Rn3Qpr4pPr7QYNfcdoS6k6HWp"}
{"level":"error","timestamp":"2024-08-14T16:40:21.300Z","logger":"awm-relayer","caller":"relayer/listener.go:207","msg":"Failed to catch up on historical blocks. Exiting listener goroutine.","sourceBlockchainID":"yH8D7ThNJkxmtkuv2jgBa4P1Rn3Qpr4pPr7QYNfcdoS6k6HWp"}
{"level":"info","timestamp":"2024-08-14T16:40:21.300Z","logger":"awm-relayer","caller":"relayer/listener.go:280","msg":"Exiting listener because context cancelled","sourceBlockchainID":"pBBjatuScY9SYsE7YgRQXPJbpQpG1HJLf8uhqWg1oHzNZkrxj"}
{"level":"error","timestamp":"2024-08-14T16:40:21.300Z","logger":"awm-relayer","caller":"main/main.go:303","msg":"Relayer exiting.","error":"failed to catch up on historical blocks"}

**Operating System**
Distributor ID:	Ubuntu
Description:	Ubuntu 24.04 LTS
Release:	24.04
Codename:	noble

Relayer: 1.3.3
AvalancheGo: 1.11.9

**Additional context**
[...]

PR #450, opened 2024-08-21
Listener catch-up fix 

iansuvak:
## Why this should be merged

Addresses #443

## How this works
- Switches catchup from using WS client that is available on the subscriber level to the RPC client used on the listener level. This means that WS issues that subscriber handles by reconnecting should no longer affect the catchup mechanism
- ~~Moves catchup process from subscriber to listener along with the headers channel~~
- Adds simple retries to the network request. 

~~After doing this I realized that this change can be made lighter by passing in an RPC client to the Subscriber as well. That would mean that we can keep the catchup and the headers on the subscriber level.~~
I pivoted to keeping things on the subscriber level and just changing the client and adding retries there

## How this was tested
CI
## How is this documented
N/A

### DIFF
--- a/vms/evm/subscriber.go
+++ b/vms/evm/subscriber.go
@@ -127,8 +127,9 @@ func (s *subscriber) processBlockRange(
 
 func (s *subscriber) getHeaderByNumberRetryable(headerNumber *big.Int) (*types.Header, error) {
 	var err error
+	var header *types.Header
 	for i := 0; i < rpcMaxRetries; i++ {
-		header, err := s.rpcClient.HeaderByNumber(context.Background(), headerNumber)
+		header, err = s.rpcClient.HeaderByNumber(context.Background(), headerNumber)
 		if err == nil {
 			return header, nil
 		}
--- a/vms/evm/subscriber.go
+++ b/vms/evm/subscriber.go
@@ -128,7 +128,8 @@ func (s *subscriber) processBlockRange(
 func (s *subscriber) getHeaderByNumberRetryable(headerNumber *big.Int) (*types.Header, error) {
 	var err error
 	var header *types.Header
-	for i := 0; i < rpcMaxRetries; i++ {
+	attempt := 1
+	for {
 		header, err = s.rpcClient.HeaderByNumber(context.Background(), headerNumber)
 		if err == nil {
 			return header, nil
@@ -136,15 +137,15 @@ func (s *subscriber) getHeaderByNumberRetryable(headerNumber *big.Int) (*types.H
 		s.logger.Warn(
 			"Failed to get header by number",
 			zap.String("blockchainID", s.blockchainID.String()),
-			zap.Int("attempt", i),
+			zap.Int("attempt", attempt),
 			zap.Error(err),
 		)
-		// Sleep if this wasn't the last retry
-		if i < rpcMaxRetries-1 {
-			time.Sleep(time.Duration(i+1) * time.Second)
+		if attempt == rpcMaxRetries {
+			return nil, err
 		}
+		time.Sleep(subscribeRetryTimeout)
+		attempt++
 	}
-	return nil, err
 }
 
 // Loops forever iff maxResubscribeAttempts == 0
--- a/vms/evm/subscriber.go
+++ b/vms/evm/subscriber.go
@@ -22,11 +22,13 @@ const (
 	maxClientSubscriptionBuffer = 20000
 	subscribeRetryTimeout       = 1 * time.Second
 	MaxBlocksPerRequest         = 200
+	rpcMaxRetries               = 5
 )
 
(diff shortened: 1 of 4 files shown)


relayer/catchup_listener.go
```diff
@@ -0,0 +1,95 @@
...
+
+		// clamp to latest known block because we've already subscribed
+		// to new blocks and we don't want to double-process any blocks
+		// created after that subscription but before the determination
+		// of this "latest"
+		if toBlock.Cmp(bigLatestBlockHeight) > 0 {
+			toBlock.Set(bigLatestBlockHeight)
+		}
+
+		for i := 0; i < maxRpcRetries; i++ {
```
iansuvak, line 61: Looking over it I just realized that it would be cleaner if the retries were on the processBlockRange level and closer to the actual network call since failures here would re-trigger reprocessing of block headers that we've already potentially seen

geoff-vball: Yeah, definitely agree here :)

vms/evm/subscriber.go
```diff
@@ -117,6 +125,27 @@ func (s *subscriber) processBlockRange(
 	return nil
 }
 
+func (s *subscriber) getHeaderByNumberRetryable(headerNumber *big.Int) (*types.Header, error) {
+	var err error
+	for i := 0; i < rpcMaxRetries; i++ {
+		header, err := s.rpcClient.HeaderByNumber(context.Background(), headerNumber)
```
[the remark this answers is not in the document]
iansuvak: Thanks for the catch, addressed in daac0d6ebbe9889d06cdc22c3e944d7082cfea2a

vms/evm/subscriber.go
```diff
@@ -117,6 +125,28 @@ func (s *subscriber) processBlockRange(
...
+		}
+		s.logger.Warn(
+			"Failed to get header by number",
+			zap.String("blockchainID", s.blockchainID.String()),
+			zap.Int("attempt", i),
+			zap.Error(err),
+		)
+		// Sleep if this wasn't the last retry
+		if i < rpcMaxRetries-1 {
+			time.Sleep(time.Duration(i+1) * time.Second)
```
bernard-avalabs, line 144: nit: it might be more clear to specify the total timeout/sleep duration (10 seconds across 5 retries) as a constant. Any reason why the subscriber timeout is a constant on each retry, but the rpc timeout grows linearly with each retry?

bernard-avalabs (commented):
Generally looks good to me. Just left a minor comment about timeouts.

geoff-vball (approved):

vms/evm/subscriber.go
iansuvak: I agree but in which direction? constant? Ideally with network resources I like exponential backoffs with jitter but seemed like an overkill here.

cam-schultz: Exponential backoff has come up in previous discussions, but we've come to similar conclusions that it's overkill. Since the overall number of RPC calls is low, optimizing how they're spaced hasn't been a top priority.

cam-schultz: I thought we were tracking that already, but apparently not. I created https://github.com/ava-labs/awm-relayer/issues/453 to do so.

vms/evm/subscriber.go
```diff
@@ -128,23 +128,24 @@ func (s *subscriber) processBlockRange(
...
 			"Failed to get header by number",
 			zap.String("blockchainID", s.blockchainID.String()),
-			zap.Int("attempt", i),
+			zap.Int("attempt", attempt),
 			zap.Error(err),
 		)
-		// Sleep if this wasn't the last retry
-		if i < rpcMaxRetries-1 {
-			time.Sleep(time.Duration(i+1) * time.Second)
+		if attempt == rpcMaxRetries {
```
cam-schultz, line 143:
```suggestion
		if attempt >= rpcMaxRetries {
```
To cover the edge case where `rpcMaxRetries == 0`. This is very unlikely to happen since it's hardcoded, so this is mostly defensive.

vms/evm/subscriber.go
```diff
@@ -117,6 +125,29 @@ func (s *subscriber) processBlockRange(
 	return nil
 }
 
+func (s *subscriber) getHeaderByNumberRetryable(headerNumber *big.Int) (*types.Header, error) {
+	var err error
+	var header *types.Header
+	attempt := 1
+	for {
```
minghinmatthewlam, line 132: why don't we do a regular for attempt=1; attempt < retryAttempts; attempt ++ here

iansuvak: matching the way that subscriber reattempts right below this

cam-schultz (approved):

geoff-vball (approved):

PR #457, opened 2024-08-22
Initial signature aggregator unit tests

### LINKED ISSUE #446: Add unit tests for signature aggregator
**Context and scope**
Currently we are relying only on E2E test for testing functionality throughout the repo. With separating out signature-aggregator and when doing #388 we should add unit tests to ensure that validity checks, aggregation logic etc are valid.

feuGeneA:
## Why this should be merged
it's a first step towards fixing #446


signature-aggregator/aggregator/aggregator.go
```diff
@@ -343,6 +343,8 @@ func (s *SignatureAggregator) CreateSignedMessage(
 	return nil, errNotEnoughSignatures
 }
 
+// TODO: consider making this function private. its only reference seems to be
```
iansuvak, line 346: Yup both of the below should be private. That was my oversight in the move

feuGeneA: addressed in https://github.com/ava-labs/awm-relayer/pull/457/commits/e5d44b441029b356278d6d13549ff6c4fdb89291

signature-aggregator/aggregator/aggregator_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/ava-labs/awm-relayer/signature-aggregator/metrics"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+)
+
+var sigAggMetrics *metrics.SignatureAggregatorMetrics
+var messageCreator message.Creator
+
+func instantiateAggregator(t *testing.T) (
```
geoff-vball, line 22: Should we create a happy-path test here? Otherwise this PR LGTM

.github/workflows/mock_checker.yml
```diff
@@ -0,0 +1,35 @@
+name: Check generated code is up to date
```
minghinmatthewlam, line 1:
```suggestion
name: Mock Checker
```

peers/app_request_network.go
```diff
@@ -1,6 +1,8 @@
 // Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
 
+//go:generate mockgen -source=$GOFILE -destination=./mocks/mock_app_request_network.go -package=mocks
```
minghinmatthewlam, line 4: why is this included in the source file? I would expect the source file to not need to have knowledge of the mock file at all. Instead, can the mock file keep track of command needed to regenerate?

feuGeneA: it's here because the mock is generated from this source file, which is represented on this line by `$GOFILE`. btw i took this from your example 😅 at https://github.com/ava-labs/awm-relayer/pull/21/files#diff-b7c4d6477bfb555408745965840c387dcfc91a74bcbdc3b30d31d594e9abc0e5R4

my understanding is that the mock source file is 100% generated so we can't really put anything in there.

signature-aggregator/aggregator/aggregator_test.go
feuGeneA: I think yes, but I'd prefer to defer it to a subsequent PR. I'm actively working on it but would like to see this PR get merged since it's already quite close.

geoff-vball (approved):

iansuvak (commented):
Generally LGTM but I do prefer `Mock Checker` that @minghinmatthewlam suggested as the name visible on the UI rather than the more verbose one, it also follows the pattern of `ABI Go Bindings Checker`.

scripts/test.sh
```diff
@@ -27,12 +27,14 @@ if [ "$HELP" = true ]; then
...
+root=$(
     cd "$(dirname "${BASH_SOURCE[0]}")"
     cd .. && pwd
 )
-source "$RELAYER_PATH"/scripts/constants.sh
+source "$root"/scripts/constants.sh
 
 go build -o tests/cmd/decider/decider ./tests/cmd/decider/
 
+"$root"/scripts/generate.sh
```
iansuvak, line 38: I would still rather call this manually than have it be re-generated automatically ever. If it's quick it might not make too much of a difference. not a blocker though

iansuvak (approved):

scripts/test.sh
cam-schultz: I agree. I wouldn't ever expect running a test to alter my git status

PR #461, opened 2024-08-27
add metrics for connect/disconnect events

iansuvak:
## Why this should be merged

Instead of moving Connected/Disconnected log events back to info level this collects metrics on the number of connect and disconnect events that occur. As of now, the applications don't initiate any disconnects so alerts can be created to monitor for excessive disconnects initiated by network peers. 

Also fixed the typo in the `app_request_network_metrics.go` filename

## How this works

I re-used recently added `AppRequestNetworkMetrics` struct since these are fundamentally still related to the `AppRequestNetwork`.

### DIFF
--- a/peers/external_handler.go
+++ b/peers/external_handler.go
@@ -43,7 +43,6 @@ type expectedResponses struct {
 func NewRelayerExternalHandler(
 	logger logging.Logger,
 	metrics *AppRequestNetworkMetrics,
-	registerer prometheus.Registerer,
 ) (*RelayerExternalHandler, error) {
 	// TODO: Leaving this static for now, but we may want to have this as a config option
 	cfg := timer.AdaptiveTimeoutConfig{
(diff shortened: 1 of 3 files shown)


geoff-vball (approved):

cam-schultz (approved):

peers/external_handler.go
```diff
@@ -41,6 +42,7 @@ type expectedResponses struct {
 // Teleporter application relayer, as well as handle timeouts.
 func NewRelayerExternalHandler(
 	logger logging.Logger,
+	metrics *AppRequestNetworkMetrics,
 	registerer prometheus.Registerer,
```
cam-schultz, line 46: It's odd that we pass in an actual metrics collector alongside an independent registerer (that we have hooked up to the noop registerer). Instead, how about we remove `registerer` as a parameter altogether, and hardcode `prometheus.DefaultRegisterer` in the call to `NewAdaptiveTimeoutManager`. We could also pass in the timeout manager as an arg.

iansuvak: I agree, I thought it was odd but not that big of a deal. I went with the first option above in [e047507](https://github.com/ava-labs/awm-relayer/pull/461/commits/e047507aef1396b48719d1e047436dc082105482)

cam-schultz (approved):

bernard-avalabs (approved):
LGTM

minghinmatthewlam (approved):

Issue #383, opened 2024-07-23
Use ACP-118 interface


iansuvak:
**Context and scope**
Once [ACP-118](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/118-warp-signature-request) has been implemented we need to switch our `AppRequests` to use the new format as well as expose the `justification` field in the `raw-message` interface 

**Discussion and alternatives**

**Open questions**

michaelkaplan13:
This should now be unblocked with the Subnet EVM implementation of the ACP-118 handlers added in this PR: https://github.com/ava-labs/subnet-evm/pull/1260/files

PR #463, opened 2024-08-28
Add `etna-time` configs

iansuvak:
## Why this should be merged
This will allow for merging of ACP-118 changes and toggling them based on the configured timestamp of the Etna upgrade

## How this works
- New config option uses Viper as before but uses it's `GetTime()` method directly instead of relying on `mapstructure` tags since `time.Time` isn't directly handled by it.
- Fixes up some documentation in the sample workflow section. 

## How this was tested
- New simple E2E test was written and confirmed that both pre and post-etna cases are functioning
- Ran a manual build of signature-aggregator against Fuji testnet with the `etna-time` configuration set in the past, unset and set in the future. As expected, only future and unset cases returned a valid signature. 
- Ran a manual build of signature-aggregator on #383 branch and confirmed that doesn't work either as expected. 
## How is this documented


relayer/config/config.go
```diff
@@ -67,6 +68,9 @@ type Config struct {
 	DeciderURL             string                   `mapstructure:"decider-url" json:"decider-url"`
 	SignatureCacheSize     uint64                   `mapstructure:"signature-cache-size" json:"signature-cache-size"`
 
+	// mapstructure doesn't handle time.Time out of the box so handle it manually
+	EtnaTime time.Time `json:"etna-time"`
```
iansuvak, line 72: I would have liked to do `omitempty` so that tests don't write the field out when marshaling JSON here but zero time value still gets written out unless the field type is `*time.Time` which seemed messier.

The value that does get written out is correctly parsed by viper to be zero time value.  As a result tests run the pre-Etna case unless otherwise specified.

signature-aggregator/aggregator/aggregator.go
```diff
@@ -594,3 +593,59 @@ func (s *SignatureAggregator) aggregateSignatures(
...
+		} else {
+			req := msg.MessageSignatureRequest{
+				MessageID: unsignedMessage.ID(),
+			}
+			return msg.RequestToBytes(codec, req)
+		}
+	}
+}
+
+func (s *SignatureAggregator) unmarshalResponse(responseBytes []byte) (blsSignatureBuf, error) {
```
geoff-vball, line 633: Can we leave TODOs and create a ticket for when this logic is okay to clean up?

geoff-vball (approved):
One comment but LGTM 👍

PR #469, opened 2024-09-03
Update docs with return payload

iansuvak:
## Why this should be merged

Returns error status codes and adds response documentation to the signature aggregator API

## How is this documented


geoff-vball (approved):

michaelkaplan13:
We have previously gone with the standard of leaving HTTP status codes to any network level issues/infra, and always returning 200 with an error message for any application level errors. To stay consistent, I'd recommend we do that here as well to match other platform APIs

iansuvak:
> We have previously gone with the standard of leaving HTTP status codes to any network level issues/infra, and always returning 200 with an error message for any application level errors. To stay consistent, I'd recommend we do that here as well to match other platform APIs

Sure thing -- I'll remove references but I am curious about the reasoning here. In general I think that there is value in distinguishing between client caused failures and internal errors

geoff-vball (approved):

cam-schultz (approved):

bernard-avalabs (approved):

PR #470, opened 2024-09-04
Update broken image links

iFrostizz:
## Why this should be merged

Images from the relayer readme were pointing to a wrong location.

## How this works

Update links

## How this was tested

Using a markdown visualisation plugin

## How is this documented


geoff-vball (approved):
Thanks for the fix :)

iansuvak (approved):
Thanks for the catch -- that was my bad =(

PR #486, opened 2024-09-11
Fix connected weight calculation

michaelkaplan13:
## Why this should be merged
It looks like the way that the relayer was previously calculating the `connectedWeight` is incorrect. The `Weight` for each validator returned by `GetCurrentCanonicalValidatorSet` is the total weight for all node IDs with a given BLS key. When summing the weight by iterating over connected node IDs then, we would be double counting in any situation where multiple node IDs were using the same BLS key.

## How this works
Checks if a node ID with the given BLS key have been accounted for already.

## How this was tested
TODO

## How is this documented
N/A


iansuvak (approved):
Good catch and LGTM

geoff-vball (approved):

michaelkaplan13:
@iansuvak thank you for looping back to add the tests here 🙏

PR #506, opened 2024-10-02
Validators at height

### LINKED ISSUE #501: Construct signatures for a specific p-chain height
**Context and scope**
`CreateSignedMessage` right now assumes that the ProposerVM P-Chain height for the destination chain is the latest height. This is not a valid assumption at all times and it should optionally accept a specific height for which to create a signature since any changes in the signing subnets validator set as viewed by the destination chain (via p-chain) can make the aggregated signature invalid as the validator bit set would change in those cases. 

- Should we expose this in the Signature aggregator interface? 
- Ideally we should be able to accept a destination chain id on the api side and translate that to a likely valid p-chain height. 

**Open questions**

iansuvak:
## Why this should be merged
Closes #501 as well as adds the proposerHeightCache to get a best guess height for which to request validators. 
## How this works

- Accepts pchain height as an argument to createSignedMessage. 
- Adds `proposerHeightCache` that updates itself on a set interval 

## How this was tested

- Adds new unit tests for `proposerHeightCache`
- existing tests should pass

## How is this documented
Comments


bernard-avalabs (commented):
A few minor questions mostly to improve my own understanding.

peers/validators/canonical_validator_client.go
```diff
@@ -36,18 +36,10 @@ func NewCanonicalValidatorClient(logger logging.Logger, apiConfig *config.APICon
 	}
 }
 
-func (v *CanonicalValidatorClient) GetCurrentCanonicalValidatorSet(
+func (v *CanonicalValidatorClient) GetCanonicalValidatorSet(
```
michaelkaplan13, line 39: This looks like a pure passthrough to `avalancheWarp.GetCanonicalValidatorSet` now. Thoughts on removing it enitrely?

relayer/application_relayer.go
```diff
@@ -207,6 +207,7 @@ func (r *ApplicationRelayer) ProcessMessage(handler messages.MessageHandler) (co
 			nil,
 			r.signingSubnetID,
 			r.warpQuorum.QuorumNumerator,
+			0,
```
michaelkaplan13, line 210: Lets leave a TODO here to get the estimated P-Chain height to be used, if we're planning on merging this prior to adding that estimation.

geoff-vball (commented):
I didn't look over the cache implementation too hard, but looks good. The rest LGTM aside from existing comments

signature-aggregator/aggregator/proposer_height_cache.go
```diff
@@ -0,0 +1,169 @@
...
+	errFailedToGetCurrentHeight = errors.New("failed to get current P-chain height")
+)
+
+const pChainLookback = 30 * time.Second
+
+type ProposerHeightCache struct {
+	logger       logging.Logger
+	pChainClient *validators.CanonicalValidatorClient
+	// protected by timeToHeightLock
+	timeToHeight     *linked.Hashmap[time.Time, uint64]
```
iansuvak, line 31: This doesn't need to be a hashmap. Should look into a dequeue.

peers/app_request_network.go
```diff
@@ -222,14 +222,31 @@ func (c *ConnectedCanonicalValidators) GetValidator(nodeID ids.NodeID) (*warp.Va
 
 // ConnectToCanonicalValidators connects to the canonical validators of the given subnet and returns the connected
 // validator information
-func (n *appRequestNetwork) ConnectToCanonicalValidators(subnetID ids.ID) (*ConnectedCanonicalValidators, error) {
-	// Get the subnet's current canonical validator set
+func (n *appRequestNetwork) ConnectToCanonicalValidators(
+	height uint64,
+	subnetID ids.ID,
+) (*ConnectedCanonicalValidators, error) {
+	var err error
```
minghinmatthewlam, line 229: no need to set a new variable here instead of usual `err :=` assignment, will only pass line 231 if err = nil.

iansuvak: I'm overriding the passed in height if it's 0 so I do need the `err` defined prior to that so that line 232 is pure assignment to already declared vars. 

I could declare a new variable to replace `height` instead but I think this is cleaner

Also hi 😄

michaelkaplan13:
With ACP-151 now being implemented, do we want to close this PR in favor of adding a `GetProposedValidators` RPC method to the P-Chain in AvalancheGo and leveraging that instead of `GetCurrentValidators` or `GetValidatorsAtHeight`?

I think this iteration of the design was super helpful to think through and conceptualize, and that it led us to design that both simpler and more robust.

iansuvak:
Closing in favor of #555

PR #507, opened 2024-10-03
bump avalanchego and update uptime interface

iansuvak:
## Why this should be merged
It shouldn't be merged until 1.12.0 is merged in. Pushed here as a convenience for developing against it. 
## How is this documented


iansuvak:
These have been merged to main in #530

PR #519, opened 2024-10-10
Add extra peer endpoints to aggregator

felipemadero:
## Why this should be merged
Certain devnet endpoints does not correctly return all external nodes connected to the devnet validators. 
As is currently the case with etna. The PR enables passing to the aggregator nodes that are not included
in canonical info.peers from the devnet endpoint.

## How this works
It enables passing to the aggregator a list of endpoints from which to generate extra peers.

## How this was tested
CLI branch that connects a local node to etna and needs the aggregator to include it in the aggregation 

## How is this documented

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,9 +1,9 @@
 module github.com/ava-labs/awm-relayer
 
-go 1.22.7
+go 1.22.8
 
 require (
-	github.com/ava-labs/avalanchego v1.11.11
+	github.com/ava-labs/avalanchego v1.12.0-initial-poc.0
 	github.com/ava-labs/coreth v0.13.8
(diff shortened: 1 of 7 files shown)


felipemadero:
> Should we convert this to a draft for now? I think that this is okay to use for testing in this instance but shouldn't canonical info reliably return the way to connect to validators in the long term?

connecting an instance to a devnet is very much needed, not for testing on an instance, but to support
acp77 blockchain generation flow. That depends on how for example etna endpoint manages the info.peers api call.

Anyway I believe it is good idea to have this as an option.

iansuvak:
closed since superseded by #521 which is merged

PR #537, opened 2024-10-31
use warpRequirePrimaryNetworkSigners config

### LINKED ISSUE #454: Add config switch for primary network validator signature verification optimization
**Context and scope**
subnet-evm and coreth support an optimization for verifying Warp messages from the primary network in which only the validators of the receiving subnet need to sign the message. For example, this allows a subnet to "self sign" a message from the C-Chain, rather than having to verify an aggregate signature consisting of 1000's of individual BLS signatures.

ACP-77 removes primary network validation as a subnet validation requirement, so such "self signing" is no longer guaranteed to be meaningful. As such, subnet-evm and coreth are making this behaviour [configurable](https://github.com/ava-labs/subnet-evm/pull/1232) at the Warp precompile level.

We'll need to support both signature aggregation methods. This new config option will be published in the chain config, so we can fetch it in the same way we fetch the Warp quorum configuration currently. When aggregating signatures for messages originating from the primary network, we'll need to take into account the destination's signature verification rules.

iansuvak:
## Why this should be merged
Fixes #454 
## How this works
Fetches the additional value from the config and uses it when creating application relayer struct.  

Open q: `WarpQuorum` struct doesn't seem very useful since denominator is hardcoded at compile time to 100 
## How this was tested
Existing tests should pass
TODO: add new ones

## How is this documented

### DIFF
--- a/relayer/application_relayer.go
+++ b/relayer/application_relayer.go
@@ -88,10 +88,21 @@ func NewApplicationRelayer(
 		)
 		return nil, err
 	}
+	requirePrimaryNetworkSigners, err := cfg.GetWarpRequirePrimaryNetworkSigners(relayerID.DestinationBlockchainID)
+	// this shouldn't be reachable since if we found a quorum, we should also find the requirePrimaryNetworkSigners
+	// but leaving the check for completeness
+	if err != nil {
+		logger.Error(
+			"Failed to get warp primary network configuration from config.",
+			zap.String("destinationBlockchainID", relayerID.DestinationBlockchainID.String()),
+			zap.Error(err),
+		)
+		return nil, err
+	}
 	var signingSubnet ids.ID
-	if sourceBlockchain.GetSubnetID() == constants.PrimaryNetworkID {
-		// If the message originates from the primary subnet, then we instead "self sign"
-		// the message using the validators of the destination subnet.
+	if sourceBlockchain.GetSubnetID() == constants.PrimaryNetworkID && !requirePrimaryNetworkSigners {
+		// If the message originates from the primary subnet, and the primary network is validated by
+		// the destination change we can "self-sign" the message using the validators of the destination subnet.
 		signingSubnet = cfg.GetSubnetID(relayerID.DestinationBlockchainID)
 	} else {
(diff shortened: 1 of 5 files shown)


relayer/config/config.go
```diff
@@ -256,6 +243,15 @@ func (c *Config) GetWarpQuorum(blockchainID ids.ID) (WarpQuorum, error) {
 	return WarpQuorum{}, errFailedToGetWarpQuorum
 }
 
+func (c *Config) GetWarpRequirePrimaryNetworkSigners(blockchainID ids.ID) (bool, error) {
+	for _, s := range c.DestinationBlockchains {
+		if blockchainID.String() == s.BlockchainID {
+			return s.warpRequirePrimaryNetworkSigners, nil
+		}
+	}
```
geoff-vball, line 251: For the number of subnets we have for now, doing a linear search is fine, but it does kind of beg the question if we should have the blockchains in a map keyed by the `BlockchainID`. If it's a big extra lift because that would be too hard to parse straight from the config, that's fine.

iansuvak: Yeah it's definitely not worth changing IMO for a one-time config parsing.

relayer/application_relayer.go
```diff
@@ -88,10 +88,21 @@ func NewApplicationRelayer(
 		)
 		return nil, err
 	}
+	requirePrimaryNetworkSigners, err := cfg.GetWarpRequirePrimaryNetworkSigners(relayerID.DestinationBlockchainID)
+	// this shouldn't be reachable since if we found a quorum, we should also find the requirePrimaryNetworkSigners
+	// but leaving the check for completeness
```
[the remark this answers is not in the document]
iansuvak: I went with the latter route and renamed it to `WarpConfig`. It's basically identical to the source `WarpConfig` defined in `subnet-evm` but doesn't implement the `Upgrade` interface and when converting it we populate it with `warpDefaultQuorumNumerator` value if it's 0

geoff-vball (approved):
🚀

cam-schultz (approved):

PR #551, opened 2024-11-18
Rename away from warp

### LINKED ISSUE #538: Rename repo and artifacts
**Context and scope**
Rename the repo and update code, documentation and artifacts to use the new naming. 
`awm-relayer` should be `icm-relayer` throughout

**Open questions**
Should the repo should either be `icm-offchain-services` or `icm-relayer`? current top level readme is titled `Avalanche ICM Off-chain Services`. If we don't want to commit to that name we can just use `icm-relayer` and update the README to reflect it

iansuvak:
## Why this should be merged

Part of general renaming from Warp -> ICM. Closes #538

## How this works

- Renames the repo and git module `awm-relayer` -> `icm-offchain-services`
- Renames the `awm-relayer` binary and expected DockerHub repo -> `icm-relayer`
- Updates links and titles `AWM Relayer` referring to the repo in general is `ICM Off-chain Services` in capitalized titles

One inconsistency right now is module/repo name having a dash in `Off-chain` in plain text but no dash in the module and proposed repo name. I can change either to make it consistent but `icm-off-chain-services` seems like too many dashes for a module/repo name IMO. 

## How this was tested
Existing tests should pass. Once the repo is created it should be tested with new test tags. 
## How is this documented
Updated the docs


geoff-vball:
Just as a note before I start reviewing. I do like `icm-offchain-services` and `ICM Off-chain Services`. 3 dashes is too many

iansuvak:
> Thoughts on condensing the name to `icm-service`? IMO the shorter the repo name the better. I think it should be pretty obvious from the lack of smart contracts that these services are not "on chain", so the "off chain distinction" feels superfluous. In the wider context of the soon-to-be-named `icm-contracts` repo, having this repo be named similarly emphasizes their relationship.

I like `icm-services`

geoff-vball (approved):

scripts/build_local_image.sh
```diff
@@ -20,12 +20,12 @@ full_commit_hash="$(git --git-dir="$RELAYER_PATH/.git" rev-parse HEAD)"
...
 
 echo "Building Docker Image with tags: $relayer_dockerhub_repo:$commit_hash , $relayer_dockerhub_repo:$current_branch"
 docker build -t "$relayer_dockerhub_repo:$commit_hash" \
         -t "$relayer_dockerhub_repo:$current_branch" \
         "$RELAYER_PATH" -f "$RELAYER_PATH/Dockerfile" \
         --build-arg GO_VERSION=$GO_VERSION
 
-rm $RELAYER_PATH/awm-relayer
\ No newline at end of file
+rm $RELAYER_PATH/icm-relayer
```
richardpringle, line 31: Can we get a new line here, please?

richardpringle (approved):
One comment, but okay to do in a follow up

Issue #534, opened 2024-10-29
Fetch peers over P2P


cam-schultz:
**Context and scope**
With Etna, primary network API nodes will no longer be expected to be connected to all L1 validators. `info.Peers` will return peers only for tracked subnets. Rather than requiring Info API nodes for all tracked subnets, we can instead fetch the global peer list over p2p. https://github.com/ava-labs/avalanchego/pull/3491 relaxes the `p2p.GetPeers` constraints such that any primary network validator will return the list of peers across all L1s to non-primary network validator nodes. We can use this in the signature aggregator to fetch the global list of peers, and manually connect to each.

**Discussion and alternatives**
Peers can be fetched over p2p using the [p2p.GetPeerList](https://github.com/ava-labs/avalanchego/blob/master/proto/p2p/p2p.proto#L29) message type. We need to ensure that we are requesting this from a primary network validator. We can do this by calling `info.Peers` on a primary network API node, which matches the current API requirements.

**Open questions**
Is manually connecting to each of the peers the preferred approach? Currently, we connect to peers like so:
- Fetch the canonical validator set of an L1
- Fetch the global peer list (previously via `info.Peers`, with this change via `p2p.GetPeerList`)
- Manually connect to the canonical validator set using the IP addresses from the previous step

Instead, we should be able to transitively connect to all L1 peers by manually connecting to a subset of the L1's validators. See the example [here](https://github.com/ava-labs/avalanchego/blob/cd0c6e152965f44a2794935c41881fb207c8f78c/network/example_test.go#L110). We may be able to instead maintain an in-memory list of connected/tracked subnets that we bootstrap connections to only once. This would likely be a larger change, but could simplify peer discovery overall. Some additional research is required here, especially given potential changes to subnet tracking semantics in the Avalanchego networking code.

perfilev-faraway:
We're waiting for this feature. Current our workaround is using manuallyTrackedPeers, but it doesn't work with PoS chains when new validators can be added permissionless.

The patch below allows us to connect only to this bootstrap validator. The signature aggregator observes new validators correctly but unable to connect to them.

```
diff --git a/signature-aggregator/main/main.go b/signature-aggregator/main/main.go
index 829f860..3416aa9 100644
--- a/signature-aggregator/main/main.go
+++ b/signature-aggregator/main/main.go
@@ -4,12 +4,14 @@
 package main
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"log"
[...]

PR #555, opened 2024-11-19
Validators at proposed

iansuvak:
## Why this should be merged

Supersedes #506 

## How this works

Removes fallback calls  to `getCurrentValidators` and replaces height passed into `getValidatorsAt` from the result of  `GetCurrentHeight` to `"proposed"` flag. 

## How this was tested
Existing tests should pass. These tests fail against previous version of `avalanchego` with the code modifications here. 

## How is this documented


peers/validators/canonical_validator_client.go
```diff
@@ -98,77 +91,14 @@ func (v *CanonicalValidatorClient) GetValidatorSet(
 ) (map[ids.NodeID]*validators.GetValidatorOutput, error) {
```
bernard-avalabs, line 91: Should the comment for this function be updated to remove the sentence about falling back to getCurrentValidators?

bernard-avalabs (commented):
Makes sense to me assuming that getValidatorsAt(height: "proposed") is available.

peers/validators/canonical_validator_client.go
```diff
@@ -89,86 +81,21 @@ func (v *CanonicalValidatorClient) GetCurrentValidatorSet(
 }
 
 // Gets the validator set of the given subnet at the given P-chain block height.
-// Attempts to use the "getValidatorsAt" API first. If not available, falls back
-// to use "getCurrentValidators", ignoring the specified P-chain block height.
+// Uses [platform.getValidatorsAt] with "proposed" height to get the validator set that
+// will likely be used for the next block.
```
cam-schultz, line 85: This comment doesn't match the implementation, which uses the passed in height

go.mod
```diff
@@ -3,9 +3,9 @@ module github.com/ava-labs/awm-relayer
 go 1.22.8
 
 require (
-	github.com/ava-labs/avalanchego v1.12.0-initial-poc.9
-	github.com/ava-labs/coreth v0.13.8
-	github.com/ava-labs/subnet-evm v0.6.11
+	github.com/ava-labs/avalanchego v1.11.13
```
cam-schultz, line 6: Don't we need to continue using the Etna POC branches for tests to pass? Or is that the case only for the Validator Manager e2e tests in `teleporter`?

peers/validators/canonical_validator_client.go
iansuvak: Good catch! Will update

go.mod
iansuvak: It should just be validator manager tests but I can bump to v1.12.0-fuji as well.

iansuvak: The failure here is because `0.6.12` subnet-evm specifies `v1.11.13` so the peers disconnect in E2E tests because of version incompatibility between `v1.12.0-fuji` and `v1.11.13`

geoff-vball (commented):
LGTM other than what others have commented

go.mod
iansuvak: Made the fix to use locally defined version of `avalanchego` in #557

geoff-vball (approved):
Don't know if we have a testing plan yet? But the code as-is LGTM

iansuvak:
> Don't know if we have a testing plan yet? But the code as-is LGTM

This PR would fail E2E if it wasn't working at all.  I'll add an explicit E2E test that relies on proposed height != current p-chain height in a separate PR once a PR resolving #534 is up since that one will bring over latest tmpnet changes over from teleporter

richardpringle (approved):
LGTM, relying of @geoff-vball's review for the deleted code as I don't have the context there.

bernard-avalabs (approved):

PR #557, opened 2024-11-20
use local copy of install_avalanchego_release.sh

iansuvak:
## Why this should be merged

Allows for easier custom avalanchego vs subnet-evm versions assuming they are compatible with each other

## How this works

Updates both the github e2e as well as `./scripts/e2e_test.sh` runner to use avalanchego version from `./scripts/versions.sh` instead of using the version from `subnet-evm` repo. 

## How this was tested

Existing tests should pass. 

## How is this documented
Need to update README.md for the simplified flow

### DIFF
--- a/scripts/install_avalanchego_release.sh
+++ b/scripts/install_avalanchego_release.sh
@@ -1,0 +1,133 @@
+#!/usr/bin/env bash
+# Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+# See the file LICENSE for licensing terms.
+
+# Following script is adapted from https://github.com/ava-labs/subnet-evm/blob/master/scripts/install_avalanchego_release.sh
+set -e
+
+# Load the versions
+ICM_OFFCHAIN_CONTRACTS_PATH=$(
+  cd "$(dirname "${BASH_SOURCE[0]}")"
+  cd .. && pwd
+)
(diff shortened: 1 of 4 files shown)


scripts/install_avalanchego_release.sh
```diff
@@ -0,0 +1,133 @@
+#!/usr/bin/env bash
```
geoff-vball, line 1: We once had this copied over from `subnet-evm`, but got rid of it for some reason. Are we not able to use the file directly from our submodule while setting some env vars? Would it possibly be worthwhile to make a PR to `subnet-evm` to make it usable if not? Just spitballing :)

iansuvak: We use it like this in teleporter. I think scripts like this are reasonable to have copy pastad in individual repos IMO  and we do it already with other scripts. Happy to be overridden though.

geoff-vball: Oh, lol forgot which repo I was in. Can we use the one from the teleporter submodule? I don't mind copying code if its the best/easiest option, just wondering if we can piggyback somewhere here

iansuvak: each of them uses their own `versions.sh` we certainly can but then we end up with the same problem of coupling repos tighter than they need to be

geoff-vball: Yeah, that's fair, you've convinced me 😅

PR #559, opened 2024-11-21
Fetch peers p2p

cam-schultz:
## Why this should be merged
Connects to peers over p2p, rather than via the info API. Post Etna, info API nodes are not expected to track peers for all subnets, so we instead need to fetch peers directly from primary network validators, which _do_ track all peers across all L1s. We do this by bootstrapping the network with connections to primary network validators (which _are_ still fetched from the info API), and discovering peers via peer gossip. Peer gossip only tracks peers that correspond to known validators, so we also need to maintain an up-to-date view of the validators of the L1s we wish to connect to (including the primary network).  This change leverages https://github.com/ava-labs/teleporter/pull/641 to confirm that we can connect to L1s with disjoint validator sets from the primary network. 

**Note:** This PR uses a modified version of Avalanchego v1.12.0-fuji that relaxes the outbound peer request handling. Once this change is formalized and included in an Avalanchego release, we will circle back and update the dependency. This workaround is safe because it only impacts outbound messages from the signature aggregator peer.

## How this works
- Create a validator manager and initialize it with the current validators of the primary network and any tracked L1s
  - Periodically update the validator manager with the latest validator sets from the P-Chain API
  - For the signature aggregator service, add new tracked subnets on demand
- Bootstrap the network with primary network validator peers (fetched from the Info API), which will gossip any known validators across all tracked L1s
- There's no need to manually connect to peers anymore, as this is handled automatically via peer gossip

## How this was tested
CI

## How is this documented
N/A


peers/app_request_network.go
```diff
@@ -50,21 +54,21 @@ type AppRequestNetwork interface {
...
 	logger          logging.Logger
-	lock            *sync.Mutex
+	lock            *sync.RWMutex
 	validatorClient *validators.CanonicalValidatorClient
 	metrics         *AppRequestNetworkMetrics
 
-	// Nodes that we should connect to that are not publicly discoverable.
-	// Should only be used for local or custom blockchains where validators are not
-	// publicly discoverable by primary network nodes.
-	manuallyTrackedPeers []info.Peer
```
iansuvak, line 67: 👍 I like removing this from the state while keeping the functionality on `NewNetwork` initialization.

peers/external_handler.go
```diff
@@ -101,7 +101,7 @@ func (h *RelayerExternalHandler) HandleInbound(_ context.Context, inboundMessage
 }
 
 func (h *RelayerExternalHandler) Connected(nodeID ids.NodeID, version *version.Application, subnetID ids.ID) {
-	h.log.Debug(
+	h.log.Info(
```
iansuvak, line 104: (not a blocker) are we keeping this at `Info` level or was this just for testing? I'd prefer having both `Connected` and `Disconnected` events at the same log level in general. I guess we can match them on the metrics side but if diagnosing anything network related it's nice to also know the interval .

tests/basic_relay.go
```diff
@@ -63,15 +64,15 @@ func BasicRelay(network *network.LocalNetwork, teleporter utils.TeleporterTestIn
 	log.Info("Test Relaying from Subnet A to Subnet B")
 
 	log.Info("Starting the relayer")
-	relayerCleanup, readyChan := testUtils.RunRelayerExecutable(
+	relayerCleanup, readyChan /*readyChan*/ := testUtils.RunRelayerExecutable(
```
iansuvak, line 67: might be a dumb question but what's the point of the in-line comment here?

tests/basic_relay.go
```diff
@@ -85,6 +86,8 @@ func BasicRelay(network *network.LocalNetwork, teleporter utils.TeleporterTestIn
 		fundedAddress,
 	)
 
+	time.Sleep(10 * time.Second)
```
iansuvak, line 89: why this custom sleep? `RelayBasicMessage` should be blocking right?

iansuvak (commented):
Generally LGTM

peers/app_request_network.go
```diff
@@ -126,95 +133,155 @@ func NewNetwork(
...
+			zap.Error(err),
+		)
+		return nil, err
 	}
+	pClient := platformvm.NewClient(cfg.GetPChainAPI().BaseURL)
+	numConnected := 0
+	for _, peer := range peers {
+		vdrs, err := pClient.GetCurrentValidators(context.Background(), constants.PrimaryNetworkID, []ids.NodeID{peer.ID})
+		if err != nil {
+			panic(err)
```
geoff-vball, line 159: We should probably return an error here to be consistent with the rest of this function

peers/app_request_network.go
```diff
@@ -126,95 +133,155 @@ func NewNetwork(
...
+		logger.Info(
+			"Manually tracking bootstrap node",
+			zap.String("ID", peer.ID.String()),
+			zap.String("IP", peer.PublicIP.String()),
+		)
+		testNetwork.ManuallyTrack(peer.ID, peer.PublicIP)
+		numConnected++
+		if numConnected >= NumBootstrapNodes {
+			break
+		}
```
geoff-vball, line 174: Should we log something if we exit the `for` loop without connecting to `NumBootstrapNodes`?

config/peer_config.go
```diff
@@ -0,0 +1,44 @@
...
+
+type PeerConfig struct {
+	ID string `mapstructure:"id" json:"id"`
+	IP string `mapstructure:"ip" json:"ip"`
+
+	id ids.NodeID
+	ip netip.AddrPort
+}
+
+func (c *PeerConfig) Validate() error {
```
richardpringle, line 20: Would it be better to have a constructor `newPeerConfig` that does the validation _before_ building the config?

peers/app_request_network.go
```diff
@@ -87,7 +91,7 @@ func NewNetwork(
 	metrics, err := newAppRequestNetworkMetrics(registerer)
 	if err != nil {
 		logger.Fatal("Failed to create app request network metrics", zap.Error(err))
-		panic(err)
+		return nil, err
```
richardpringle, line 94: Should the logger still log a `Fatal` if we're not panicking here anymore?

peers/app_request_network.go
```diff
@@ -126,95 +133,155 @@ func NewNetwork(
...
-// ConnectPeers connects the network to peers with the given nodeIDs.
-// Returns the set of nodeIDs that were successfully connected to.
-func (n *appRequestNetwork) ConnectPeers(nodeIDs set.Set[ids.NodeID]) set.Set[ids.NodeID] {
+func (n *appRequestNetwork) TrackSubnet(subnetID ids.ID) {
+	n.lock.RLock()
+	if n.trackedSubnets.Contains(subnetID) {
+		n.lock.RUnlock()
+		return
+	}
+	n.lock.RUnlock()
```
richardpringle, line 204: Could pop this out into its own method: 
```go
func (n *appRequestNetwork) containsSubnet(subnetID ids.ID) {
    n.lock.RLock()
    defer n.lock.RUnlock()
    
    return n.trackedSubnets.Contains(subnetID)
}
```
Which cleans up the callsite a little 
```go
func (n *appRequestNetwork) TrackSubnet(subnetID ids.ID) {
    if n.containsSubnet(subnetID) {
        return
    }
    
    // ... 
}
```
🤷 

I always worry that someone's going to make a change that'll cause a deadlock. Might border on paranoia though haha 😅

peers/app_request_network.go
```diff
@@ -126,95 +133,155 @@ func NewNetwork(
...
+	numConnected := 0
+	for _, peer := range peers {
+		vdrs, err := pClient.GetCurrentValidators(context.Background(), constants.PrimaryNetworkID, []ids.NodeID{peer.ID})
+		if err != nil {
+			panic(err)
+		}
+		// Only track the peer if it is a primary network validator
+		if len(vdrs) == 0 {
+			continue
+		}
```
geoff-vball, line 164: So we're assuming that if the peer knows of any p-chain validators, it itself is a primary network validator? Shouldn't subnet-only validators know of some p-chain validators because they have to sync the p-chain?

peers/app_request_network.go
```diff
@@ -126,95 +133,155 @@ func NewNetwork(
...
+		n.lock.RUnlock()
+		return
+	}
+	n.lock.RUnlock()
+
 	n.lock.Lock()
 	defer n.lock.Unlock()
+	n.logger.Debug("Tracking subnet", zap.String("subnetID", subnetID.String()))
+	n.trackedSubnets.Add(subnetID)
+	updateValidatorSet(context.Background(), n.logger, subnetID, n.manager, n.validatorClient)
```
richardpringle, line 210: Is it a common pattern to hold the lock outside of the function that performs the update?

peers/app_request_network.go
```diff
@@ -126,95 +133,155 @@ func NewNetwork(
...
+func updateValidatorSet(
+	ctx context.Context,
+	logger logging.Logger,
+	subnetID ids.ID,
+	manager vdrs.Manager,
+	client *validators.CanonicalValidatorClient,
+) error {
+	logger.Debug("Fetching validators for subnet ID", zap.Stringer("subnetID", subnetID))
+
+	// Fetch the primary network validators from the P-Chain
```
geoff-vball, line 238: I think this comment is wrong? We're fetching the validator set for `subnetID`

tests/basic_relay.go
cam-schultz: This was a test artifact. Removed now.

tests/basic_relay.go
cam-schultz: This was a test artifact. Removed now.

peers/external_handler.go
cam-schultz: This was a test artifact. Removed now.

signature-aggregator/aggregator/aggregator.go
```diff
@@ -39,10 +39,10 @@ type blsSignatureBuf [bls.SignatureLen]byte
 
 const (
 	// Number of retries to collect signatures from validators
-	maxRelayerQueryAttempts = 5
+	maxRelayerQueryAttempts = 10
 	// Maximum amount of time to spend waiting (in addition to network round trip time per attempt)
 	// during relayer signature query routine
-	signatureRequestRetryWaitPeriodMs = 10_000
+	signatureRequestRetryWaitPeriodMs = 20_000
```
richardpringle, line 45: Not sure if it's in the description, but why were these values doubled? (Just for my own knowledge)

tests/signature_aggregator_api.go
```diff
@@ -122,110 +118,20 @@ func SignatureAggregatorAPI(network *network.LocalNetwork, teleporter utils.Tele
...
-
-				samples[metricName] = value
-			} else {
-				log.Debug("Ignoring non-metric line", "line", line)
-			}
-		}
+	warpMessage = getWarpMessageFromLog(ctx, receipt, subnetBInfo)
+
+	reqBody = api.AggregateSignatureRequest{
+		Message: "0x" + hex.EncodeToString(warpMessage.Bytes()),
```
richardpringle, line 134: We don't have a helper for prefixing with a `0x`!?

richardpringle (commented):
I think my only real comment of value for you is the `logger.Fatal` comment. Everything else is more of an opportunity for you to inform me 😄

peers/app_request_network.go
geoff-vball: @richardpringle I agree. I think unlocks should always be `defer`'d if possible, so I think putting this in it's own helper is good practice.

peers/app_request_network.go
geoff-vball: Yeah, it might be better to make `updateValidatorSet` a receiver on the `network`? And then we just pass in the `subnetID` to the function.

peers/app_request_network.go
iansuvak: I believe this is just a confirmation that the peer returned by the `infoAPI.Peers` is indeed a primary network validator

peers/app_request_network.go
iansuvak: My personal rule is under 5 lines spanning Lock and non-deferred unlock and no potential errors in between and this passes that test but fine with extracting as well.

peers/app_request_network.go
```diff
@@ -126,95 +133,155 @@ func NewNetwork(
...
+	height, err := client.GetCurrentHeight(ctx)
 	if err != nil {
-		n.logger.Error(
-			"Failed to get peers",
-			zap.Error(err),
-		)
-		return nil
+		return err
+	}
+	validators, err := client.GetValidatorSet(ctx, height, subnetID)
```
iansuvak, line 243: this won't work publicly with a non-"proposed" height though right?

peers/app_request_network.go
```diff
@@ -126,95 +133,155 @@ func NewNetwork(
...
-			n.logger.Error(
-				"Failed to get API Node IP",
-				zap.Error(err),
-			)
-		} else {
-			trackedNodes.Add(apiNodeID)
-			n.network.ManuallyTrack(apiNodeID, apiNodeIPPort)
+	// Add any elements from the new validator set that are not in the manager
+	for _, vdr := range validators {
+		if _, ok := manager.GetValidator(subnetID, vdr.NodeID); !ok {
```
iansuvak, line 267: just confirming but I assume it's okay to never need to update the validators in the managers here manually?

peers/app_request_network.go
iansuvak: yes but from the p-chain

config/peer_config.go
cam-schultz: We do validate at construction time in the top-level config constructor: https://github.com/ava-labs/awm-relayer/blob/main/relayer/config/viper.go#L21

cam-schultz: Not the cleanest pattern, but we have to unmarshal the config and validate it in separate steps regardless, it's just a  question of code organization.

signature-aggregator/aggregator/aggregator.go
cam-schultz: Might not be necessary anymore, but I increased this to account for the increased warm up time when 1) tracking a new subnet or 2) adding/removing validators. With `ValidatorManager` [churn tracking](https://github.com/ava-labs/teleporter/blob/main/contracts/validator-manager/ValidatorManager.sol#L544), the latter case is likely rare, but the former will be common on the signature aggregator side. The current implementation eliminates this warm up latency for the new subnet case, so it's probably safe to reduce this back to the original value.

peers/app_request_network.go
cam-schultz: > I believe this is just a confirmation that the peer returned by the infoAPI.Peers is indeed a primary network validator

Yes, this is the case here. We're using `pClient.GetCurrentValidators` to filter the peers returned by `info.Peers`. We could definitely do this in one rpc call though. Making individual rpc calls for each peer is pretty inefficient.

peers/app_request_network.go
cam-schultz: Not sure I follow. The result of this function call should be that the manager matches the validator set retreived from the P-Chain API.

iansuvak: I might be missing something but I was wondering about the weight change cases. I see the removal and addition cases for validator mismatches between the P-Chain API and the manager but I also see that the manager keeps track of weights, and I don't see the case where they both exist but have different weights (if that's possible)

Weights can change through delegation while the validator is still active so I wasn't sure if there was a need to keep the manager state somehow manually in sync

peers/app_request_network.go
cam-schultz: Addressed

peers/app_request_network.go
cam-schultz: I went ahead and added the helper. I agree that unlocking with `defer` whenever possible is healthy practice.

peers/app_request_network.go
cam-schultz: Moved the lock management inside of `updateValidatorSet` and made it a `network` receiver

tests/signature_aggregator_api.go
cam-schultz: There's some helpers here and there in various dependencies, but nothing immediately available that I'm able to find. We only do this a couple times and then only in tests, so I'm not too concerned.

peers/app_request_network.go
cam-schultz: Updated this to make one call to `GetCurrentValidators`

peers/app_request_network.go
cam-schultz: Good point. I updated this to call `getValidatorsAt(proposed)` under the hood.

peers/app_request_network.go
cam-schultz: The validator manager is used purely for peer discovery, so the weights don't matter here. A validator's weight doesn't affect the way it's gossiped to peers AFAIK.

peers/app_request_network.go
iansuvak: Wait I was wrong here. @geoff-vball is right that the comment is wrong and should match the debug statement above

geoff-vball: Love the thumbs down of your own message 😛

iansuvak (approved):
LGTM, the only outstanding comment I think is @geoff-vball 's about the comment being misleading.

peers/app_request_network.go
```diff
@@ -151,35 +154,43 @@ func NewNetwork(
...
+		} else {
 			logger.Warn(
 				"Failed to connect to enough bootstrap nodes",
 				zap.Int("targetBootstrapNodes", NumBootstrapNodes),
 				zap.Int("numAvailablePeers", len(peers)),
 				zap.Int("connectedBootstrapNodes", numConnected),
 			)
+			if numConnected == 0 {
+				return nil, fmt.Errorf("failed to connect to any bootstrap nodes")
+			}
```
geoff-vball: I'm not convinced the logic here is right. Let's say `NumBootstrapNodes = 2` and `len(vdrs) = 2`, and we're able to connect to one node but not the other. I think we will hit the `else` case here in an infinite loop.

peers/app_request_network.go
```diff
@@ -126,95 +135,171 @@ func NewNetwork(
...
+				"Failed to connect to enough bootstrap nodes",
+				zap.Int("targetBootstrapNodes", NumBootstrapNodes),
+				zap.Int("numAvailablePeers", len(peers)),
+				zap.Int("connectedBootstrapNodes", numConnected),
+			)
+			if numConnected == 0 {
+				return nil, fmt.Errorf("failed to connect to any bootstrap nodes")
+			}
+			break
+		}
```
geoff-vball, line 197: This is pretty nitpicky, but I think the bug you had here not breaking out of the loop probably would have been caught faster if you followed the pattern of checking for error cases first, and then continuing function execution as normal, ex: 

```suggestion
		i, ok := s.Next() 
		if !ok {
			// If we've sampled all the nodes and still haven't connected to the target number of bootstrap nodes,
			// then warn and stop sampling by either returning an error or breaking
			logger.Warn(
				"Failed to connect to enough bootstrap nodes",
				zap.Int("targetBootstrapNodes", NumBootstrapNodes),
				zap.Int("numAvailablePeers", len(peers)),
				zap.Int("connectedBootstrapNodes", numConnected),
			)
			if numConnected == 0 {
				return nil, fmt.Errorf("failed to connect to any bootstrap nodes")
			}
			break
		}
	        if peer, ok := peersMap[vdrs[i].NodeID]; ok {
		        logger.Info(
			        "Manually tracking bootstrap node",
			        zap.Stringer("ID", peer.ID),
			        zap.Stringer("IP", peer.PublicIP),
		        )
		        testNetwork.ManuallyTrack(peer.ID, peer.PublicIP)
		        numConnected++
	        }
```

geoff-vball (approved):
Left 3 comments that are all related to a bug you had in one commit but then subsequently fixed. You're welcome to take the code organization suggestion or just merge as is :)

peers/app_request_network.go
cam-schultz: I like this better - updated now

richardpringle (approved):

iansuvak (approved):

PR #567, opened 2024-11-26
use rpc options when bootstrapping network

cam-schultz:
## Why this should be merged
Fixes a bug introduced in https://github.com/ava-labs/awm-relayer/pull/559 that does not use the configured RPC parameters when calling the P-Chain API on network startup. This causes deployments that utilize API tokens or query parameters to fail to start.

## How this works
Fetches the configured RPC settings before calling the C-Chain client.

## How is this documented


geoff-vball (approved):

iansuvak (approved):
I think I prefer #568 but we can get this out in the meantime

PR #568, opened 2024-11-26
Add pchain client wrapper with options

cam-schultz:
## Why this should be merged
Alternative to https://github.com/ava-labs/awm-relayer/pull/567

## How is this documented

### DIFF
--- a/peers/info_client.go
+++ b/peers/info_client.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/ava-labs/avalanchego/api/info"
 	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/upgrade"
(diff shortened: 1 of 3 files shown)


bernard-avalabs (commented):
Generally looks good to me. It requires more code than the alternative, but it is also less error prone.

peers/info_client.go
```diff
@@ -23,13 +24,13 @@ type InfoAPI struct {
 	options []rpc.Option
 }
 
-func NewInfoAPI(apiConfig *config.APIConfig) (*InfoAPI, error) {
+func NewInfoAPI(apiConfig *config.APIConfig) *InfoAPI {
```
richardpringle, line 27: 🥇

richardpringle (approved):

PR #579, opened 2024-12-05
move setting to unhealthy after retries

iansuvak:
## Why this should be merged

Currently the healthcheck fatals immediately on failure. This means that for intermittent connection failures this leads to a race between the reconnect setting the status back to healthy before the healthcheck is called again. If the healthcheck is called before the reconnect then the process shuts down. 

Alternative and potentially cleaner approach would be to not have the healthcheck failure shut down the process but let the kubernetes or whatever manages the process to shut it down based on the failed check and additional conditions/thresholds. 

## How this works
Doesn't set the status to unhealthy until the reconnect attempts have failed. 

## How is this documented

### DIFF
--- a/relayer/listener.go
+++ b/relayer/listener.go
@@ -204,8 +204,6 @@ func (lstnr *Listener) processLogs(ctx context.Context) error {
 				zap.String("sourceBlockchainID", lstnr.sourceBlockchain.GetBlockchainID().String()),
 				zap.Error(err),
 			)
-			// TODO try to resubscribe in perpetuity once we have a mechanism for refreshing state
-			// variables such as Quorum values and processing missed blocks.
 			err = lstnr.reconnectToSubscriber()
 			if err != nil {
 				lstnr.healthStatus.Store(false)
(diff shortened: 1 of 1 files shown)


relayer/listener.go
```diff
@@ -209,6 +208,7 @@ func (lstnr *Listener) processLogs(ctx context.Context) error {
 			// variables such as Quorum values and processing missed blocks.
```
cam-schultz, line 208: I think we can get rid of this TODO. If implemented, we'd likely still want to mark the relayer as unhealthy after some period of time or number of failed retries. To me, there's no functional difference between that and simply retrying a fixed number of times before marking as unhealthy, as we do now.

iansuvak: I removed the TODO. Thinking about this more I think we should not fatal on the unhealthy status and let the outside caller of the binary decide what to do and how long to give it to become healthy again.

cam-schultz (approved):

geoff-vball (approved):

PR #582, opened 2024-12-09
Darwin amd64

iansuvak:
## Why this should be merged
allows darwin amd64 builds as part of releases
## How is this documented


cam-schultz (commented):
I think we'll also need to override the compiler, e.g. https://github.com/ava-labs/avalanche-cli/blob/main/.goreleaser.yml#L31-L35

cam-schultz (approved):

geoff-vball (approved):

PR #583, opened 2024-12-09
Remove slashes from the tag-prefix for GH action trigger

iansuvak:
## Why this should be merged
#575 removed them from the goreleaser prefix but GH actions still trigger off of tags with slashes. this fixes it.
## How is this documented

### DIFF
--- a/.github/workflows/release_icm_relayer.yml
+++ b/.github/workflows/release_icm_relayer.yml
@@ -6,7 +6,7 @@ name: Release icm-relayer
 on:
   push:
     tags:
-      - "icm-relayer/*"
+      - "icm-relayer-*"
 
 jobs:
   release:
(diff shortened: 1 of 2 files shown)


geoff-vball (approved):

.github/workflows/release_icm_relayer.yml
```diff
@@ -6,7 +6,7 @@ name: Release icm-relayer
 on:
   push:
     tags:
-      - "icm-relayer/*"
+      - "icm-relayer-*"
```
cam-schultz, line 9: To confirm, we'll tag releases as `icm-relayer-v1.0.0` and `signature-aggregator-v1.0.0` moving forward?

cam-schultz (approved):

Issue #453, opened 2024-08-21
Add retries to RPC clients


cam-schultz:
**Context and scope**
It's currently up to the caller to implement retry loops when making RPC calls via packages such as `ethclient`. We should provide utilities or extensions that wrap RPC calls with retries handled automatically.

**Discussion and alternatives**
- We should investigate adding a generic utility that takes a function to be called in a retry loop, and a set of parameters dictating how retries are handled.
- We should consider integrating exponential backoff (rather than constant or linearly increasing retry intervals). Packages such as https://pkg.go.dev/github.com/cenkalti/backoff/v4 provide this out of the box.

**Open questions**
Questions that are still being discussed.

PR #593, opened 2024-12-12
add exponential retry mechanism for rpc requests using utility function

najeal:
## Why this should be merged

Relates to #453 

## How this works

Functions needing retry are passed to the utility function managing the retry mechanism.

## How this was tested

unit test has been added for `WithMaxRetries()` embedding `WithMaxRetriesLog()`

## How is this documented


utils/backoff.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/ava-labs/avalanchego/utils/logging"
+	"github.com/cenkalti/backoff/v4"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+// WithMaxRetriesLog runs the operation until it succeeds or max retries has been reached.
+// It uses exponential back off.
+// It optionally logs information if logger is set.
+func WithMaxRetriesLog(
```
cam-schultz, line 15: Rather than separate with/without log functions, I think we should instead have a single method that takes a `logging.Logger` and emits warning logs on retries. I can't think of any use cases where we'd want retry attempts to be logged as Warn for some operations, but not logged at all for others.

utils/backoff.go
```diff
@@ -0,0 +1,44 @@
...
+	operation backoff.Operation,
+	max uint64,
+	logger logging.Logger,
+	msg string,
+	fields ...zapcore.Field,
```
cam-schultz, line 20: I think logging `msg` and `fields` on each attempt is not worth the added complexity of having to pass them in as arguments. Rather, `WithMaxRetries` should emit a generic warning log on each attempt failure, and we can leave it to the caller to construct a cohesive error log in the failure case.

utils/backoff.go
```diff
@@ -0,0 +1,44 @@
...
+) error {
+	attempt := uint(1)
+	expBackOff := backoff.WithMaxRetries(backoff.NewExponentialBackOff(), max)
```
cam-schultz, line 23: Let's use `backoff.WithMaxElapsedTime` instead. At present, we choose the number of retries and the delay between each to resolve to a set elapsed time before we emit an error.

relayer/application_relayer.go
```diff
@@ -276,53 +271,36 @@ func (r *ApplicationRelayer) createSignedMessage(
...
-		if attempt != maxRelayerQueryAttempts {
-			// Sleep such that all retries are uniformly spread across totalRelayerQueryPeriodMs
-			// TODO: We may want to consider an exponential back off rather than a uniform sleep period.
-			time.Sleep(time.Duration(signatureRequestRetryWaitPeriodMs/maxRelayerQueryAttempts) * time.Millisecond)
-		}
-	}
-	r.logger.Warn(
-		"Failed to get aggregate signature from node endpoint",
-		zap.Int("attempts", maxRelayerQueryAttempts),
+	err = utils.WithMaxRetriesLog(
```
cam-schultz, line 274: Just a head's up, the Warp API integration is planned to be deprecated in the near future. It's still reasonable to integrate exponential backoff here for the time being.

cam-schultz (commented):
I left a handful of comments, but overall the backoff utility looks good.

There are a couple of other places that exponential backoff would make sense. Can you please integrate the utility in these places as well?
- [p2p signature aggregation](https://github.com/ava-labs/icm-services/blob/main/signature-aggregator/aggregator/aggregator.go#L337) If this proves to be non-trivial, we can defer to a later ticket
- The existing [call with retry](https://github.com/ava-labs/icm-services/blob/main/utils/utils.go#L62) utility and its callsites

utils/backoff.go
najeal: If I understand well, you want to use `backoff.WithMaxElapsedTime` instead of `backoff.WithMaxRetries`.

Instead of giving a `maxRetry` to the utility function, I will give a `maxElapsedTime` derived from the existing values (`numberOfRetries * delayBetweenEach`)

vms/evm/subscriber.go
```diff
@@ -20,9 +21,8 @@ import (
 const (
 	// Max buffer size for ethereum subscription channels
 	maxClientSubscriptionBuffer = 20000
-	subscribeRetryTimeout       = 1 * time.Second
 	MaxBlocksPerRequest         = 200
-	rpcMaxRetries               = 5
+	retryMaxElapsedTime         = 5 * time.Second
```
cam-schultz, line 25: Let's rename this `resubscribeMaxElapsedTime` to reflect its usage.

vms/evm/subscriber.go
```diff
@@ -128,75 +128,53 @@ func (s *subscriber) processBlockRange(
...
-	for {
+	operation := func() (err error) {
 		header, err = s.rpcClient.HeaderByNumber(context.Background(), headerNumber)
-		if err == nil {
-			return header, nil
-		}
-		s.logger.Warn(
+		return err
+	}
+	err = utils.WithMaxRetries(operation, retryMaxElapsedTime, s.logger)
```
cam-schultz, line 135: Let's use a default rpc retry timeout specified in the utils package, that I suggested adding in another comment.

vms/evm/subscriber.go
```diff
@@ -128,75 +128,53 @@ func (s *subscriber) processBlockRange(
...
+	if err != nil {
+		return fmt.Errorf("failed to subscribe to node with all %d attempts", maxResubscribeAttempts)
+	}
+	return nil
 }
 
-func (s *subscriber) subscribe() error {
-	sub, err := s.wsClient.SubscribeNewHead(context.Background(), s.headers)
+// subscribe until it succeeds or reached maxSubscribeAttempts.
+func (s *subscriber) subscribe(maxSubscribeAttempts uint64) error {
```
cam-schultz, line 163: Looks like `maxSubscribeAttempts` is unused.

cam-schultz (commented):
I left a handful of minor comments, but overall this looks great!

relayer/application_relayer.go
```diff
@@ -28,17 +28,11 @@ import (
 )
 
 const (
-	// Number of retries to collect signatures from validators
-	maxRelayerQueryAttempts = 5
-	// Maximum amount of time to spend waiting (in addition to network round trip time per attempt)
-	// during relayer signature query routine
-	signatureRequestRetryWaitPeriodMs = 10_000
+	retryMaxElapsedTime = 10 * time.Second
```
najeal, line 31: I think we can rename the consts using `MaxElapsedTime` (ex: `retryMaxElapsedTime` -> `retryTimeout`) as we are now using `utils.DefaultRPCTimeout`. Otherwise we are mixing the use of `Timeout` and `MaxElapsedTime` when calling the function

cam-schultz: Sounds good to me! My main concern is that we have distinct `const`s for the various timeout scenarios.

najeal: I have changed consts as `Timeout` and kept distinct ones 👍

relayer/application_relayer.go
```diff
@@ -28,17 +28,11 @@ import (
 )
 
 const (
-	// Number of retries to collect signatures from validators
-	maxRelayerQueryAttempts = 5
-	// Maximum amount of time to spend waiting (in addition to network round trip time per attempt)
-	// during relayer signature query routine
-	signatureRequestRetryWaitPeriodMs = 10_000
+	retryTimeout = 10 * time.Second
```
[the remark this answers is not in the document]
najeal: I think we were waiting `time.Duration(signatureRequestRetryWaitPeriodMs/maxRelayerQueryAttempts) * time.Millisecond`
-> `(10_000 / 5) milliseconds` -> `2_000` milliseconds each round

signature-aggregator/aggregator/aggregator.go
```diff
@@ -28,18 +28,17 @@ import (
...
 
 type blsSignatureBuf [bls.SignatureLen]byte
 
 const (
-	// Number of retries to collect signatures from validators
-	maxRelayerQueryAttempts = 10
 	// Maximum amount of time to spend waiting (in addition to network round trip time per attempt)
 	// during relayer signature query routine
-	signatureRequestRetryWaitPeriodMs = 20_000
+	signatureRequestTimeout = 20 * time.Second
```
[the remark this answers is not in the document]
najeal: It is the same calcul here

vms/evm/subscriber.go
```diff
@@ -22,7 +22,7 @@ const (
 	// Max buffer size for ethereum subscription channels
 	maxClientSubscriptionBuffer = 20000
 	MaxBlocksPerRequest         = 200
-	retryMaxElapsedTime         = 5 * time.Second
+	resubscribeTimeout          = 5 * time.Second
```
cam-schultz, line 25: Looks like this is unused

relayer/listener.go
```diff
@@ -20,10 +21,10 @@ import (
 )
 
 const (
-	maxSubscribeAttempts = 10
+	retrySubscribeTimeout = 10 * time.Second
 	// TODO attempt to resubscribe in perpetuity once we are able to process missed blocks and
 	// refresh the chain config on reconnect.
-	maxResubscribeAttempts = 10
+	retryResubscribeTimeout = 10 * time.Second
```
cam-schultz, line 27: We can remove `retry` from these names to be consistent with the timeout constants in other files.

cam-schultz (commented):
Two more minor comments, otherwise LGTM 👍

vms/evm/subscriber.go
```diff
@@ -128,75 +128,53 @@ func (s *subscriber) processBlockRange(
...
-func (s *subscriber) subscribe() error {
-	sub, err := s.wsClient.SubscribeNewHead(context.Background(), s.headers)
+// subscribe until it succeeds or reached timeout.
+func (s *subscriber) subscribe(retryTimeout time.Duration) error {
+	var sub interfaces.Subscription
+	operation := func() (err error) {
+		sub, err = s.wsClient.SubscribeNewHead(context.Background(), s.headers)
+		return err
+	}
+	err := utils.WithRetriesTimeout(s.logger, operation, retryTimeout)
```
najeal, line 169: I would personally keep the `retry` to explicitly make the difference between the `Retry Timeout` and `Unique Running Operation Timeout`
For example right here there is no `timeout` added to the context for the ws call, but it would be possible.
@cam-schultz what do you think?

relayer/listener.go
najeal: I thought about the way to differentiate retry timeout and operation timeout. Please see my comment https://github.com/ava-labs/icm-services/pull/593#discussion_r1899165757

vms/evm/subscriber.go
cam-schultz: I don't feel strongly either way. This reasoning makes sense.

PR #616, opened 2025-01-03
Allow passing tracked L1s config

iansuvak:
## Why this should be merged
Enables optional passing in of tracked L1s to signature-aggregator which allows for earlier connections and speeds up the tests in #694 
## How this was tested
Tested via E2E in #694 on the 
## How is this documented
TODO


geoff-vball (approved):

bernard-avalabs (approved):

signature-aggregator/config/config.go
```diff
@@ -35,6 +37,10 @@ type Config struct {
 	MetricsPort        uint16             `mapstructure:"metrics-port" json:"metrics-port"`
 	SignatureCacheSize uint64             `mapstructure:"signature-cache-size" json:"signature-cache-size"`
 	AllowPrivateIPs    bool               `mapstructure:"allow-private-ips" json:"allow-private-ips"`
+	TrackedL1s         []string           `mapstructure:"tracked-l1s" json:"tracked-l1s"`
+
+	// convenience fields
+	trackedL1s set.Set[ids.ID]
```
cam-schultz, line 43: Let's initialize this with set.NewSet() in `Validate`. It should be safe as is, but I prefer explicit constructor calls over implicit zero value initialization whenever possible.

iansuvak: Sorry I missed this comment earlier but initialized now.

cam-schultz (approved):

PR #621, opened 2025-01-09
Configurable teleporter-max-gas-limit

### LINKED ISSUE #491: Parameterize block gas limit
**Context and scope**
https://github.com/ava-labs/awm-relayer/pull/430/files added a hardcoded tx limit of 12 million gas to comply with the C-Chain's 15 million gas limit per block. This limit may vary from subnet to subnet, so we should make this configurable.

We must add validation checks to cap the gas limit for C-Chain transactions, since this is a known (fixed) block limit.

**Open questions**

najeal:
## Why this should be merged

It closes #491 

## How this works

`DestinationClient` now embeds configured `teleporter-max-gas-limit`, `messageHandler` can now decide to reject or not the message based on the `teleporter-max-gas-limit` value coming from each `DestinationClient` instead of using default hardcoded value.

If the new config field is not provided, previous default value is set to stay consistent with older versions of `icm-services`

## How this was tested

Existing tests have been updated.

## How is this documented


vms/evm/destination_client.go
```diff
@@ -44,41 +45,55 @@ type destinationClient struct {
...
+			"Could not decode destination chain ID from string",
 			zap.Error(err),
 		)
 		return nil, err
 	}
 
-	destinationID, err := ids.FromString(destinationBlockchain.BlockchainID)
+	maxGasLimit := destinationBlockchain.TeleporterMaxGasLimit
+	isCChain := destinationID == cChainID
+	if isCChain && maxGasLimit > config.DefaultTeleporterMaxGasLimit {
```
cam-schultz, line 68: Can we do this validation in `DestinationBlockchain.Validate` instead? If we do so, there shouldn't be any need to pass in `cChainID` as a parameter.

vms/evm/destination_client.go
```diff
@@ -210,3 +226,7 @@ func (c *destinationClient) SenderAddress() common.Address {
 func (c *destinationClient) DestinationBlockchainID() ids.ID {
 	return c.destinationBlockchainID
 }
+
+func (c *destinationClient) TeleporterMaxGasLimit() uint64 {
```
cam-schultz, line 230: Let's rename `TeleporterMaxGasLimit` to `BlockGasLimit` here and elsewhere. The destination client tx construction limits are not exclusive to Teleporter.

relayer/config/destination_blockchain.go
```diff
@@ -13,16 +13,23 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 )
 
+const (
+	// The maximum gas limit that can be specified for a Teleporter message
+	// Based on the C-Chain 15_000_000 gas limit per block, with other Warp message gas overhead conservatively estimated.
+	DefaultTeleporterMaxGasLimit = 12_000_000
```
cam-schultz, line 19: Once the validation is moved to this package as I suggested in another comment, let's not export this const.

vms/destination_client.go
```diff
@@ -51,6 +58,16 @@ func CreateDestinationClients(
...
+	infoClient, err := peers.NewInfoAPI(relayerConfig.InfoAPI)
+	if err != nil {
+		logger.Error("Failed to create info API", zap.Error(err))
+		return nil, err
+	}
+	cChainID, err := infoClient.GetBlockchainID(context.Background(), "C")
+	if err != nil {
+		logger.Error("Failed to get C-Chain ID", zap.Error(err))
+		return nil, err
+	}
```
cam-schultz, line 70: We shouldn't need this either if we move the C-Chain gas limit validation to the config package, as I suggested in another comment.

cam-schultz (commented):
Thanks for putting this together! I left some relatively minor comments.

relayer/config/destination_blockchain.go
najeal: Do we really want to make it `unexported` ?
We use it in two places :
- https://github.com/ava-labs/icm-services/blob/d9040bbf1b5a6a6166e99ab22b8402f4cb31372d/messages/teleporter/message_handler_test.go#L120
- https://github.com/ava-labs/icm-services/blob/d9040bbf1b5a6a6166e99ab22b8402f4cb31372d/messages/teleporter/message_handler_test.go#L236
It would require to switch these with hardcoded values.

cam-schultz: That unit test doesn't need the value of `DefaultTeleporterMaxGasLimit` specifically. Instead, we can define a gas limit const in the test itself.

Not a huge deal either way, but I'd prefer to not export application variables for unit tests as much as possible.

najeal: Makes sense, I will make the change 👍

cam-schultz (approved):

geoff-vball (approved):

PR #637, opened 2025-01-22
chore: fix some comments

petercover:
## Why this should be merged

 fix some comments

## How is this documented

### DIFF
--- a/scripts/install_avalanchego_release.sh
+++ b/scripts/install_avalanchego_release.sh
@@ -72,7 +72,7 @@ else
       git clone --no-checkout ${GIT_CLONE_URL} ${GIT_CLONE_PATH}
     fi
 
-    # check to see if the commitish exists in the repo
+    # check to see if the commit hash exists in the repo
     WORKDIR=$(pwd)
 
     cd ${GIT_CLONE_PATH}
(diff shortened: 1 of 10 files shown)


geoff-vball (approved):
Thanks for these!!

scripts/install_avalanchego_release.sh
```diff
@@ -72,7 +72,7 @@ else
       git clone --no-checkout ${GIT_CLONE_URL} ${GIT_CLONE_PATH}
     fi
 
-    # check to see if the commitish exists in the repo
+    # check to see if the committish exists in the repo
```
cam-schultz, line 75: I think this should be `commit hash`

cam-schultz (approved):

iansuvak:
@petercover We require all commits to be signed. Can you make sure that your commits are signed with your public key and then force push them again? Here are the [github docs](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification) about signing

Also thanks for making the fixes =)

petercover:
> @petercover We require all commits to be signed. Can you make sure that your commits are signed with your public key and then force push them again? Here are the [github docs](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification) about signing
> 
> Also thanks for making the fixes =)

Thanks. Signed!

Issue #619, opened 2025-01-08
Startup health checks


cam-schultz:
**Context and scope**
Currently, the relayer and signature aggregator return positive health checks as soon as the health API is initialized. Instead, we should wait to signal healthy until the respective application is ready to relay messages/aggregate signatures.

The sole requirement for this is that the relayer/aggregator is connected to sufficient stake for any L1 that it will be aggregating signatures for. 
- In the relayer case, all such L1s are provided in the config, and are known at startup. Further, the relayer tracks the health of source chains separately. We can mark each source chain as healthy as soon as we are connected to a quorum of stake necessary to construct a valid signature. (The required quorum is determined by the _receiving_ chain, but using the default value of 67% for health check purposes seems reasonable)
- For the aggregator, an initial list of L1s may be provided via config, but it also is able to aggregate signatures from an L1 _not_ provided via config. The health check criteria should be that we're connected to sufficient stake on the primary network, and for any initial L1s.

**Open questions**
This change would expand the semantics of the health check to include readiness on startup. Before, a failed health check was only possible after initialization, and always indicated an unrecoverable error. Here, a failed health check may also indicate that the application is not ready, but may become so. **Should this functionality be split into a separate readiness check?**

PR #644, opened 2025-01-27
Add network connectivity to health-checks

### LINKED ISSUE #447: Add more meaningful health checks to signature-aggregator
**Context and scope**
For https://github.com/ava-labs/awm-relayer/issues/424 a very basic health check was added, effectively just indicating when the API is up and listening and serving requests.

We should consider what other health indicators the signature aggregator should have and integrate those with the initial health check that was added.

back-reference: https://github.com/ava-labs/awm-relayer/pull/442#discussion_r1723606826

iansuvak:
## Why this should be merged
Closes #447 
Partially addresses #619 

## How this works
- Removes the `TrackSubnet()` call from previously called `ConnectToCanonicalValidators`. This was the only thing actually initializing connections and it was asynchronous and the return of that function was primarily concerned with checking the current status of the validator set and connections to them
- Fixes the bug where `ConnectToCanonicalValidators` was always returning as connected to all validators without checking the actual network state of completing handshakes
- Adds healthcheck function to the signatureAggregator

## How this was tested
- E2E testing
- New unit test for the `GetConnectedCanonicalValidators`
- Manually starting up against Fuji and confirming that signature-aggregator doesn't return healthy until it's ready to aggregate signatures from the C-chain. 

## How is this documented
Comments


peers/app_request_network.go
```diff
@@ -391,6 +408,26 @@ func (n *appRequestNetwork) setPChainAPICallLatencyMS(latency float64) {
 
 // Non-receiver util functions
 
+func GetNetworkHealthFunc(network AppRequestNetwork, subnetIDs []ids.ID) func(context.Context) error {
+	return func(context.Context) error {
+		for _, subnetID := range subnetIDs {
+			connectedValidators, err := network.GetConnectedCanonicalValidators(subnetID)
+			if err != nil {
+				return fmt.Errorf(
+					"Failed to connect to quorum of validator for subnetID: %s, %w", subnetID, err)
```
cam-schultz, line 417: nits:
- by convention we lower case error strings
- I think "failed to get connected validators" is more accurate, since the error says nothing about whether we have a quorum.

peers/validators/canonical_validator_client.go
```diff
@@ -19,7 +21,15 @@ import (
 	pchainapi "github.com/ava-labs/avalanchego/vms/platformvm/api"
 )
 
-var _ validators.State = &CanonicalValidatorClient{}
+var _ CanonicalValidatorState = &CanonicalValidatorClient{}
+
+// CanonicalValidatorState is an interface that wraps validators.State and adds additional necessary
```
cam-schultz, line 26: Looks like this comment is incomplete

signature-aggregator/aggregator/aggregator.go
```diff
@@ -38,7 +38,10 @@ type blsSignatureBuf [bls.SignatureLen]byte
 const (
 	// Maximum amount of time to spend waiting (in addition to network round trip time per attempt)
 	// during relayer signature query routine
-	signatureRequestTimeout = 20 * time.Second
+	signatureRequestTimeout = 5 * time.Second
+	// Maximum amount of time to spend waiting for a connection to a quorum of validators for
+	// a given subnetID
+	connectToValidatorsTimeout = 5 * time.Second
```
cam-schultz, line 44: Seems odd that these timeouts would be less than the default AppRequest timout of 10s (which we use)

signature-aggregator/aggregator/aggregator.go
```diff
@@ -90,6 +93,54 @@ func (s *SignatureAggregator) Shutdown() {
...
+	quorumPercentage uint64,
+) (*peers.ConnectedCanonicalValidators, error) {
+	s.network.TrackSubnet(signingSubnet)
+
+	var connectedValidators *peers.ConnectedCanonicalValidators
+	var err error
+	connectOp := func() error {
+		connectedValidators, err = s.network.GetConnectedCanonicalValidators(signingSubnet)
+		if err != nil {
+			msg := "Failed to connect to canonical validators"
```
cam-schultz, line 107:
```suggestion
			msg := "Failed to fetch connected canonical validators"
```

signature-aggregator/main/main.go
```diff
@@ -139,7 +139,11 @@ func main() {
 		metricsInstance,
 		signatureAggregator,
 	)
-	healthcheck.HandleHealthCheckRequest()
+
+	healthCheckSubnets := cfg.GetTrackedSubnets().List()
```
cam-schultz, line 143: [Not for this PR] A neat extension would be to add newly requested L1's to `healthCheckSubnets`, and perhaps evict them after some duration.

cam-schultz (commented):
Left a few nits, but overall this LGTM. I like the separation between network and listener health checks.

signature-aggregator/main/main.go
iansuvak: Thanks for bringing this up! This is a neat idea but I don't like that this would make it possible to force public facing signature-aggregator instances into unhealthy states by requesting signatures from valid subnets with unreachable nodes.

signature-aggregator/aggregator/aggregator.go
iansuvak: I do think that timeout management here could use more thought but I'm not sure what 10s timeout are you referencing here.

 https://github.com/ava-labs/icm-services/blob/5d697b51c7d4dcce4c0cd21b3ae61b31a8763273/peers/app_request_network.go#L37 defines 2 seconds as `DefaultAppRequestTimeout`, which effectively makes signature requests a 2-3 max tries operation. 

The 20 second timeout I replaced was masking the fact that by the time we made initial requests we weren't connected to any of the validators.

peers/app_request_network.go
```diff
@@ -325,12 +337,9 @@ func (c *ConnectedCanonicalValidators) GetValidator(nodeID ids.NodeID) (*warp.Va
 	return c.ValidatorSet[c.NodeValidatorIndexMap[nodeID]], c.NodeValidatorIndexMap[nodeID]
 }
 
-// ConnectToCanonicalValidators connects to the canonical validators of the given subnet and returns the connected
-// validator information
-func (n *appRequestNetwork) ConnectToCanonicalValidators(subnetID ids.ID) (*ConnectedCanonicalValidators, error) {
-	// Track the subnet
-	n.TrackSubnet(subnetID)
-
+// GetConnectedCanonicalValidators returns the connected validator information for the given subnet
```
bernard-avalabs, line 340: Quick clarification question: what is a "connected" validator? Is it a validator that one or more of the canonical validators have seen in the network? Or given the name of the function (GetConnectedCanonicalValidators), are we only returning canonical validators that are connected to the network? The comment here doesn't make any mention of canonical validators.

bernard-avalabs (commented):
Just have a clarification question about connected validators.

signature-aggregator/aggregator/aggregator.go
cam-schultz: I may have been remembering an old value for `DefaultAppRequestTimeout`. The adaptive timeout [config](https://github.com/ava-labs/icm-services/blob/main/peers/external_handler.go#L48) also factors here. Min and max timeouts are 5 and 10s. 

I do think these timeouts should be longer than the max expected app request timeout, since these operations comprise multiple app requests.

signature-aggregator/main/main.go
cam-schultz: Fair point. Adding an L1 to `healthCheckSubnets` only after an aggregate signature has successfully been constructed may be a fair compromise. It would be nice to support health checks for high traffic L1's that are not included in the config. Either way, we can defer to a future issue.

peers/app_request_network.go
iansuvak: I will improve the comment but in this case:

- "connected" means that this network struct has completed a handshake with them and we therefore have active connection to them. 
- "canonical" is just making sure that we are returning validators in the canonical ordering to enable valid signature construction. 

This struct returns the full validator set information necessary to construct signatures as well as the `ConnectedValidatorWeight` integer which is really the only information on the "connected" portion. I think it might be potentially valuable to also return information on which nodeIds are we actually connected to not attempt to send to nodeIDs that will just fail on the outbound side.

peers/external_handler.go
```diff
@@ -21,6 +21,13 @@ import (
 
 var _ router.ExternalHandler = &RelayerExternalHandler{}
 
+const (
+	maximumNetworkTimeout = constants.DefaultNetworkMaximumTimeout
```
cam-schultz, line 25: Can we just use `constants.DefaultNetworkMaximumTimeout` directly on lines 28 and 58?

cam-schultz (approved):
LGTM. Just one nit

geoff-vball (approved):

PR #667, opened 2025-02-07
Empty app response handling

iansuvak:
## Why this should be merged
Fixes a bug in handling empty SignatureResponses. This was introduced with ACP-118 support since the SignatureResponse only has a single Signature field with `omitempty` tag, meaning that the empty signature serializes to empty slice. 

Trying to unmarshal an empty slices however results in a runtime panic. 

## How this was tested
E2E + added new unit test for `aggregator.unmarshalResponse`

## How is this documented

### DIFF
--- a/signature-aggregator/aggregator/aggregator.go
+++ b/signature-aggregator/aggregator/aggregator.go
@@ -48,6 +48,7 @@ var (
 	// Errors
 	errNotEnoughSignatures     = errors.New("failed to collect a threshold of signatures")
 	errNotEnoughConnectedStake = errors.New("failed to connect to a threshold of stake")
+	errEmptyResponseBytes      = errors.New("received empty app response bytes")
 )
 
 type SignatureAggregator struct {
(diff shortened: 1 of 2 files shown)


signature-aggregator/aggregator/aggregator_test.go
```diff
@@ -346,6 +348,54 @@ func TestCreateSignedMessageSucceeds(t *testing.T) {
...
+		expectedSignature blsSignatureBuf
+	}{
+		{
+			name:              "empty slice",
+			appResponseBytes:  []byte{},
+			expectedSignature: blsSignatureBuf{},
+		},
+		{
+			name:              "nil slice",
+			appResponseBytes:  []byte{},
```
cam-schultz, line 376: should appResponseBytes be `nil` here?

signature-aggregator/aggregator/aggregator.go
```diff
@@ -644,6 +644,9 @@ func (s *SignatureAggregator) marshalRequest(
 }
 
 func (s *SignatureAggregator) unmarshalResponse(responseBytes []byte) (blsSignatureBuf, error) {
+	if len(responseBytes) == 0 { // empty responses are valid and indicate the node has not seen the message
```
cam-schultz, line 647: nit: IMO putting the comment on its own line would be more readable

cam-schultz (approved):
LGTM. Just two minor comments/questions

PR #668, opened 2025-02-07
Don't go through json

richardpringle:
## Why this should be merged

Serializing something then deserializing it just to use the automatic code generation doesn't seem like the best idea, especially when there's only a single field to handle.

## How this works

The code should be self-explanatory. Let me know if it's not.

## How this was tested

I updated the test and added another case


geoff-vball (approved):
🙏

PR #672, opened 2025-02-12
Put the right handler in the right place

richardpringle:
## Why this should be merged

The actual change makes no difference because both handlers are being used with the same `MessageCoordinator` instance. If we were to change things up, it would be really confusing as to why this was not working as expected


geoff-vball (approved):

geoff-vball (approved):

iansuvak (approved):
on a roll of finding gems 💎  🤣

PR #684, opened 2025-02-28
Manually tracked ips config for signature-aggregator

iansuvak:
## Why this should be merged
This is already possible for the relayer but not for sig-agg. Allowing this would enable folks running on private networks and without reachable public ips set on their validators to still test the L1 conversion process. 
## How this works
Exposes the same config available in relayer. 
## How this was tested
E2E
## How is this documented
Added the doc reference fixed the link. Should follow up with a docs PR that elaborates on the reference here and somehow better merges/references shared config between relayer and signature aggregator


bernard-avalabs (approved):

geoff-vball (approved):

signature-aggregator/README.md
```diff
@@ -10,14 +10,19 @@ To build the application run `scripts/build_signature_aggregator.sh` which will
...
 Namely:
 - `LogLevel`: string
 - `PChainAPI`: APIConfig
 - `InfoAPI` : APIConfig
 - `APIPort` : (optional) defaults to 8080
 - `MetricsPort`: (optional) defaults to 8081
+- `SignatureCacheSize` integer (optional) defaults to 1024KB
+- `AllowPrivateIPs` bool (optional)
+- `TrackedSubnetIDs` []string (optional)
+- `ManuallyTrackedPeers` []PeerConfig
```
cam-schultz, line 24:
```suggestion
- `ManuallyTrackedPeers` []PeerConfig (optional)
```

cam-schultz (approved):

PR #685, opened 2025-03-04
replace snyk with semgrep

iansuvak:
## Why this should be merged
Looks like this repo was missed when we did this for others: https://github.com/ava-labs/icm-contracts/pull/705
## How is this documented


cam-schultz (approved):
Thanks for catching this

geoff-vball (approved):

PR #694, opened 2025-03-11
Static NodeIDs via TLS cert configs

### LINKED ISSUE #693: Enable static NodeIDs for private networks.
**Context and scope**
Subnets can be configured with `validatorOnly` option which makes them unwilling to accept connections from any node that's not a validator of their subnet or isn't explicitly included in the `allowedNodes` config field. 

For Relayer and Signature Aggregator to support these use-case they need to be able to start up with a consistent NodeID. NodeIDs are calculated dynamically based on the TLS cert which is generated dynamically on the Test Network startup.

For this implementation we should enable manual configuration of TLS cert/key pair paths.

iansuvak:
## Why this should be merged

Closes #693

## How this works

Allows for manual setting of TLS Cert and key paths. If these configuration values are set but the files are not there it will automatically generate them and save them to the location. 

## How this was tested
Added a new integration test. 

## How is this documented
TODO


relayer/config/config.go
```diff
@@ -65,8 +66,11 @@ type Config struct {
 	SignatureCacheSize     uint64                   `mapstructure:"signature-cache-size" json:"signature-cache-size"`
 	ManuallyTrackedPeers   []*basecfg.PeerConfig    `mapstructure:"manually-tracked-peers" json:"manually-tracked-peers"`
 	AllowPrivateIPs        bool                     `mapstructure:"allow-private-ips" json:"allow-private-ips"`
+	TLSCertPath            string                   `json:"tls-cert-path,omitempty"`
+	TLSKeyPath             string                   `json:"tls-key-path,omitempty"`
 
 	// convenience field to fetch a blockchain's subnet ID
+	tlsCert                *tls.Certificate `json:"-"`
```
geoff-vball, line 73: Does `json:"-"` do something special?

config/common_config.go
```diff
@@ -0,0 +1,62 @@
...
+
+	var keyMissing, certMissing bool
+
+	if _, err := os.Stat(keyPath); os.IsNotExist(err) {
+		keyMissing = true
+	}
+	if _, err := os.Stat(certPath); os.IsNotExist(err) {
+		certMissing = true
+	}
+	if !(keyMissing && certMissing) && (keyMissing || certMissing) {
```
geoff-vball, line 34:
```suggestion
	if keyMissing != certMissing {
```

config/common_config.go
```diff
@@ -0,0 +1,62 @@
...
+	"os"
+
+	"github.com/ava-labs/avalanchego/staking"
+	"github.com/spf13/viper"
+)
+
+const TLSCertPathKey = "tls-cert-path"
+const TLSKeyPathKey = "tls-key-path"
+
+func GetTLSCertFromFile(v *viper.Viper) (tls.Certificate, error) {
```
geoff-vball, line 18: I feel like this should return `(*tls.Certificate, error)`. `LoadTLSCertFromFiles` returns a pointer and our one call to this function converts what we're currently returning into a pointer.

relayer/config/config.go
iansuvak: Makes it not serialize but also unnecessary since it's not exported. Thanks

tests/utils/utils.go
```diff
@@ -425,7 +425,7 @@ func WriteSignatureAggregatorConfig(signatureAggregatorConfig signatureaggregato
 	data, err := json.MarshalIndent(signatureAggregatorConfig, "", "\t")
 	Expect(err).Should(BeNil())
 
-	f, err := os.CreateTemp(os.TempDir(), fname)
+	f, err := os.CreateTemp("/tmp", fname)
```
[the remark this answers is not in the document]
iansuvak: Testing artifact --will change back.

relayer/config/viper.go
```diff
@@ -113,5 +114,13 @@ func BuildConfig(v *viper.Viper) (Config, error) {
 		cfg.DestinationBlockchains[i].AccountPrivateKey = utils.SanitizeHexString(privateKey)
 	}
 
+	if v.IsSet(commonConfig.TLSKeyPathKey) || v.IsSet(commonConfig.TLSCertPathKey) {
```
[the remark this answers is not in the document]
iansuvak: Yeah -- I'm iffy on this but  the idea is that setting just one of the keys is undefined behavior and we want to error out on it since it's likely a user error, and since we are already potentially returning errors on from the `GetTLSCertFromFile` wanted to keep it together with the rest of the validation.

config/common_config.go
```diff
@@ -0,0 +1,62 @@
...
+	cert, err := staking.LoadTLSCertFromFiles(keyPath, certPath)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't read staking certificate: %w", err)
+	}
+	return cert, nil
+}
+
+// getExpandedPath gets the string in viper corresponding to [key] and expands
+// any variables using the OS env.
+func getExpandedPath(v *viper.Viper, key string) string {
```
cam-schultz, line 55: IIUC, this is intended to be used to expand config values the contain environment variables. E.g. `TLS_CERT_PATH=${HOME}/path/to/cert`. 

Why not just require that env var expansion be done outside of the application when the cfg file is written?

config/common_config.go
```diff
@@ -0,0 +1,62 @@
...
+
+	if _, err := os.Stat(keyPath); os.IsNotExist(err) {
+		keyMissing = true
+	}
+	if _, err := os.Stat(certPath); os.IsNotExist(err) {
+		certMissing = true
+	}
+	if keyMissing != certMissing {
+		// If only one of the key/cert pair is missing return an error
+		// otherwise, create the staking key/cert pair
```
cam-schultz, line 36: The phrasing of this `otherwise` implies that if both the key and cert are present, then we will overwrite the staking key/cert pair

peers/app_request_network.go
```diff
@@ -142,6 +146,20 @@ func NewNetwork(
...
+		testNetworkConfig.TLSConfig = peer.TLSConfig(*cert, nil)
+		testNetworkConfig.TLSKey = cert.PrivateKey.(crypto.Signer)
+	} else {
+		cert = &testNetworkConfig.TLSConfig.Certificates[0]
+	}
+	parsedCert, err := staking.ParseCertificate(cert.Leaf.Raw)
+	if err != nil {
+		return nil, err
+	}
+	nodeId := ids.NodeIDFromCert(parsedCert)
```
cam-schultz, line 161: nit, let's name this var `nodeID` for consistency with similar namings in this repo

relayer/config/config.go
```diff
@@ -65,8 +66,11 @@ type Config struct {
 	SignatureCacheSize     uint64                   `mapstructure:"signature-cache-size" json:"signature-cache-size"`
 	ManuallyTrackedPeers   []*basecfg.PeerConfig    `mapstructure:"manually-tracked-peers" json:"manually-tracked-peers"`
 	AllowPrivateIPs        bool                     `mapstructure:"allow-private-ips" json:"allow-private-ips"`
+	TLSCertPath            string                   `json:"tls-cert-path,omitempty"`
+	TLSKeyPath             string                   `json:"tls-key-path,omitempty"`
```
cam-schultz, line 70: Do we also need the `mapstructure` tags here? Viper uses mapstructure tags, but there may be some json tag compatibility I'm not aware of.

cam-schultz, line 70: We should add these config options to the top-level readme.

config/common_config.go
```diff
@@ -0,0 +1,62 @@
...
+	"os"
+
+	"github.com/ava-labs/avalanchego/staking"
+	"github.com/spf13/viper"
+)
+
+const TLSCertPathKey = "tls-cert-path"
+const TLSKeyPathKey = "tls-key-path"
+
+func GetTLSCertFromFile(v *viper.Viper) (*tls.Certificate, error) {
```
cam-schultz, line 18: Can we add a unit test for this funciton?

tests/validators_only_network.go
```diff
@@ -0,0 +1,189 @@
...
+	requestURL := fmt.Sprintf("http://localhost:%d%s", signatureAggregatorConfig.APIPort, api.APIPath)
+
+	reqBody := api.AggregateSignatureRequest{
+		Message: "0x" + hex.EncodeToString(warpMessage.Bytes()),
+	}
+	client := http.Client{
+		Timeout: 20 * time.Second,
+	}
+
+	var sendRequestToAPI = func() {
```
cam-schultz, line 158: Does this function need to be a var?

tests/validators_only_network.go
```diff
@@ -0,0 +1,189 @@
...
+			tmpnetNode.RuntimeConfig.ReuseDynamicPorts = true
+			// Restart the network to apply the new chain configs
+			ctx, cancel := context.WithTimeout(context.Background(), 120*time.Second)
+			defer cancel()
+			err := network.RestartNode(ctx, logging.NoLog{}, tmpnetNode)
+			Expect(err).Should(BeNil())
+		}
+	}
+
+	// start sig-agg again with the same TLS cert
```
cam-schultz, line 133: Can we add a signature request using a signature aggregator without the same TLS cert to prove that the private network is correctly rejecting traffic from disallowed nodeIDs?

cam-schultz (commented):
Generally LGTM. I wonder if there are any scenarios where us write a TLS cert to the supplied paths would be unexpected. If a user is supplying a path, they are definitely expecting a read, but maybe not a write.

We could add a third config option to allow writing the cert if we hit that code path, or conversely, a config option to disallow writing. Curious what your thoughts here are.

config/common_config.go
iansuvak: Removed the wording since the next case describes it more precisely.

geoff-vball (approved):

PR #699, opened 2025-03-13
make initial connections blocking

iansuvak:
## How is this documented


relayer/network_utils.go
```diff
@@ -52,33 +61,43 @@ func InitializeConnectionsAndCheckStake(
...
+			connectedValidators, err := network.GetConnectedCanonicalValidators(subnetID)
+			if err != nil {
+				logger.Error(
+					"Failed to connect to canonical validators",
+					zap.String("subnetID", subnetID.String()),
+					zap.Error(err),
+				)
+				return err
+			}
+			if ok, warpConfig, err := checkForSufficientConnectedStake(logger, cfg, connectedValidators, blockchainID); ok && err == nil {
```
geoff-vball, line 84: I think we should still return `err` if `err != nil` here. It only happens if the warp config is invalid. I think we should only be looping if we return `err == nil && !ok`

Same thing applies for `connectToPrimaryNetworkPeers()`

relayer/network_utils.go
```diff
@@ -52,33 +61,47 @@ func InitializeConnectionsAndCheckStake(
...
+					zap.Uint64("totalValidatorWeight", connectedValidators.ValidatorSet.TotalWeight),
+					zap.Any("WarpConfig", warpConfig),
+				)
+				select {
+				case <-ctx.Done():
+					return ctx.Err()
+				default:
+					time.Sleep(5 * time.Second) // Retry after a short delay
+				}
+			}
```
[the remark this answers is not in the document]
geoff-vball: I also removed `warpConfig` from the log in the error case because we don't return it from `checkForSufficientConnectedStake` on error.

relayer/network_utils.go
```diff
@@ -16,6 +18,8 @@ import (
 	"go.uber.org/zap"
 )
 
+const initialConnectionTimeoutSeconds = 1800
```
cam-schultz, line 21: 30 minutes is excessive for general purposes. Let's reduce this back to 5 mins, and expose a config to allow for longer timeouts.

relayer/network_utils.go
```diff
@@ -29,16 +33,21 @@ func InitializeConnectionsAndCheckStake(
...
 					"failed to connect to primary network peers: %w",
 					err,
 				)
 			}
 		} else {
-			if err := connectToNonPrimaryNetworkPeers(logger, network, cfg, sourceBlockchain); err != nil {
+			if err := connectToNonPrimaryNetworkPeers(ctx, logger, network, cfg, sourceBlockchain); err != nil {
 				return fmt.Errorf(
 					"failed to connect to non-primary network peers: %w",
 					err,
```
cam-schultz, line 53: Since we're waiting on an async process for each source chain, a cleaner approach would be to call `connectToPrimaryNetworkPeers` in a goroutine for each source chain, and use a wait group to block until we're fully connected.

relayer/network_utils.go
```diff
@@ -52,33 +61,47 @@ func InitializeConnectionsAndCheckStake(
...
+					zap.Uint64("connectedWeight", connectedValidators.ConnectedWeight),
+					zap.Uint64("totalValidatorWeight", connectedValidators.ValidatorSet.TotalWeight),
+					zap.Any("WarpConfig", warpConfig),
+				)
+				select {
+				case <-ctx.Done():
+					return ctx.Err()
+				default:
+					time.Sleep(5 * time.Second) // Retry after a short delay
+				}
```
cam-schultz, line 103: Stylistically, putting the `select` statement immediately under the enclosing `for` would make this more readable. Instead of sleeping in the default case, we should add a ticker that triggers `checkForSufficientConnectedStake` every 5 seconds.

cam-schultz: We should also use a variable instead of a magic number for the retry period.

cam-schultz (commented):
I know this is marked as `DO NOT MERGE`, but IMO we should implement this blocking behavior for all relayers. There's not really a use case in which we wouldn't want to do so. 

I left a handful of comments to clean up this logic. The code here flew under the radar around Etna activation, and I suspect only continued to work as expected for fully connected relayer configurations.

relayer/network_utils.go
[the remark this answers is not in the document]
iansuvak: I made the change and made it match in the other function but now I am thinking we should add warp config back. It is a pointer and I believe that zap.Any should handle nil pointers fine. The benefit of this is that we no longer have a tripartite return though which might be worth it just for that case alone.

relayer/config/config.go
```diff
@@ -31,12 +31,13 @@ const (
...
-	defaultMetricsPort         = uint16(9090)
-	defaultIntervalSeconds     = uint64(10)
-	defaultSignatureCacheSize  = uint64(1024 * 1024)
+	defaultStorageLocation                 = "./.icm-relayer-storage"
+	defaultProcessMissedBlocks             = true
+	defaultAPIPort                         = uint16(8080)
+	defaultMetricsPort                     = uint16(9090)
+	defaultIntervalSeconds                 = uint64(10)
+	defaultSignatureCacheSize              = uint64(1024 * 1024)
+	defaultInitialConnectionTimeoutSeconds = uint64(300)
```
cam-schultz, line 40: [no action needed] We could consider calculating this based on the number of nodes we're attempting to connect to, though with the high environment variability we've observed, that might be tricky to get right.

relayer/network_utils.go
```diff
@@ -87,6 +123,7 @@ func connectToNonPrimaryNetworkPeers(
 // Connect to the validators of the destination blockchains. Verify that we have connected
 // to a threshold of stake for each blockchain.
 func connectToPrimaryNetworkPeers(
```
cam-schultz, line 125: We should get rid of this function altogether, and rename `connectToNonPrimaryNetworkPeers` to something like `connectToPeers`. The self-signing optimization is no longer relevant, so we should connect to peers of any subnet in the same way.

relayer/network_utils.go
```diff
@@ -29,56 +32,89 @@ func InitializeConnectionsAndCheckStake(
...
-	connectedValidators, err := network.GetConnectedCanonicalValidators(subnetID)
-	if err != nil {
-		logger.Error(
-			"Failed to connect to canonical validators",
-			zap.String("subnetID", subnetID.String()),
-			zap.Error(err),
-		)
-		return err
-	}
 	for _, destination := range sourceBlockchain.SupportedDestinations {
```
cam-schultz, line 83: Is connecting to destination peers an existing mistake? We should only need to connect to peers of the signing subnet, which will always be the source chain's subnet in the relayer's case.

iansuvak: It's not connecting to destination peers but checking for destination ICM verification rules to determine what is the sufficient percentage which is correct behavior I think.

geoff-vball: Yeah. Since the destinations can have different warp quorum numerators, we need to make sure we have enough stake to connect to the destination with the highest numerator. I've actually written some logic that will check the highest numerator instead of reconnecting every time we want to check a different destination.

iansuvak: This isn't reconnecting. It's still effectively just checking in the loop.

relayer/network_utils.go
```diff
@@ -29,92 +31,78 @@ func InitializeConnectionsAndCheckStake(
...
-			if err := connectToPrimaryNetworkPeers(logger, network, cfg, sourceBlockchain); err != nil {
-				return fmt.Errorf(
-					"failed to connect to primary network peers: %w",
-					err,
-				)
-			}
-		} else {
-			if err := connectToNonPrimaryNetworkPeers(logger, network, cfg, sourceBlockchain); err != nil {
+		eg.Go(func() error {
+			if err := connectToPeers(ctx, logger, network, cfg, sourceBlockchain); err != nil {
```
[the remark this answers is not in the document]
iansuvak: I prefer wrapping an error here since otherwise it would just be a context exceeded error which is IMO less helpful.

relayer/network_utils.go
geoff-vball: Yeah, sorry, not reconnecting, but we are making a p-chain api call every time in `GetConnectedCanonicalValidators()`

relayer/network_utils.go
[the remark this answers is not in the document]
geoff-vball: In any case, we should change the error message

relayer/network_utils.go
iansuvak: But there are still more issues here including the function name

relayer/network_utils.go
iansuvak: - re: ticker vs sleep.
I actually prefer a one line sleep for simple cases like this vs instantiating a new object that will have to get GCed eventually but can make the change
- re: location of the select
I put it at the bottom to avoid having to decide between another helper function or a sizeable chunk of code at a deeper indentation level. After the most recent change there is no reason to break it up further since the whole select block is just fetching the connected validators again and doing the calculation. Let me know which of the 2 options is preferred.

cam-schultz: - The ticker will take into account execution time of the rest of the function, while time.Sleep will not. Since we're making API calls, execution time can vary significantly. I'm not really worried about an extra object for the GC
- The for-select pattern is common enough that I'd argue deviation from that pattern is more difficult to read than the extra indentation. I don't think an additional helper is necessary. Ultimately not going to block anything on stylistic opinions, but that's my 2c.

relayer/network_utils.go
cam-schultz: Ah right, thanks for clarifying. I'm realizing now though that we don't account for the "allow all destinations" case here. Instead of returning immediately (and waiting to connect to peers until relay time), requiring connection with the default 67% of stake seems like a reasonable default condition.

geoff-vball: In the "allow all destinations" case, don't we populate the supported destinations with all of configured destinations?

cam-schultz: You're totally right - no need to handle th e default case then

relayer/network_utils.go
```diff
@@ -29,116 +30,78 @@ func InitializeConnectionsAndCheckStake(
...
+	subnetID := sourceBlockchain.GetSubnetID()
+	// Loop over destination blockchains here to confirm connections to a threshold of stake
+	// which is determined by the Warp Quorum configs of the destination blockchains.
 	for _, destination := range sourceBlockchain.SupportedDestinations {
-		blockchainID := destination.GetBlockchainID()
-		subnetID := cfg.GetSubnetID(blockchainID)
-		network.TrackSubnet(subnetID)
-		connectedValidators, err := network.GetConnectedCanonicalValidators(subnetID)
+		destinationBlockchainID := destination.GetBlockchainID()
+		warpConfig, err := cfg.GetWarpConfig(destinationBlockchainID)
```
geoff-vball, line 65: We should loop over all the `SupportedDestinations` to find the largest `quorumNumerator` and then only enter the below `for` loop once

relayer/network_utils.go
```diff
@@ -29,116 +30,82 @@ func InitializeConnectionsAndCheckStake(
...
+		maxQuorumNumerator = max(maxQuorumNumerator, warpConfig.QuorumNumerator)
+	}
 
-		if ok, warpConfig, err := checkForSufficientConnectedStake(logger, cfg, connectedValidators, blockchainID); !ok {
+	ticker := time.Tick(retryPeriodSeconds * time.Second)
+	select {
+	case <-ctx.Done():
+		return fmt.Errorf("failed to connect to sufficient stake: %w", ctx.Err())
+	case <-ticker:
+		for {
```
cam-schultz, line 84: This `for` needs to be outside of the `select` statement. As is, we enter this loop once when the ticker fires, and execute the loop body in a tight loop.

relayer/network_utils.go
```diff
@@ -29,116 +30,88 @@ func InitializeConnectionsAndCheckStake(
...
+		warpConfig, err := cfg.GetWarpConfig(destinationBlockchainID)
+		if err != nil {
+			logger.Error(
+				"Failed to get warp config from chain config",
+				zap.Stringer("destinationBlockchainID", destinationBlockchainID),
+				zap.Error(err),
 			)
 			return err
 		}
+		maxQuorumNumerator = max(maxQuorumNumerator, warpConfig.QuorumNumerator)
```
geoff-vball, line 76: I totally forgot `max` was added in `1.21` 😄

relayer/network_utils.go
geoff-vball: I like the break (or return), avoiding the anonymous function. I'm ambivalent on ticker vs sleep, as it looks like go `1.23` fixed garbage collection of unstopped tickers

cam-schultz: No strong feelings here, all these patterns are valid control flow

cam-schultz (approved):

bernard-avalabs (approved):

PR #700, opened 2025-03-13
Only query connected nodes

cam-schultz:
## Why this should be merged
Only issues outbound requests to nodes with active connections

## How is this documented


signature-aggregator/aggregator/aggregator.go
```diff
@@ -260,24 +260,26 @@ func (s *SignatureAggregator) CreateSignedMessage(
...
+					)
+				}
+				// Register a timeout response for each queried node
+				reqID := ids.RequestID{
+					NodeID:    nodeID,
+					ChainID:   unsignedMessage.SourceChainID,
+					RequestID: requestID,
+					Op:        byte(message.AppResponseOp),
+				}
+				s.network.RegisterAppRequest(reqID)
```
geoff-vball, line 281: Does this possibly register an AppRequest multiple times per `nodeID` now that it's in this inner loop?

cam-schultz: It shouldn't, since `nodeID` can only map to a single validator. However, I added a check against the existing `vdrSet` just in case.

The call to `RegisterAppRequest` should have also been inside the if block - I fixed this as well.

signature-aggregator/aggregator/aggregator.go
```diff
@@ -260,24 +260,26 @@ func (s *SignatureAggregator) CreateSignedMessage(
...
-
-			// Register a timeout response for each queried node
-			reqID := ids.RequestID{
-				NodeID:    nodeID,
-				ChainID:   unsignedMessage.SourceChainID,
-				RequestID: requestID,
-				Op:        byte(message.AppResponseOp),
+			// Add connected nodes to the request
+			for _, nodeID := range vdr.NodeIDs {
+				if s.network.IsConnected(nodeID) {
```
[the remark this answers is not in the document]
cam-schultz: Good call. It makes sense for that type to track connected nodes as well.

iansuvak (dismissed):

geoff-vball (approved):

geoff-vball (approved):

iansuvak (approved):

krud491 (commented):
Remove -18

PR #704, opened 2025-03-19
explicitly set tracked subnets in the relayer application

iansuvak:
## Why this should be merged
Explicitly set tracked subnets at `AppRequestNetwork` creation time since for relayer they are all available from the config. 
## How this was tested
CI
## How is this documented
N/A


geoff-vball (approved):

relayer/main/main.go
```diff
@@ -170,10 +171,14 @@ func main() {
 		})
 	}
 
+	trackedSubnets := set.NewSet[ids.ID](len(cfg.SourceBlockchains))
+	for _, sourceChain := range cfg.SourceBlockchains {
```
michaelkaplan13, line 175: Do we need to consider the case where a non-converted Subnet is still using the self-signing optimization for messages from the C-Chain here? If for whatever reason that chain is only listed as a destination, we probably still want to add it as a tracked subnet since we will attempt to aggregate signatures from it.

iansuvak: Indeed, thanks for the catch

relayer/config/config.go
```diff
@@ -229,6 +229,26 @@ func (c *Config) InitializeWarpConfigs() error {
 	return nil
 }
 
+// Initializes the tracked subnets list. This should only be called after the configuration has been validated and
+// [Config.InitializeWarpConfigs] has been called
+func (c *Config) InitializeTrackedSubnets() error {
```
cam-schultz, line 234: Thoughts on combining this and `InitializeWarpConfigs` into a single `Initialize` function?

cam-schultz, line 234: Thoughts on renaming this function and the corresponding config field from `TrackedSubnets` to `SigningSubnets`? I think that would make the code here more self-documenting. Since this list contains the primary network ID (which I agree it should), it's not the same `trackedSubnets` list that's used by the `Network`.

relayer/config/config.go
cam-schultz, line 1: Can we add a unit test for the new function?

relayer/config/config.go
```diff
@@ -229,6 +229,26 @@ func (c *Config) InitializeWarpConfigs() error {
...
+	for _, sourceBlockchain := range c.SourceBlockchains {
+		c.trackedSubnets.Add(sourceBlockchain.GetSubnetID())
+	}
+	for _, destinationBlockchain := range c.DestinationBlockchains {
+		warpCfg, err := c.GetWarpConfig(destinationBlockchain.GetBlockchainID())
+		if err != nil {
+			return fmt.Errorf(
+				"failed to get warp config for destination blockchain %s: %w",
```
geoff-vball, line 242:
```suggestion
				"failed to get warp config for destination blockchain %s: %w", err
```

peers/app_request_network.go
```diff
@@ -43,10 +43,12 @@ const (
 	InboundMessageChannelSize = 1000
 	ValidatorRefreshPeriod    = time.Second * 5
 	NumBootstrapNodes         = 5
+	maxTrackedSubnets         = 16
```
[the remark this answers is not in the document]
iansuvak: Does this need to be exported? We have plenty of unexported constants throughout the codebase.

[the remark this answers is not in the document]
geoff-vball: omg. Sorry, early in the morning... I completely forgot how Go works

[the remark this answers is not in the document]
geoff-vball: Can we link to it so that we don't accidentally increase the number? Ideally it would be exported from Avalanchego

relayer/config/config.go
iansuvak: `TrackedSubnets` is part of an interface for the `peer.Config` so we'd have to change that on the sig-agg config side. That by itself is not that huge of a deal but then we still have the problem  of `AppRequestNetwork.TrackSubnet`  not modifying this list and being called the same thing and also not being the same thing as anything on the avago network side.

I agree that this needs a re-work but I'd like to punt on this until general network rework.

cam-schultz (approved):

cam-schultz (approved):

PR #711, opened 2025-03-21
Return errors from sig-agg api

iansuvak:
## Why this should be merged
Users asked for this when interacting with APIs and when they don't have access to logs to be able to distinguish between connection errors and bad signature requests
## How is this documented


signature-aggregator/api/api.go
```diff
@@ -184,8 +185,8 @@ func signatureAggregationAPIHandler(
 			quorumPercentage,
 		)
 		if err != nil {
-			msg := "Failed to aggregate signatures"
-			logger.Warn(msg, zap.Error(err))
+			msg := fmt.Errorf("Failed to aggregate signatures. Error: %w", err).Error()
+			logger.Warn(msg)
```
geoff-vball, line 189: Do you know if this is equivalent to using `zap.Error(err)`?

PR #715, opened 2025-03-24
Protect trackedSubnets access

cam-schultz:
## Why this should be merged
A common `trackedSubnets` instance is used in both `icm-services` and in the Avalanchego `Network` object to manage the subnets that the local peer tracks. `Network` does not expect this to change, however, `icm-services` mutates `trackedSubnets` based on incoming signature requests. Since concurrent map reads and writes can cause panics, we must protect access to `trackedSubnets`.

Also fixes a testing bug introduced in https://github.com/ava-labs/icm-services/pull/705 where the Avalanchego nodes in the local network run the `icm-services` workaround [branch](https://github.com/ava-labs/avalanchego/tree/icm-services-patch) instead of n official release.

## How this works
Adds a  shared `*sync.RWMutex` to all types that access `trackedSubnets`.

## How this was tested
CI

## How is this documented
N/A

### DIFF
--- a/peers/app_request_network_test.go
+++ b/peers/app_request_network_test.go
@@ -177,7 +177,7 @@ func TestTrackSubnets(t *testing.T) {
 		metrics:            metrics,
 		manager:            snowVdrs.NewManager(),
 		lruSubnets:         linked.NewHashmapWithSize[ids.ID, interface{}](maxNumSubnets),
-		lock:               new(sync.Mutex),
+		validatorSetLock:   new(sync.Mutex),
 		trackedSubnetsLock: new(sync.RWMutex),
 	}
 	require.Zero(t, arNetwork.trackedSubnets.Len())


scripts/versions.sh
```diff
@@ -28,9 +28,8 @@ function extract_commit() {
 
 # This needs to be exported to be picked up by the dockerfile.
 export GO_VERSION=${GO_VERSION:-$(getDepVersion go)}
-
 # Don't export them as they're used in the context of other calls
-AVALANCHEGO_VERSION=${AVALANCHEGO_VERSION:-$(extract_commit "$(getDepVersion github.com/ava-labs/avalanchego)")}
```
geoff-vball, line 33: Why this change?

geoff-vball (approved):

geoff-vball (approved):

PR #718, opened 2025-03-26
Set fortuna upgrade time

cam-schultz:
## Why this should be merged
The peer handshake logic requires that connecting peers have similar minor versions starting at the activation time for the network upgrade corresponding to that minor version. We use `TestNetwork`, which instead hardcodes this compatibiliity time to 0, meaning that peers must _always_ have similar minor versions.

This is an issue when rolling out network upgrade versions to peers ahead of the activation time. During that window, a relayer using the upgrade version will be unwilling to connect to peers that have not yet upgraded, even though the network upgrade has not yet occurred.

## How this works

Updates the Avalanchego custom dependency branch to add a `minCompatibilityTime` to the `NewTestNetwork` constructor, and adds logic to set this time to the Fortuna activation time for the corresponding network. We will need to update this with every subsequent network activation time as they become available.

## How this was tested
Local testing, CI

## How is this documented
N/A

### DIFF
--- a/peers/app_request_network.go
+++ b/peers/app_request_network.go
@@ -27,6 +27,7 @@ import (
 	vdrs "github.com/ava-labs/avalanchego/snow/validators"
 	"github.com/ava-labs/avalanchego/staking"
 	"github.com/ava-labs/avalanchego/subnets"
+	"github.com/ava-labs/avalanchego/upgrade"
 	"github.com/ava-labs/avalanchego/utils/constants"
 	"github.com/ava-labs/avalanchego/utils/linked"
 	"github.com/ava-labs/avalanchego/utils/logging"
@@ -78,6 +79,7 @@ type AppRequestNetwork interface {
 	) set.Set[ids.NodeID]
 	Shutdown()
 	TrackSubnet(subnetID ids.ID)
+	NumConnectedPeers() int
 }
 
 type appRequestNetwork struct {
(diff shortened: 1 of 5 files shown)


geoff-vball (approved):

michaelkaplan13 (approved):
I think the AvalancheGo side change for this is worthy of an independent PR for the `TestNetwork`, but this LGTM.

peers/app_request_network.go
```diff
@@ -182,7 +184,17 @@ func NewNetwork(
...
-	testNetwork, err := network.NewTestNetwork(logger, networkMetrics, testNetworkConfig, handler)
+	// Set the activation time for the latest network upgrade
+	upgradeTime := upgrade.InitiallyActiveTime
+	switch networkID {
+	case constants.MainnetID:
+		upgradeTime = upgrade.Mainnet.FortunaTime
+	case constants.FujiID:
+		upgradeTime = upgrade.Fuji.FortunaTime
+	default:
+	}
```
michaelkaplan13, line 195:
```suggestion
	upgradeTime := upgrade.GetConfig(networkID).FortunaTime
```

cam-schultz:
I also tested that setting `upgradeTime := upgrade.UnscheduledActivationTime` also works as expected pre-Fortuna activation. This _should_ be safe to use through Fortuna activation, since that time is only ever checked [here](https://github.com/ava-labs/avalanchego/blob/b8cd687c1e36cf3fa2722522ed55809668b6a780/version/compatibility.go#L71), but to be safe I'd like to merge this using the Fortuna activation time, and check `upgrade.UnscheduledActivationTime` after activation so that we can have an A/B against the same validator set.

michaelkaplan13 (approved):

PR #723, opened 2025-03-28
Exclude zero balance validators from aggregated signature

cam-schultz:
## Why this should be merged
Inactive L1 validators still contribute to their L1's overall stake weight, but there signatures must not be included in ICM messages. Validators only consider the in-memory validator set when verifying an ICM signature, but inactive validators are ejected from memory. Therefore, if an inactive valdiator's signature is included in the aggregate, the aggregate will not be verifiable.

## How this works
1. Checks if the signing subnet is an L1
2. Finds all of the zero-balance NodeIDs in the L1 validator set
4. Filters the canonical validator signature map to remove any validators for which *all* NodeIDs have zero balance
5. Filtered canonical validators are never requested for signatures, nor are their signatures ever added to the aggregate. (This is a redundant check)

Note that even if a single NodeID has zero balance, if its BLS public key is registered by another NodeID with a non-zero balance, that key's signature should still be included in the aggregate. These two NodeIDs will have distinct validationIDs, but will still be considered the same canonical validator for purposes of ICM verification.

## How this was tested
CI

TODO: Fix unit tests. This will require a mock P-Chain client

## How is this documented


signature-aggregator/aggregator/aggregator.go
```diff
@@ -179,12 +190,71 @@ func (s *SignatureAggregator) CreateSignedMessage(
...
+			isL1 = true
+		}
+	}
+
+	// Tracks all collected signatures.
+	// For L1s, we must take care to *not* include inactive validators in the signature map.
+	// Inactive validator's stake weight still contributes to the total weight, but the verifying
+	// node will not be able to verify the aggregate signature if it includes an inactive validator.
 	signatureMap := make(map[int][bls.SignatureLen]byte)
+	excludedValidators := set.NewSet[int](0)
```
cam-schultz, line 214: To do: consider fully populating `signatureMap`, and only applying `excludedValidators` at aggregation time.

signature-aggregator/aggregator/aggregator.go
```diff
@@ -179,12 +190,71 @@ func (s *SignatureAggregator) CreateSignedMessage(
 		return nil, err
 	}
 
-	accumulatedSignatureWeight := big.NewInt(0)
+	isL1 := false
+	if signingSubnet != constants.PrimaryNetworkID {
+		subnet, err := s.pChainClient.GetSubnet(context.Background(), signingSubnet, s.pChainClientOptions...)
```
michaelkaplan13, line 195: Not for this PR, but we should probably pass through a context to be used here for the purposes of properly setting/handling timeouts, etc.

signature-aggregator/aggregator/aggregator.go
```diff
@@ -179,12 +190,71 @@ func (s *SignatureAggregator) CreateSignedMessage(
...
+			s.logger.Error(
+				"Failed to get subnet",
+				zap.String("signingSubnetID", signingSubnet.String()),
+				zap.Error(err),
+			)
+			return nil, err
+		}
+		if subnet.ConversionID != ids.Empty {
+			isL1 = true
+		}
```
michaelkaplan13, line 206:
```suggestion
		isL1 = subnet.ConversionID != ids.Empty
```

Also not required for this PR, but we may want to make this check for given source/destination chains once on startup, or maybe periodically after for non-converted subnets, rather than every time we try to construct an aggregate signature. 

If it ever is `true`, we know if will always be true in the future, so we can avoid the extra calls.

signature-aggregator/aggregator/aggregator.go
michaelkaplan13: For now, I think it'd make sense to create a single context to be used within this function and put some reasonable timeout on it, just to ensure we don't hang for whatever reason.

signature-aggregator/aggregator/aggregator.go
```diff
@@ -179,12 +190,71 @@ func (s *SignatureAggregator) CreateSignedMessage(
...
+			s.logger.Error(
+				"Failed to get L1 validators",
+				zap.String("signingSubnetID", signingSubnet.String()),
+				zap.Error(err),
+			)
+			return nil, err
+		}
+		zeroBalanceNodes := set.NewSet[ids.NodeID](0)
+		for _, validator := range l1Validators {
+			if validator.Balance == 0 {
```
michaelkaplan13, line 232: It probably doesn't make too big of difference either way, but we may want to set some low balance threshold where validator get excluded that's greater than 0. At a minimum fee rate of 512 nAVAX/sec, if there is any new P-Chain blocks in between the time we aggregate and the time the transaction is accepted, it could still be evicted. 

Obviously it's a bit of a race no matter what because the chains are moving forward independently, but requiring that validators have a balance of at least 1024 or 2048 nAVAX seems practical to me here.

signature-aggregator/aggregator/aggregator.go
```diff
@@ -179,12 +190,71 @@ func (s *SignatureAggregator) CreateSignedMessage(
...
+	// Tracks all collected signatures.
+	// For L1s, we must take care to *not* include inactive validators in the signature map.
+	// Inactive validator's stake weight still contributes to the total weight, but the verifying
+	// node will not be able to verify the aggregate signature if it includes an inactive validator.
 	signatureMap := make(map[int][bls.SignatureLen]byte)
+	excludedValidators := set.NewSet[int](0)
+
+	// Fetch L1 validators and find the Node IDs with Balance = 0
+	// Find the corresponding canonical validator set index for each of these, and add to the exlusion list
+	// if ALL of the Node IDs for a validator have Balance = 0
```
michaelkaplan13, line 218:
```suggestion
	// Fetch L1 validators and find the node IDs with Balance = 0
	// Find the corresponding canonical validator set index for each of these, and add to the exlusion list
	// if ALL of the node IDs for a validator have Balance = 0
```

super nit

signature-aggregator/aggregator/aggregator.go
```diff
@@ -179,12 +190,71 @@ func (s *SignatureAggregator) CreateSignedMessage(
...
+		}
+		zeroBalanceNodes := set.NewSet[ids.NodeID](0)
+		for _, validator := range l1Validators {
+			if validator.Balance == 0 {
+				s.logger.Debug("Node has zero balance", zap.String("nodeID", validator.NodeID.String()))
+				zeroBalanceNodes.Add(validator.NodeID)
+			}
+		}
+
+		// Find all the canonical validators that have all their Node IDs in the zeroBalanceNodes set
```
michaelkaplan13, line 238:
```suggestion
		// Find all the canonical validators that have all their node IDs in the zeroBalanceNodes set
```

signature-aggregator/aggregator/aggregator.go
```diff
@@ -179,12 +190,71 @@ func (s *SignatureAggregator) CreateSignedMessage(
...
+			for _, nodeID := range validator.NodeIDs {
+				if !zeroBalanceNodes.Contains(nodeID) {
+					break
+				}
+				s.logger.Debug(
+					"Excluding validator",
+					zap.Any("nodeIDs", validator.NodeIDs),
+				)
+				excludedValidators.Add(i)
+			}
```
michaelkaplan13, line 249: Think there's a logic issue here. If the first node ID has zero balance, but not the second, we'll still be adding them to `excludedValidators`.

signature-aggregator/aggregator/aggregator.go
```diff
@@ -403,6 +483,7 @@ func (s *SignatureAggregator) handleResponse(
 	connectedValidators *peers.ConnectedCanonicalValidators,
 	unsignedMessage *avalancheWarp.UnsignedMessage,
 	signatureMap map[int][bls.SignatureLen]byte,
+	excludedValidators set.Set[int],
```
michaelkaplan13, line 486: Just for my own understanding, we shouldn't be getting any responses from excludedValidators given that we're not querying them right? Is this just to be defensive?

michaelkaplan13 (commented):
Thanks for putting this together.

I'd also like to better understand how AvalancheGo handles connections with zero balance validators. Does it eventually disconnect from them if previously connected already, or does it just not establish new connections if their balance is zero?

signature-aggregator/aggregator/aggregator.go
cam-schultz: Yes, we defensively exclude them at aggregation time as well as request time. I'm leaning towards removing this redundancy to simplify the control flow. A functional reason to still query (but not aggregate) inactive nodes is that their signatures can be used to populate the signature cache, saving us from having to request them in the future if their balance is restored.

signature-aggregator/aggregator/aggregator.go
cam-schultz: I'm in favor of adding a non zero floor. The worst case is that we're unable to aggregate enough weight, which would have happened in a matter of seconds anyway as the balance falls to zero, which seems like less of a concern to me than the edge case you mentioned.

signature-aggregator/aggregator/aggregator.go
cam-schultz: 👀 Whoops, nice catch. I refactored this a bit to populate the set with sufficiently funded nodes, rather than insufficiently funded nodes, as I found that easier to reason about.

cam-schultz:
> I'd also like to better understand how AvalancheGo handles connections with zero balance validators. Does it eventually disconnect from them if previously connected already, or does it just not establish new connections if their balance is zero?

AFAIK L1 validator balance is not taken into account in the p2p networking layer. The set of validators with which a peer is willing to connect is dictated by its validator [manager](https://github.com/ava-labs/avalanchego/blob/32cf5fd3de8198b9ff4553a55c631ce778ff9e12/snow/validators/manager.go#L40), which only tracks weight. I was also not able to find any reference to the validator's balance in the `network` package, which defines peer. gossip.

signature-aggregator/aggregator/aggregator.go
cam-schultz: >  I refactored this a bit to populate the set with sufficiently funded nodes

We actually need to track unfunded nodes here, that way the exclusionary check also accounts for legacy non-L1 validator nodes. These nodes are expected to have zero balance, but should still contribute to L1 aggregate signatures.

geoff-vball (commented):
Generally looks good to me!

signature-aggregator/aggregator/aggregator.go
cam-schultz: I have a wip of this here: https://github.com/ava-labs/icm-services/pull/724

But agree that it may not be worth blocking this PR on.

signature-aggregator/aggregator/aggregator.go
cam-schultz: Added a context parameter.

geoff-vball (approved):

signature-aggregator/aggregator/aggregator.go
```diff
@@ -179,12 +195,89 @@ func (s *SignatureAggregator) CreateSignedMessage(
...
+
+		// Set of unfunded L1 validator nodes
+		unfundedNodes := set.NewSet[ids.NodeID](0)
+		for _, validator := range l1Validators {
+			s.logger.Debug(
+				"Node has balance",
+				zap.String("nodeID", validator.NodeID.String()),
+				zap.Uint64("balance", uint64(validator.Balance)),
+			)
+			if uint64(validator.Balance) <= minimumL1ValidatorBalance {
```
iansuvak, line 242: Should this be `<`? Minimum implies that the edge-case balance should still be valid, also semi relatedly -- is this min balance defined as a constant somewhere in `avalanchego`?

signature-aggregator/aggregator/aggregator.go
```diff
@@ -179,12 +195,89 @@ func (s *SignatureAggregator) CreateSignedMessage(
...
+				unfundedNodes.Add(validator.NodeID)
+			} else {
+				s.logger.Debug(
+					"Node has insufficient balance",
```
iansuvak, line 246: should this be "sufficient" balance instead?

signature-aggregator/aggregator/aggregator.go
iansuvak: I prefer simplifying the control flow whenever possible since it's already convoluted =) although I think that querying underfunded validators is an edge case that I don't think is worth it on it's own.

iansuvak (commented):
Generally LGTM, few minor questions. Also not for this PR but `CreateSignedMessage` is growing unwieldy. Maybe we should break up some of this into more unit-testable components.

signature-aggregator/aggregator/aggregator.go
cam-schultz: Yes it should. I originally had `<=` to account for `minimumL1ValidatorBalance = 0`, but I think it makes more sense to disallow `0` altogether. The description of this constant implies that `0` is an invalid value: "The minimum balance that an L1 validator must maintain in order to participate in the aggregate signature."

signature-aggregator/aggregator/aggregator.go
cam-schultz: Yep, good catch. Also removed unconditional logging of validator balances, since it's not particularly useful to track this code path.

iansuvak (approved):

PR #728, opened 2025-04-04
add retries to relayer ProcessMessage

iansuvak:
## Why this should be merged

Allows for the sig-agg cache to be used at times of high validator churn to regenerate aggregate signatures by doing internal retries before setting relayers state to unhealthy 

## How this works

Simple retries with no delays

## How this was tested
Existing CI
TODO: Add new test that explicitly tests validator churn 

## How is this documented


relayer/application_relayer.go
```diff
@@ -262,6 +263,34 @@ func (r *ApplicationRelayer) ProcessMessage(handler messages.MessageHandler) (co
 	return txHash, nil
 }
 
+func (r *ApplicationRelayer) ProcessMessage(handler messages.MessageHandler) (common.Hash, error) {
+	var txHash common.Hash
```
cam-schultz, line 267: nit: we could bring `txHash` into the for loop's scope as a slight improvement to readability, since it's not otherwise set outside the for loop.

cam-schultz (approved):

michaelkaplan13 (approved):

PR #733, opened 2025-04-07
Quorum Percentage Buffer

michaelkaplan13:
## Why this should be merged
Fixes https://github.com/ava-labs/icm-services/issues/729

## How this works
Adds an optional `quorumPercentageBuffer` to `CreateSignedMessage`. Signature requests and aggregation are retried until the required quorum percentage plus the buffer amount is reached. If this percentage is not reached after all retries, an error is returned only if the represented weight is still beneath the required quorum percentage for the given L1.

## How this was tested
Additional unit tests.

## How is this documented
Updated `signature-aggregator/` README with new optional API field.


iansuvak (approved):
I like the clean handling of the edge case where we don't meet the desired buffer but still meet quorum.

relayer/application_relayer.go
```diff
@@ -30,6 +30,11 @@ import (
 const (
 	retryTimeout  = 10 * time.Second
 	maxRetryCount = 5
+
+	// The additional percentage of stake weight that we will try to aggregate signatures from above the required
+	// quorum. This allows for small weight changes in between the time the signature is constructed and the time
+	// it is verified to not cause the verification to fail.
+	defaultQuorumPercentageBuffer = uint64(3)
```
geoff-vball, line 37: Should we expose this is a config option for the relayer? It's a little strange that we have a default here of 3 but sig agg defaults to 0.

iansuvak: I think the sig-agg / relayer distinction makes sense since the user controls every request to sig-agg. 

But yeah the `signature-aggregator/README.md` change should be removed and added to the relayer if we only want it exposed there and then also add the actual handling to have it populated from the config.

michaelkaplan13: I like it being able to be optionally set on a per-request basis for the signature aggregator, but agree making it configurable on the relayer side could make sense also to be able to more easily increase it. 

Thinking about it now, ideally I'd prefer to default to 3 in both cases if omitted, but be able to be set to 0 if provided. Is there and idiomatic way we have of doing that already?

geoff-vball: I think this is how we set non-zero defaults https://github.com/ava-labs/icm-services/blob/main/signature-aggregator/config/viper.go#L54

michaelkaplan13: It's not readily clear to me the best way to make it configurable for the relayer. i.e. Should it be a global config value, or configurable by source or destination chain? And if it's not global, I think having a non-zero default becomes a bit more complicated also.

Are there any strong opinions here? Starting to think it's best to roll with it as is:
- Relayer always uses the default 3% buffer
- Signature aggregator defaults to 0 (unchanged behavior), but allows it be specified on a per request basis

signature-aggregator/aggregator/aggregator.go
```diff
@@ -281,12 +282,15 @@ func (s *SignatureAggregator) CreateSignedMessage(
...
+
+	// Only return early if we have enough signatures to meet the quorum percentage
+	// plus the buffer percentage.
 	if signedMsg, err := s.aggregateIfSufficientWeight(
 		unsignedMessage,
 		signatureMap,
 		accumulatedSignatureWeight,
 		connectedValidators.ValidatorSet.TotalWeight,
-		quorumPercentage,
+		requiredQuorumPercentage+quorumPercentageBuffer,
```
bernard-avalabs, line 293: Should there be checks for more than 100% in case the system is misconfigured? It looks like this check already exists in CalculateQuorumPercentageBuffer.

bernard-avalabs (commented):
Generally makes sense to me.

geoff-vball (dismissed):

iansuvak (approved):

geoff-vball (approved):

PR #737, opened 2025-04-08
Handler refactor to deduplicate logging

iansuvak:
## Why this should be merged
Don't love this but this is an attempt to clean up (?) the handler/handler factory logic in order to have all of the necessary logging information up front.

## How this works

- It moves the `GetMessageRoutingInfo` to the factory instead since it doesn't depend on any handler state
- This enables passing the destinationClient into the handler constructor and storing it in the struct
- Removes destination client from method args since now it's available up front
- Logs destinationChain, warpMessageID and subnetID withContext instead of repeating it as well as passing it into the sig-agg (with the parent branch)

## How this was tested
CI
## How is this documented

### DIFF
--- a/messages/message_handler.go
+++ b/messages/message_handler.go
@@ -17,33 +17,37 @@ import (
 // for each message protocol, and performs the sending to the destination chain.
 type MessageHandlerFactory interface {
 	// Create a message handler to relay the Warp message
-	NewMessageHandler(unsignedMessage *warp.UnsignedMessage) (MessageHandler, error)
+	NewMessageHandler(
+		unsignedMessage *warp.UnsignedMessage,
+		destinationClient vms.DestinationClient,
+	) (MessageHandler, error)
+
+	// Return info for routing the message to the correct relayer
+	GetMessageRoutingInfo(unsignedMessage *warp.UnsignedMessage) (MessageRoutingInfo, error)
+}
+
+// Struct containing fields for routing messages to the correct relayer.
+type MessageRoutingInfo struct {
+	SourceChainID      ids.ID
+	SenderAddress      common.Address
+	DestinationChainID ids.ID
+	DestinationAddress common.Address
 }
 
 // MessageHandlers relay a single Warp message. A new instance should be created for each Warp message.
 type MessageHandler interface {
 	// ShouldSendMessage returns true if the message should be sent to the destination chain
 	// If an error is returned, the boolean should be ignored by the caller.
-	ShouldSendMessage(destinationClient vms.DestinationClient) (bool, error)
+	ShouldSendMessage() (bool, error)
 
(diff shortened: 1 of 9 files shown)


bernard-avalabs (approved):

messages/message_handler.go
```diff
@@ -17,34 +17,38 @@ import (
...
 
 	// SendMessage sends the signed message to the destination chain. The payload parsed according to
 	// the VM rules is also passed in, since MessageManager does not assume any particular VM
 	// returns the transaction hash if the transaction is successful.
-	SendMessage(signedMessage *warp.Message, destinationClient vms.DestinationClient) (common.Hash, error)
+	SendMessage(signedMessage *warp.Message) (common.Hash, error)
 
 	// GetLogContext returns extra fields to be set in the logger
 	// when passing it along to the signature aggregator
 	GetLogContext() []zap.Field
```
cam-schultz, line 50: Since the expected usage is to immediately pass the return value to `logger.With`, can we instead return a formatted `logging.Logger` directly?

cam-schultz (commented):
Generally looks good to me. I left a handful of cleanup suggestions, but I agree that injecting a logger with the desired fields is the best approach.

messages/message_handler.go
iansuvak: For this case right now we could and I considered it but IMO returning the `zap.Field`s instead of the logger is more flexible in general. 

If we ever wanted to add something to the parent logger then getting a child logger from here would force us to either manually re-add any missing fields to it or to just not use them. This allows the caller to decide what do they want to do with the fields that relevant to this handler.

messages/teleporter/message_handler.go
```diff
@@ -41,11 +41,14 @@ type factory struct {
...
 type messageHandler struct {
-	logger            logging.Logger
-	teleporterMessage *teleportermessenger.TeleporterMessage
-	unsignedMessage   *warp.UnsignedMessage
-	factory           *factory
-	deciderClient     pbDecider.DeciderServiceClient
+	logger              logging.Logger
+	teleporterMessage   *teleportermessenger.TeleporterMessage
+	unsignedMessage     *warp.UnsignedMessage
+	factory             *factory
```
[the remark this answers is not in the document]
iansuvak: Had the mental bookmark to remove the backref as well but didn't get around to it =) will do.

messages/message_handler.go
cam-schultz: Makes sense, thanks

geoff-vball (approved):
I think moving the destination client makes a lot of sense. 

Having to keep the `logFields` around basically so we can merge the contexts of two loggers together is clunky, but I can see there's no way to do with cleanly with the current avalanchego wrapper. I did some quick searching and it looks like zap doesn't support this under the hood either :(

PR #738, opened 2025-04-09
Use AUTOMAXPROCS in relayer

### LINKED ISSUE #732: Reconsider relayer catchup parallelism
**Context and scope**
Currently the relayer catchup does a full fanout of goroutines for each block/message. There is evidence of catchup messages being slower to process post restart. We should consider reducing this to reduce contention and minimize unpredictable scheduling when it comes to the gap between signing and delivering messages.

**Open questions**

cam-schultz:
## Why this should be merged

Fixes #732 

On startup, the relayer does a full fan out of all blocks and ICM messages since its last checkpointed height. GOMAXPROCS by default is set to the number of logical cores, which is often greater than the number of CPUs actually available to a containerized deployment. This can lead to massive CPU throttling, even for code that is not CPU-intensive, and therefore extremely long latency on startup. An extreme example saw ~15 minutes to back process 50 ICM messages.

I originally considered adding a fixed size worker pool to throttle CPU usage within the application itself, however that is 1) a larger (and more restrictive) hammer and 2) more appropriate for CPU bound tasks, which `CreateSignedMessage` is not.

Also adds additional `Info` level logs around ICM message relaying that the relayer is configured for.

## How this works

Uses [automaxprocs](https://github.com/uber-go/automaxprocs) to set GOMAXPROCS to the container's CPU quota. This [article](https://medium.com/@aditimishra_541/go-performance-with-ubers-automaxprocs-8f31226a92cd) gives a more in-depth description. 

## How this was tested

CI

## How is this documented
N/A


cam-schultz:
> Is there a easy way for us to empirically validate that automaxprocs is working as intended in our deployment scenario?

The package prints to stdout when it sets (or retains) GOMAXPROCS. We should be able to see that in our logs.

Otherwise, we should see a significant speed up when processing missed blocks, since we do a full fan out of goroutines for all missed blocks and ICM transactions.

iansuvak (approved):
I agree that this is probably a much cleaner solution than manual limiting of the fanout. Ideally would love to profile before/after to see where the CPU time goes but fine making the change as is.

michaelkaplan13 (approved):
👍 

We may want to add a comment explaining the blank import of `automaxprocs` and its effect since it may not be immediately obvious/intuitive in the future, but this LGTM

geoff-vball (approved):

PR #745, opened 2025-04-14
add context + timeout to GetSubnetID API calls

iansuvak:
## Why this should be merged
Passes in context to the PChain API calls to GetSubnetID and rewraps it with a 1 second timeout to prevent too many hanging requests during high traffic periods. 

## How this was tested
CI
## How is this documented

### DIFF
--- a/signature-aggregator/aggregator/aggregator.go
+++ b/signature-aggregator/aggregator/aggregator.go
@@ -51,6 +51,9 @@ const (
 	// The minimum balance that an L1 validator must maintain in order to participate
 	// in the aggregate signature.
 	minimumL1ValidatorBalance = 2048 * units.NanoAvax
+
+	// The maximum amount of time to spend waiting for a subnet ID from the PChain
+	pChainSubnetIDTimeout = 1 * time.Second
 )
 
 var (
@@ -177,7 +180,7 @@ func (s *SignatureAggregator) CreateSignedMessage(
 	var signingSubnet ids.ID
 	var err error
 	// If signingSubnet is not set we default to the subnet of the source blockchain
-	sourceSubnet, err := s.getSubnetID(log, unsignedMessage.SourceChainID)
+	sourceSubnet, err := s.getSubnetID(ctx, log, unsignedMessage.SourceChainID)
 	if err != nil {
 		return nil, fmt.Errorf(
 			"source message subnet not found for chainID %s",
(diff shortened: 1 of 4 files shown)


geoff-vball (approved):

signature-aggregator/aggregator/aggregator.go
```diff
@@ -499,15 +502,21 @@ func (s *SignatureAggregator) CreateSignedMessage(
...
+) (ids.ID, error) {
 	s.subnetsMapLock.RLock()
 	subnetID, ok := s.subnetIDsByBlockchainID[blockchainID]
 	s.subnetsMapLock.RUnlock()
 	if ok {
 		return subnetID, nil
 	}
 	log.Info("Signing subnet not found, requesting from PChain", zap.String("blockchainID", blockchainID.String()))
-	subnetID, err := s.network.GetSubnetID(blockchainID)
+	ctx, cancel := context.WithTimeout(ctx, pChainSubnetIDTimeout)
```
michaelkaplan13, line 517: Nit: Should we name this `getSubnetIDCtx` so it's not overwriting the `ctx` passed in as a param? No functional difference, but it's error prone with we re-use ctx again below in the future as is.

signature-aggregator/aggregator/aggregator.go
```diff
@@ -51,6 +51,9 @@ const (
 	// The minimum balance that an L1 validator must maintain in order to participate
 	// in the aggregate signature.
 	minimumL1ValidatorBalance = 2048 * units.NanoAvax
+
+	// The maximum amount of time to spend waiting for a subnet ID from the PChain
+	pChainSubnetIDTimeout = 1 * time.Second
```
michaelkaplan13, line 56: This is just a gut feeling, but 1 second seems a little low to me in the event that we're firing off a bunch of requests all at once. Was there any reasoning you had for it? I don't know how I would go about choosing a reasonable value, but just thinking outloud.

cam-schultz: Calling out that we do already have a variable to represent the default rpc timeout https://github.com/ava-labs/icm-services/blob/main/utils/utils.go#L32

Unless there's a specific reason to do otherwise here, I'd recommend just using that

iansuvak: Yeah I thought this was lowish in general but the whole step to establish connections to the full validator set which includes getting the API response and making any new connections has a 5 second timeout right now. As does signature collection.

I think this will primarily trigger in the backfill cases but combined with the change to the maximum fanout I think a somewhat aggressive timeout might be appropriate. I'm still happy to bump it to whatever the consensus is but I don't have hard data for what a good number would be

iansuvak: > Calling out that we do already have a variable to represent the default rpc timeout https://github.com/ava-labs/icm-services/blob/main/utils/utils.go#L32
> 
> Unless there's a specific reason to do otherwise here, I'd recommend just using that

Thanks made the update -- it does seem like quite a long one to me though for a basic RPC call though.

michaelkaplan13 (approved):

cam-schultz (approved):

PR #750, opened 2025-04-15
rename log field to not use reserved name

iansuvak:
## Why this should be merged
Noticed that the v1.6.3-rc3 relayers running are logging correct context fields  from inside the `utils.WithRetriesTimeout` but the `message` field is missing most likely because it's a reserved attribute. 
## How this works
renames to a non-reserved value
## How is this documented


cam-schultz (approved):
Nice catch. Checked that we're not using any of the other reserved fields enumerated [here](https://docs.datadoghq.com/service_management/events/explorer/attributes/)

geoff-vball (approved):

PR #754, opened 2025-04-21
Configurable max fees per destination blockchain

michaelkaplan13:
## Why this should be merged
Allows for configuring the `max-base-fee` and `max-priority-fee-per-gas` that should be used for transactions sent by the relayer on a given blockchain.

## How this works
If the `max-base-fee` is set for a given blockchain, that value is always used since the actual amount paid is determined by the current base fee of the chain. If it is not set, a value 3x the current base fee estimate is used.

The "gas fee cap" is still checked on chain in all cases, but is now capped a maximum of the configured `maximum-priority-fee-per-gas`, with a default value of 2.5 gwei.

## How this was tested
CI

## How is this documented
Config values documented in README.


vms/evm/destination_client.go
```diff
@@ -128,20 +166,28 @@ func (c *destinationClient) SendTx(
...
-	baseFeeCtx, baseFeeCtxCancel := context.WithTimeout(context.Background(), utils.DefaultRPCTimeout)
-	defer baseFeeCtxCancel()
-	baseFee, err := c.client.EstimateBaseFee(baseFeeCtx)
-	if err != nil {
-		c.logger.Error(
-			"Failed to get base fee",
-			zap.Error(err),
-		)
-		return common.Hash{}, err
+	// If the max base fee isn't explitictly set, then default to fetching the
```
cam-schultz, line 169:
```suggestion
	// If the max base fee isn't explicitly set, then default to fetching the
```

vms/evm/destination_client.go
```diff
@@ -152,10 +198,12 @@ func (c *destinationClient) SendTx(
 		)
 		return common.Hash{}, err
 	}
+	if gasTipCap.Cmp(c.maxPriorityFeePerGas) > 0 {
+		gasTipCap = c.maxPriorityFeePerGas
+	}
```
cam-schultz, line 203: Why would we override the configured `maxPriorityFeePerGas` here?

relayer/README.md
```diff
@@ -309,6 +309,18 @@ The relayer is configured via a JSON file, the path to which is passed in via th
...
+
+  - The maximum amount of gas that can be used in a single block on this blockchain. The relayer will not attempt to deliver messages that require more gas than this limit to the given chain.
+
+  `"max-base-fee": string`
+
+  - The maximum base fee gas price the relayer is willing to pay on this blockchain. Represented as a decimal encoded unsigned integer string to allow for large values.
+
+  `"max-priority-fee-per-gas": string`
+
+  - The maximum priority fee per gas that the relayer is willing to pay to incentivize transactions being including on this blockchain. The relayer will use the current estimation of the required gas tip cap for this blockchain, up to a maximum of this configured value. Represented as a decimal encoded unsigned integer string to allow for large values.
```
cam-schultz, line 322: Do we realistically need to be concerned with values greater than uint64max? The current values on the C-Chain are a `baseFee` of `105664059` and a `maxPriorityFeePerGas` of `3449218`. For `baseFee` to exceed uint64max, we'd need to see a factor of `1e11` increase, and even more so for `maxPriorityFeePerGas`.

Not a huge deal at the end of the day, but requiring two different JSON datatypes to represent integer values is a bit cumbersome.

bernard-avalabs (commented):
Generally makes sense to me. Left a few minor comments.

vms/evm/destination_client.go
michaelkaplan13: If the current estimated gas tip cap is less than configured maximum, then we use that estimate. This is different than the base fee handling because the transaction actually pays whatever the priority fee is set to, as compared to only paying whatever the actual base fee is on chain no matter how high the `maxBaseFee` is set.

relayer/README.md
michaelkaplan13: Very fair point. I was thinking that uint64max was around 18 nAVAX, but it's 18 AVAX 🤦 

I'll switch this now to clean up the code

vms/evm/destination_client.go
```diff
@@ -25,10 +26,15 @@ import (
...
 
 const (
-	// Set the max fee to twice the estimated base fee.
-	// TODO: Revisit this constant factor when we add profit determination, or make it configurable
-	BaseFeeFactor        = 2
-	MaxPriorityFeePerGas = 2500000000 // 2.5 gwei
+	// If the max base fee is not explicitly set, use 3x the current
+	// base fee estimate
+	defaultBaseFeeFactor        = 3
+	defaultMaxPriorityFeePerGas = 2500000000 // 2.5 gwei
```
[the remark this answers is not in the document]
michaelkaplan13: It's a bit different since the default needs to be set for each `DestinationBlockchain` rather than a top level config value, but I moved the default max priority fee per gas to `destination_blockchain.go`. Worth noting that as a side effect of making the values `uint64`s, we don't distinguish between it being set to 0 and it not being set. Meaning the lowest it can be set to is `1`, which I think is fine.

vms/evm/destination_client.go
```diff
@@ -110,6 +118,34 @@ func NewDestinationClient(
...
+		var ok bool
+		maxBaseFee, ok = new(big.Int).SetString(destinationBlockchain.MaxBaseFee, 10)
+		if !ok || maxBaseFee.Cmp(big.NewInt(0)) <= 0 {
+			logger.Error(
+				"Invalid max base fee",
+				zap.String("maxBaseFee", destinationBlockchain.MaxBaseFee),
+			)
+			return nil, errInvalidMaxBaseFee
+		}
+	}
```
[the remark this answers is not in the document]
michaelkaplan13: Yeah that was correct, but it's updated now to always set the value since we're using `uint64`s in the configuration so we can't distinguish between 0 (which is invalid anyways) and unset. We now use the current estimate times 3 in both of those cases.

vms/evm/destination_client.go
```diff
@@ -152,10 +162,12 @@ func (c *destinationClient) SendTx(
...
 		return common.Hash{}, err
 	}
+	if gasTipCap.Cmp(c.maxPriorityFeePerGas) > 0 {
+		gasTipCap = c.maxPriorityFeePerGas
+	}
 
 	to := common.HexToAddress(toAddress)
-	gasFeeCap := baseFee.Mul(baseFee, big.NewInt(BaseFeeFactor))
-	gasFeeCap.Add(gasFeeCap, big.NewInt(MaxPriorityFeePerGas))
+	gasFeeCap := new(big.Int).Add(maxBaseFee, gasTipCap)
```
cam-schultz, line 170: This looks correct to me, but pointing out that before we were adding a constant 2.5gwei to the base fee, rather than the tip fetched from `SuggestGasTipCap`.

relayer/README.md
```diff
@@ -309,6 +309,18 @@ The relayer is configured via a JSON file, the path to which is passed in via th
...
+  - The maximum amount of gas that can be used in a single block on this blockchain. The relayer will not attempt to deliver messages that require more gas than this limit to the given chain.
+
+  `"max-base-fee": unsigned integer`
+
+  - The maximum base fee gas price (in WEI) the relayer is willing to pay on this blockchain.
+
+  `"max-priority-fee-per-gas": unsigned interger`
+
+  - The maximum priority fee per gas (in WEI) that the relayer is willing to pay to incentivize transactions being included on this blockchain. The relayer will use the current estimation of the required gas tip cap for this blockchain, up to a maximum of this configured value.
+
```
cam-schultz, line 323: Can we document the default values for `block-gas-limit` and `max-priority-fee-per-gas`?

For `max-base-fee`, can we add that it is estimated using `EstimateBaseFee` if unset?

cam-schultz (approved):
One small suggestion, and a comment. Otherwise LGTM

vms/evm/destination_client.go
michaelkaplan13: Yeah, I think what we were doing before wasn't wrong, but likely was incorrect for what we wanted.

We setting:
```
maxFeePerGas = (baseFeeEstimate*3) + 2.5 gwei
maxPriorityFeePerGas = gasTipEstimate
```
which means if the `gasTipEstimate` was high, we would pay it assuming the actual base fee wasn't actually 3x the estimate.

michaelkaplan13: Now it's:

```
maxBaseFee = configured_max, or if unset,  estimate*3
maxTip = min(estimate, configured max)

maxFeePerGas = maxBaseFee + maxTip
maxPriorityFeePerGas = maxTip
```

relayer/README.md
michaelkaplan13: Added 👍

vms/evm/destination_client.go
cam-schultz: > ```
> maxBaseFee = configured_max, or if unset,  estimate*3
> maxTip = min(estimate, configured max)
> 
> maxFeePerGas = maxBaseFee + maxTip
> maxPriorityFeePerGas = maxTip
> ```

This would be a good top-level comment for `SendTx`

iansuvak (approved):
LGTM -- I agree with @cam-schultz about the summary being an appropriate comment for SendTx to make it easier to parse how these are set in the future.

iansuvak (approved):

PR #759, opened 2025-04-21
Allow tx revert if message already delivered

### LINKED ISSUE #756: Handle Teleporter message already delivered case
**Context and scope**
The relayer marks itself as unhealthy (and is killed) when a tx reverts. One expected case for this is if a separate relayer delivers a Teleporter message first, in which case the tx will revert with `"TeleporterMessenger: message already received"`. We should handle this case gracefully, and mark the block as successfully processed.

One way to do this is to call `Teleporter.messageReceived` after a revert. If that returns `true`, eat the error and continue processing as normal.

**Open questions**
Questions that are still being discussed.

cam-schultz:
## Why this should be merged
Absorbs the error if a tx fails due to a Teleporter message already having been delivered. This is a common scenario if multiple relayers are configured to relay between the same two chains.

Fixes #756 

## How this works

On tx revert, calls `TeleporterMessenger.messageReceived`. Emits a log and continues if `true`.

## How this was tested

New unit test. Testing this e2e is difficult because we always call `ShouldSendMessage` first, which similarly checks the message status by calling `TeleporterMessenger.messageReceived`.

## How is this documented
N/A


messages/teleporter/message_handler_test.go
```diff
@@ -254,3 +255,90 @@ func TestShouldSendMessage(t *testing.T) {
 		})
 	}
 }
+
+func TestSendMessageAlreadyDelivered(t *testing.T) {
```
iansuvak, line 259: nit: non-blocking

I do think that this test could use some commenting to make it easier to parse with all of the mocking happening as well as possibly renaming the `ethClient` to `mockEthClient`

iansuvak (approved):

messages/teleporter/message_handler_test.go
cam-schultz: Cleaned this up a bit. There's more opportunity to dedpulicate logic between the `SendMessage` and `ShouldSendMessage` tests, but I'll defer that for now until we have a more comprhensive `SendMessage` test suite.

iansuvak (approved):

bernard-avalabs (approved):

messages/teleporter/message_handler.go
```diff
@@ -305,10 +305,32 @@ func (m *messageHandler) SendMessage(
...
+		delivered, err := teleporterMessenger.MessageReceived(&bind.CallOpts{}, m.teleporterMessageID)
+		if err != nil {
+			m.logger.Error(
+				"Failed to check if message has been delivered to destination chain.",
+				zap.Error(err),
+			)
+			return common.Hash{}, fmt.Errorf("failed to check if message has been delivered: %w", err)
+		}
+		if delivered {
+			m.logger.Info("Execution reverted: message already delivered to destination.")
```
michaelkaplan13, line 324: Nit: Should we add the tx ID and message ID fields to this log?

messages/teleporter/message_handler.go
```diff
@@ -305,10 +305,32 @@ func (m *messageHandler) SendMessage(
...
+			return txHash, nil
+		}
+
+		m.logger.Error(
```
michaelkaplan13, line 328: I know this log was already existing, but should we add the blockchain ID field to for easier debugging going forward?

michaelkaplan13 (approved):
To logging nits, but LGTM

messages/teleporter/message_handler.go
```diff
@@ -305,10 +305,32 @@ func (m *messageHandler) SendMessage(
...
 
 	// Wait for the message to be included in a block before returning
-	err = m.waitForReceipt(signedMessage, m.destinationClient, txHash, m.teleporterMessageID)
+	receipt, err := m.waitForReceipt(signedMessage, m.destinationClient, txHash, m.teleporterMessageID)
 	if err != nil {
 		return common.Hash{}, err
 	}
+	if receipt.Status != types.ReceiptStatusSuccessful {
+		// Check if the message has already been delivered to the destination chain
+		teleporterMessenger := m.getTeleporterMessenger()
```
geoff-vball, line 314: nit: I'd just inline this on the next line

messages/teleporter/message_handler.go
geoff-vball: We add the warp message ID and the destination blockchain to all logs in the message handler [using logger.With()](https://github.com/ava-labs/icm-services/blob/main/messages/off-chain-registry/message_handler.go#L82), but we should definitely add the tx ID here.

messages/teleporter/message_handler.go
geoff-vball: See [my comment here](https://github.com/ava-labs/icm-services/pull/759/files?diff=unified&w=1#r2057296450), the blockchainID is already included. I'm not sure if there's a great way to document this though.

messages/teleporter/message_handler.go
cam-schultz: Added `txID` to the logs, and updated one other spot to use the same key.

michaelkaplan13 (approved):

PR #760, opened 2025-04-22
Fetch pending nonce

### LINKED ISSUE #755: Fetch pending nonce at startup
**Context and scope**
On startup, we fetch the configured account's _current_ nonce, which reflects the number of transactions that have been accepted from that address. However, in high throughput scenarios, it's possible for the relayer to issue a large number of transactions that remain in the mempool after the relayer times out waiting for tx acceptance and is killed. On startup, the current nonce does not reflect the pending tx's, causing nonce mismatch reverts as the pending txs begin to be accepted.

Fetch the nonce using the `pending` flag of the `eth_getTransactionCount` method call on startup

**Open questions**
Are any changes needed on subnet-evm / coreth? Or should we modify the client call on our end?

cam-schultz:
## Why this should be merged
Fixes #755 by fetching the pending nonce rather than the latest nonce on startup.

## How this works
Uses `rpc. PendingBlockNumber` in the call to `eth_getTransactionCount`

## How this was tested
CI

## How is this documented
N/A

### DIFF
--- a/tests/e2e_test.go
+++ b/tests/e2e_test.go
@@ -186,11 +186,11 @@ func cleanup() {
 var _ = ginkgo.AfterSuite(cleanup)
 
 var _ = ginkgo.Describe("[ICM Relayer Integration Tests", func() {
-	ginkgo.It("Manually Provided Message", func() {
-		ManualMessage(localNetworkInstance, teleporterInfo)
-	})
 	ginkgo.It("Basic Relay", func() {
 		BasicRelay(localNetworkInstance, teleporterInfo)
+	})
+	ginkgo.It("Manually Provided Message", func() {
+		ManualMessage(localNetworkInstance, teleporterInfo)
 	})
 	ginkgo.It("Shared Database", func() {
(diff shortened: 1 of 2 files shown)


tests/e2e_test.go
```diff
@@ -186,12 +186,12 @@ func cleanup() {
 var _ = ginkgo.AfterSuite(cleanup)
 
 var _ = ginkgo.Describe("[ICM Relayer Integration Tests", func() {
-	ginkgo.It("Manually Provided Message", func() {
-		ManualMessage(localNetworkInstance, teleporterInfo)
-	})
 	ginkgo.It("Basic Relay", func() {
 		BasicRelay(localNetworkInstance, teleporterInfo)
 	})
+	ginkgo.It("Manually Provided Message", func() {
```
iansuvak, line 192: I don't think it matters but just curious why this move?

iansuvak (approved):

geoff-vball (approved):

PR #762, opened 2025-04-23
configurable block acceptance timeout

### LINKED ISSUE #761: Configurable wait for `TransactionReceipt`
**Context and scope**
Currently the wait time for `eth_getTransactionReceipt` is hardcoded to 30 seconds through `defaultBlockAcceptanceTimeout` 

During high volume times or at times when the gas price is above what the relayer operator submitted the tx with it's possible for this limit to be exceeded while the transaction is waiting in the mempool. We should make the timeout configurable to allow operators to set timeouts with accordance to their preference based on their configured max fee (see #754) and throughput expectations. 

It is knowable whether the transaction is expected to be included based on the max fee provided and the current gas prices so the wait time could be adjusted based on that but this is likely an overkill and allowing operators to set this timeout and the fees separately should be enough to achieve desired customization. 

**Open questions**

iansuvak:
## Why this should be merged
Closes #761 
## How this works

- Adds a new configuration option for the destination clients that if not provided defaults to the existing constant of 30 seconds.
- Replaces the inner timeout on the actual call to `TransactionReceipt` with a default RPC timeout and sets the outer one to this new configurable value

## How this was tested
- Existing CI should pass
- Should probably add a test to `SendTx` but that should test cases beyond just this timeout

## How is this documented
- New field documented in the README

### DIFF
--- a/messages/teleporter/message_handler.go
+++ b/messages/teleporter/message_handler.go
@@ -352,7 +352,7 @@ func (m *messageHandler) waitForReceipt(
 		receipt, err = destinationClient.Client().(ethclient.Client).TransactionReceipt(callCtx, txHash)
 		return err
 	}
-	err := utils.WithRetriesTimeout(m.logger, operation, destinationClient.BlockAcceptanceTimeout(), "waitForReceipt")
+	err := utils.WithRetriesTimeout(m.logger, operation, destinationClient.TxInclusionTimeout(), "waitForReceipt")
 	if err != nil {
 		m.logger.Error(
 			"Failed to get transaction receipt",
--- a/messages/teleporter/message_handler.go
+++ b/messages/teleporter/message_handler.go
@@ -29,10 +29,6 @@ import (
 	"google.golang.org/grpc"
 )
 
-const (
-	defaultBlockAcceptanceTimeout = 30 * time.Second
-)
-
 type factory struct {
 	messageConfig   *Config
 	protocolAddress common.Address
@@ -329,12 +325,12 @@ func (m *messageHandler) waitForReceipt(
 ) error {
 	var receipt *types.Receipt
 	operation := func() (err error) {
-		callCtx, callCtxCancel := context.WithTimeout(context.Background(), defaultBlockAcceptanceTimeout)
+		callCtx, callCtxCancel := context.WithTimeout(context.Background(), utils.DefaultRPCTimeout)
 		defer callCtxCancel()
 		receipt, err = destinationClient.Client().(ethclient.Client).TransactionReceipt(callCtx, txHash)
 		return err
 	}
-	err := utils.WithRetriesTimeout(m.logger, operation, defaultBlockAcceptanceTimeout, "waitForReceipt")
+	err := utils.WithRetriesTimeout(m.logger, operation, destinationClient.BlockAcceptanceTimeout(), "waitForReceipt")
 	if err != nil {
 		m.logger.Error(
 			"Failed to get transaction receipt",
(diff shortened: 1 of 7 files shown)


vms/evm/destination_client.go
```diff
@@ -9,27 +9,29 @@ import (
 	"context"
 	"math/big"
 	"sync"
+	"time"
 
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/utils/logging"
 	avalancheWarp "github.com/ava-labs/avalanchego/vms/platformvm/warp"
-	"github.com/ava-labs/coreth/rpc"
```
iansuvak, line 16: Unrelated to this PR but swapped `subnet-evm/rpc` in for `coreth/rpc` since my IDE was complaining that coreth should be direct dependency and this was the only place it was required explicitly.

Nor sure why the `go mod tidy` check didn't pick it up in CI

relayer/README.md
```diff
@@ -309,6 +309,10 @@ The relayer is configured via a JSON file, the path to which is passed in via th
 
   - The AWS region in which the KMS key is located. Required if `kms-key-id` is provided.
 
+  `"block-acceptance-timeout-seconds": uint64`
```
michaelkaplan13, line 312: Would a more accurate name for this be `tx-inclusion-timeout-seconds`?

I know we already had an existing variable named `defaultBlockAcceptanceTimeout`, but I this it's a bit of a misnomer.

vms/evm/destination_client.go
```diff
@@ -9,27 +9,29 @@ import (
...
 )
 
 const (
 	// Set the max fee to twice the estimated base fee.
 	// TODO: Revisit this constant factor when we add profit determination, or make it configurable
-	BaseFeeFactor        = 2
-	MaxPriorityFeePerGas = 2500000000 // 2.5 gwei
+	BaseFeeFactor                 = 2
+	MaxPriorityFeePerGas          = 2500000000 // 2.5 gwei
+	DefaultBlockAcceptanceTimeout = 30 * time.Second
```
michaelkaplan13, line 34: Should we move setting the default to `DestinationBlockchain.Validate()` like we have now for the block gas limit and max priority fee per gas? That will also simplify the initialization logic below.

michaelkaplan13 (commented):
Two small suggestions but otherwise LGTM

geoff-vball (commented):
I agree with Michael's suggestions. Otherwise LGTM

messages/teleporter/message_handler.go
```diff
@@ -329,12 +325,12 @@ func (m *messageHandler) waitForReceipt(
 ) error {
 	var receipt *types.Receipt
 	operation := func() (err error) {
-		callCtx, callCtxCancel := context.WithTimeout(context.Background(), defaultBlockAcceptanceTimeout)
```
cam-schultz, line 332: Nice catch

cam-schultz (commented):
LGTM once other comments are addressed

bernard-avalabs (approved):

iansuvak:
> We should include the tx acceptance timeout in the overall `CreateSignedMessage` context [timeout](https://github.com/ava-labs/icm-services/blob/main/relayer/application_relayer.go#L225)

I'm not sure I understand what you mean here. The txAcceptanceTimeout is strictly after the message has been signed and sent already.

geoff-vball (approved):

cam-schultz:
> I'm not sure I understand what you mean here. The txAcceptanceTimeout is strictly after the message has been signed and sent already.

Brainfart, my bad. Disregard

cam-schultz (approved):

Issue #758, opened 2025-04-21
Reduce RPC call frequency


cam-schultz:
**Context and scope**
We make a handful of RPC calls on the `CreateSignedMessage` critical path that can kill performance under load. We should aim to minimize the expected number of RPC calls per message. Below are the calls we make in `CreateSignedMessage`, along with possible techniques to remove or reduce their frequency on the critical path.

- `platformvm.GetSubnetID`. Called on the first time a message is relayed from a given subnet ID. We guard this with a global lock and then cache the result, so there's not much more to optimize here
- `platformvm.GetSubnet`. Called to determine if the signing subnet is an L1.
  - Once a subnet is converted to an L1, it can never be converted back. Once we determine it's an L1, we should cache that result and use it for all future messages, rather than making an RPC call
  - For non-L1 subnets, conversion is an extremely rare event (in fact, it can only happen once), so incurring the cost of an RPC call to dynamically check this on every message is overly aggressive. Similar to the L1 case, caching the first result and using that for the lifetime of the application is more practical. The worst case is that signature verification fails, the relayer is killed, and it succeeds on startup once we re-query the subnet's status.
- `platformvm.GetCurrentL1Validators`. Called for each messag from an L1 to determine the balance of L1 validators, and exclude low balance ones from the aggregate signature.
  - Validator balances change only as often as P-Chain blocks are produced. Instead of querying validators every message, we could cache the returned value and eject it after 2 seconds (the P-Chain average block time). If there's a cached result, use that, otherwise re-query
- `platformvm.GetValidatorsAt`. Used to fetch the connected Warp validators to construct the aggregate signature.
  - We could employ a similar validator set caching strategy as `platformvm.GetCurrentL1Validators`.

**Discussion and alternatives**
The above items should be selectively implemented based on load test experiment results.

PR #763, opened 2025-04-23
Cache p-chain calls for IsL1 checks

iansuvak:
## Why this should be merged
Partially addresses #758, separate PR will address the remaining P-chain calls since caching those is likely going to produce more discussion and need for testing.  
## How this works

- Added a new lock protected map field to cache the result
- Reusing the same lock since we don't want to make this call more than once and when it's cached it's only locked for a negligible time to do the map lookup. 

## How this was tested
Existing CI 
## How is this documented
added a comment to the `SignatureAggregator` struct that the field is protected by the now multipurpose `subnetMapsLock`

### DIFF
--- a/signature-aggregator/aggregator/aggregator.go
+++ b/signature-aggregator/aggregator/aggregator.go
@@ -61,18 +61,19 @@ var (
 )
 
 type SignatureAggregator struct {
-	network peers.AppRequestNetwork
-	// protected by subnetMapsLock
-	subnetIDsByBlockchainID map[ids.ID]ids.ID
-	// protected by subnetMapsLock
-	subnetIDIsL1        map[ids.ID]bool
+	network             peers.AppRequestNetwork
 	messageCreator      message.Creator
 	currentRequestID    atomic.Uint32
-	subnetMapsLock      sync.Mutex
 	metrics             *metrics.SignatureAggregatorMetrics
 	cache               *cache.Cache
 	pChainClient        platformvm.Client
 	pChainClientOptions []rpc.Option
+
+	subnetMapsLock sync.Mutex
+
+	// following block of fields is protected by the subnetMapsLock
+	subnetIDsByBlockchainID map[ids.ID]ids.ID
+	subnetIDIsL1            map[ids.ID]bool
 }
 
 func NewSignatureAggregator(
--- a/signature-aggregator/aggregator/aggregator.go
+++ b/signature-aggregator/aggregator/aggregator.go
@@ -62,15 +62,17 @@ var (
 
 type SignatureAggregator struct {
 	network peers.AppRequestNetwork
-	// protected by subnetsMapLock
+	// protected by subnetMapsLock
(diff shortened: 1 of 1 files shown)


signature-aggregator/aggregator/aggregator.go
```diff
@@ -62,15 +62,17 @@ var (
...
+	// protected by subnetMapsLock
 	subnetIDsByBlockchainID map[ids.ID]ids.ID
-	messageCreator          message.Creator
-	currentRequestID        atomic.Uint32
-	subnetsMapLock          sync.Mutex
-	metrics                 *metrics.SignatureAggregatorMetrics
-	cache                   *cache.Cache
-	pChainClient            platformvm.Client
-	pChainClientOptions     []rpc.Option
+	// protected by subnetMapsLock
```
geoff-vball, line 67:
```suggestion
```

signature-aggregator/aggregator/aggregator.go
```diff
@@ -522,6 +519,28 @@ func (s *SignatureAggregator) getSubnetID(
 	return subnetID, nil
 }
 
+// Looks up whether a subnet is an L1 and caches the result in the map for the lifetime of the application.
+// since this value can change only once in the lifetime of the subnet.
```
geoff-vball, line 523: What's the failure mode if the subnet converts to an L1 while the signature aggregator is running?

signature-aggregator/aggregator/aggregator.go
geoff-vball: This line currently appears twice

cam-schultz: I believe this is to indicate that `subnetIDIsL1` specifically is protected by that lock. We could make this a bit more readable by grouping the protected fields together.

signature-aggregator/aggregator/aggregator.go
cam-schultz: Most likely ICM verification would fail, though we wouldn't be able to tell directly from the logs. Given that subnet conversion is rare and we'd likely know ahead of time when it occurs for any running relayers, this seems like an acceptable tradeoff to me.

cam-schultz (approved):
LGTM

signature-aggregator/aggregator/aggregator.go
michaelkaplan13: I agree on the tradeoff for now, but think we should explicitly document this somewhere.

Alternatively, could have some reasonable TTL for subnets that are not L1s to recheck.

bernard-avalabs (commented):
It may be prudent to add expiries to the entries in subnetIDIsL1, especially for those that are currently subnets as they may convert to L1s in the future. Otherwise looks good to me.

signature-aggregator/aggregator/aggregator.go
geoff-vball: We should only need to expire `false` entries, right?

geoff-vball (approved):

PR #764, opened 2025-04-24
Throttle account txs

### LINKED ISSUE #322: Throttle single signer transactions
**Context and scope**
As a DOS protection mechanism, Avalanche nodes are commonly configured to drop transactions from the mempool with a nonce greater than the latest accepted nonce by [some amount](https://build.avax.network/docs/nodes/chain-configs/c-chain#tx-pool-account-slots) (by default this is 16). If, say 17 transactions from the same account are submitted at once, then this limit will be exceeded, and the 17th transaction will be dropped from the mempool. This would cause the corresponding Warp message to fail to be delivered. #321 ensures that this is recoverable, but we should implement a throttling or queueing strategy to prevent this scenario altogether. #256 also mitigates this scenario by distributing the load across many keys.

**Open questions**
The relayer would need to monitor block production on the destination so that it can distribute transactions over multiple blocks. A time based approach may suffice, but relies on assumptions of the destination's consensus parameters. A more robust approach would be to listen for new blocks and meter transactions at the same rate that blocks are produced.

cam-schultz:
## Why this should be merged
Fixes #322

## How this works

Adds a semaphore of size 16 (the default value of [tx-pool-account-slots](https://build.avax.network/docs/nodes/chain-configs/c-chain#tx-pool-account-slots)) that is acquired when a tx is submitted, and released once it is accepted.

On startup, we fetch the current nonce and pending nonce to get the number of pending txs in the mempool. We acquire this number of semaphores. We then start a goroutine to monitor the progress of the current nonce and release the initially acquired semaphores as txs are processed.

## How this was tested
CI

## How is this documented
N/A


vms/evm/destination_client.go
```diff
@@ -109,7 +154,7 @@ func NewDestinationClient(
 		"Initialized destination client",
 		zap.String("blockchainID", destinationID.String()),
 		zap.String("evmChainID", evmChainID.String()),
-		zap.Uint64("nonce", nonce),
+		zap.Uint64("nonce", pendingNonce),
```
iansuvak, line 157: should we report the gap here as well? i.e both the pendingNonce and the current nonce at the time of client initialization?

vms/evm/destination_client.go
```diff
@@ -86,16 +90,57 @@ func NewDestinationClient(
...
+
+	numPendingTxs := max(currentNonce-pendingNonce, 0)
+
+	poolTxsSemaphore := make(chan struct{}, poolTxsPerAccount)
+	for i := uint64(0); i < numPendingTxs; i++ {
+		poolTxsSemaphore <- struct{}{}
+	}
+
+	// Start a goroutine to release the numPendingTxs semaphores once the pending txs are accepted
+	go func() {
```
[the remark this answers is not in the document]
cam-schultz: This should never happen as long as the pending nonce does not increment in between the two calls, which is the case since we assume the relayer has exclusive use of its private key. When we first fetch the current nonce, that's guaranteed to be the number of pending txs.

vms/evm/destination_client.go
```diff
@@ -86,16 +90,57 @@ func NewDestinationClient(
...
+			"Failed to get current nonce",
+			zap.Error(err),
+		)
+		return nil, err
+	}
+
+	numPendingTxs := max(currentNonce-pendingNonce, 0)
+
+	poolTxsSemaphore := make(chan struct{}, poolTxsPerAccount)
+	for i := uint64(0); i < numPendingTxs; i++ {
```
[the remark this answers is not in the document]
cam-schultz: Good call out. I handled this by counting the number of pending txs that are in excess of `poolTxsPerAccount`, and gating semaphore access behind that. If we are in that scenario, the semaphore can't be acquired until that counter reaches zero, during which time all relayers will be blocked. This results in us properly processing any number of pending txs, but maxing out at `poolTxsPerAccount` in steady state.

iansuvak (commented):
This solution to the overrun transactions is a lot cleaner/simpler than what I was initially imagining.

LGTM

vms/evm/destination_client.go
```diff
@@ -88,43 +93,111 @@ func NewDestinationClient(
...
+			zap.Uint64("overrunPendingTxs", overrunPendingTxs),
+		)
+		for i := uint64(0); i < numPendingTxs-overrunPendingTxs; i++ {
+			poolTxsSemaphore <- struct{}{}
+		}
+
+		// Asynchronously release the numPendingTxs semaphores once the pending txs are accepted
+		go func() {
+			initialCurrentNonce := currentNonce
+			for range time.Tick(2 * time.Second) {
```
geoff-vball, line 150: I'd pull this out as a constant

vms/evm/destination_client.go
```diff
@@ -88,43 +93,111 @@ func NewDestinationClient(
...
+						zap.Error(err),
+					)
+					continue
+				}
+				processedPendingTxs := currentNonce - initialCurrentNonce
+				logger.Info(
+					"Processed pending transactions on startup",
+					zap.Uint64("processedPendingTxs", processedPendingTxs),
+				)
+				for i := uint64(0); i < processedPendingTxs; i++ {
```
geoff-vball, line 164: I think this currently has a bug. If our `initialCurrentNonce` is 10, and our `currentNonce` is 11, and we have 10 pending txs, every tick `processedPendingTxs` will be 1, meaning we will think we have processed one tx every tick even if we haven't.

vms/evm/destination_client.go
```diff
@@ -181,10 +254,9 @@ func (c *destinationClient) SendTx(
 	gasFeeCap := new(big.Int).Add(maxBaseFee, gasTipCap)
 
 	// Synchronize nonce access so that we send transactions in nonce order.
-	// Hold the lock until the transaction is sent to minimize the chance of
-	// an out-of-order transaction being dropped from the mempool.
-	c.lock.Lock()
-	defer c.lock.Unlock()
+	// Hold the lock until the transaction is sent, otherwise we may deadlock
+	// on poolTxsSemaphore if there is a nonce gap in the mempool.
+	c.nonceLock.Lock()
```
geoff-vball, line 259: Without a deferred unlock, we run into an issue here if we return from an error because we won't release the lock. I would probably recommend encapsulating the critical section in a helper function so that we can defer the unlock

vms/evm/destination_client.go
```diff
@@ -208,26 +280,76 @@ func (c *destinationClient) SendTx(
...
+
+	c.logger.Info(
+		"Sending transaction",
+		zap.String("txID", signedTx.Hash().String()),
+		zap.Uint64("nonce", c.currentNonce),
+		zap.Int("poolTxSlotsAvailable", cap(c.poolTxsSemaphore)-len(c.poolTxsSemaphore)),
+	)
+
+	// Acquire a semaphore to limit the number of transactions in the mempool
+	c.poolTxsSemaphore <- struct{}{}
```
geoff-vball, line 297: The same goes with this. If this function errors, we never release the semaphore.

vms/evm/destination_client.go
```diff
@@ -64,6 +67,8 @@ func NewDestinationClient(
 		return nil, err
 	}
 
+	logger = logger.With(zap.String("blockchainID", destinationBlockchain.BlockchainID))
```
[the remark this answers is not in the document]
geoff-vball: Right, got it, makes sense!

vms/evm/destination_client.go
cam-schultz: Addressed in https://github.com/ava-labs/icm-services/pull/767

vms/evm/destination_client.go
cam-schultz: No longer relevant with https://github.com/ava-labs/icm-services/pull/767

geoff-vball (approved):

PR #765, opened 2025-04-25
implement a cache for outstanding P-chain calls

### LINKED ISSUE #758: Reduce RPC call frequency

iansuvak:
## Why this should be merged
Closes #758 

## How this works
- Implements a generic TTL cache and uses it for the two outstanding P-chain calls to get L1 validator balances and canonical validator sets
- Sets both constants to be 2 seconds independently to match average block time on p-chain as discussed although looking at live blocks it looks to be much slower than that indicating that we can probably cache them more aggressively. 
- Renames the existing cache to signatureCache since it's not generic and moves it into the aggregator package scope

## How this was tested
Existing CI
New  unit tests for the TTLCache struct

## How is this documented


peers/app_request_network.go
```diff
@@ -53,6 +54,9 @@ const (
 	// This value is defined in avalanchego peers package
 	// TODO: use the avalanchego constant when it is exported
 	maxNumSubnets = 16
+
+	// The amount of time to cache canonical validator sets in seconds
+	canonicalValidatorSetCacheTTL = uint64(2)
```
geoff-vball, line 59: Can we include `seconds` in the name here?

.gitignore
```diff
@@ -11,9 +11,6 @@ main.log
 server.log
 *.test
 
-# Foundry outputs
-cache/
-out/
```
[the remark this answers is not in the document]
iansuvak: Yeah, since I did the re-org of calling this new top level package `cache` I did it to not require force adding an ignored folder. I assumed that this `.gitignore` set of lines was copied over from `icm-contracts`. I don't think that we do contract compilation or forge tests in this repo.

bernard-avalabs (approved):
LGTM. 

For our use case, I don't think cleanup of stale/unused entries are necessary. Otherwise we would need to add a minheap to track expiry times.

geoff-vball (approved):

cache/ttl_cache.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+}
+
+// GetValue checks if the cached value is fresh for a given key, otherwise fetches
+// the value from the fetchFunc and caches it.
+// If skipCache is true, the value will be fetched from the fetchFunc regardless of
+// whether it is fresh or not and stored in the cache.
+func (c *TTLCache[K, V]) Get(key K, fetchFunc func(K) (V, error), skipCache bool) (V, error) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
```
cam-schultz, line 36: I think we should optimize lock acquisition (i.e. use a RWLock and only acquire the write lock for keys whose TTL has expired), since the `fetchFunc` may be long-running. We also shouldn't hold the lock while `fetchFunc` is running so that we don't block read lock acquisition.

cam-schultz (commented):
A couple of relatively minor comments, but this seems like a sensible approach. We can make some of these P-Chain API calls fully async in the future, but amortizing the number of API calls per message is the biggest win.

cache/ttl_cache.go
geoff-vball: So, if a goroutine is trying to get a validator set that is currently being fetched, we actually do want it to wait at the read lock acquisition phase so that it can just read it from the cache once it's been fetched. The problem is that we don't want to block other chains from being accessed. 

I think this leads me to believe that we should have a lock per chain. We should probably use a `syncmap` inside our map to handle one lock per chian, and then use `LoadOrStore()` to either get the lock or create a new one if we haven't seen that chain before.

cam-schultz: Yup, a lock per key is what I was envisioning as well. We should be able to achieve the same with a single RW lock though

geoff-vball: A single RW lock would only be possible in the case where we're updating validator sets async on a timer. 

If we're doing it synchronously, if there are two goroutines that want the same info that's not in the cache, there's no way for the second goroutine to wait on the lock while still allowing the first to write the data once it's fetched. The other option for the second goroutine would be to fetch the data itself as well, which would be redundant.

cam-schultz: Ah yeah you're right. If we need to fetch a new value, the only way to prevent access to the stale key without forcing other key accesses to wait is with a key-specific lock.

iansuvak: So haven't used this before but came across the `singleflight.Group` while looking for alternative ways to implement this. It looks like it gives us the locking semantics that we want with the cost of restricting down to meaningfully stringable types which should be fine for our only usecases right now restricting to `ids.ID` for `subnetID`.

To me this seems easier to reason about as we only have the single map now and the lock just protects the write access here while we use another structure to dedupe multiple requests on a per-key basis while not holding a lock any longer than the duration of the write.

cache/ttl_cache.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// Get checks if the cached value is fresh for a given key, otherwise fetches
+// the value using fetchFunc. Concurrent fetches for the same key are deduplicated.
+// If skipCache is true, the value will be fetched regardless of cache state,
+// but concurrent fetches for the same key during skipCache are still deduplicated.
+func (c *TTLCache[K, V]) Get(key K, fetchFunc func(K) (V, error), skipCache bool) (V, error) {
+	if !skipCache {
+		c.lock.Lock()
+		item, exists := c.data[key]
```
cam-schultz, line 41: This can be simplified by releasing the lock immediately after accessing the map.

cam-schultz: Alternatively, we can consider using a [sync.Map](https://pkg.go.dev/sync#Map) instead.

cache/ttl_cache.go
```diff
@@ -0,0 +1,80 @@
...
+		if exists && time.Since(item.timestamp) < c.ttl {
+			c.lock.Unlock()
+			return item.value, nil
+		}
+		c.lock.Unlock()
+	}
+
+	keyStr := keyToString(key)
```
cam-schultz, line 49: Seems simpler to replace `K` with `string` in the `TTLCache` definition. Otherwise, you could get some odd behavior with`fmt.Sprintf("%v", key)`, for instance if a `key` is a pointer to a slice that is reallocated at a different address.

cache/ttl_cache.go
```diff
@@ -0,0 +1,80 @@
...
+		data: make(map[K]TTLCacheItem[V]),
+		ttl:  ttl,
+	}
+}
+
+// Get checks if the cached value is fresh for a given key, otherwise fetches
+// the value using fetchFunc. Concurrent fetches for the same key are deduplicated.
+// If skipCache is true, the value will be fetched regardless of cache state,
+// but concurrent fetches for the same key during skipCache are still deduplicated.
+func (c *TTLCache[K, V]) Get(key K, fetchFunc func(K) (V, error), skipCache bool) (V, error) {
```
cam-schultz, line 38: As @geoff-vball suggested, I think removing `skipCache` in favor of manually ejecting the validator set on Warp verification failure would be easier to reason about.

iansuvak: Yeah thinking more about it I did like how the cache logic was self-contained here. But I don't feel strongly and I do think that invalidating does save us potentially some requests that are doomed to fail and will waste gas -- will make the change

cache/ttl_cache.go
iansuvak: IMO using plain slices as cache keys is odder than the behavior this would produce which would just be preventing singleflight from deduping. 

To keep the cache general I do like using generics and comparable here. Originally I had this accept a `ComparableStringer` type which was `comparable fmt.Stringer` which is really equivalent to `fmt.Stringer` that you don't have to call `String()` on before putting in the map. The downside of that is that is that it prevents primitive `string` type and its aliases from being used as the key as they don't implement Stringer and go doesn't let you define a type interface that is union of fmt.Stringer and string.

My vote is to either keep it as is or use `ComparableStringer` that way if you really want to use a string type you can add another alias that implements `String()` as a pass through and then not use `fmt.Sprintf` at all. 

If your preference for key being string I am happy to do that as well.

geoff-vball: Just chiming in to say I don't have a strong preference here

cache/ttl_cache.go
iansuvak: I opted to rename the `skipCache` flag to `invalidate` and do the delete to get the benefit of not returning known stale values but keeping the implementation simple. 

Manual explicit eviction from the cache would complicate the implementation since these are two different instances of caches on two different structs (aggregator and the network) so doing this would require modifying the interfaces of both and the point at which the signature verification fails is far removed from where these cache values are being accessed.

cache/ttl_cache.go
iansuvak: With the deletes and updates happening fairly often I don't think that paying sync.Map overhead is worth it to avoid the very tight and simple locking that is implemented here after the feedback.

cache/ttl_cache.go
cam-schultz: > go doesn't let you define a type interface that is union of fmt.Stringer and string.

TIL: https://github.com/golang/go/issues/45346#issuecomment-862505803

I'm not seeing any application code that instantiates the cache with a raw `string`. My preference would be to make the interface more restrictive versus permissive, especially considering that we resolve everything to a string internally anyway. Forcing the caller to be aware of this up-front seems safer to me.

Hopefully a future Go version will allow us to define a more inclusive interface.

cam-schultz (commented):
One comment on the TTLCache type specifier. I think the change from `skipCache -> invalidate` is more readable. Agreed that keeping the invalidation at message processing retry scope is fine for now; doing that closer to tx verification failure would require some refactoring.

cam-schultz (approved):

geoff-vball (approved):

PR #767, opened 2025-04-29
throttle tx usage using condition variable

cam-schultz:
## Why this should be merged
Modifies https://github.com/ava-labs/icm-services/pull/764 to synchronize nonce access and transaction issuance using a condition variable rather than a decoupled lock and mutex.

The only functional difference is that on startup we block until all pending transactions have been processed, rather than handling that async. This has no meaningful effect on message relaying latency, since any new relay requests must wait for pending transactions to be accepted anyway before the message may be delivered in a new transaction. 

## How is this documented

### DIFF
--- a/vms/evm/destination_client.go
+++ b/vms/evm/destination_client.go
@@ -41,7 +41,7 @@ type Client interface {
 // Implements DestinationClient
 type destinationClient struct {
 	client                  ethclient.Client
-	nonceLock               *sync.Mutex
+	nonceCond               *sync.Cond
 	destinationBlockchainID ids.ID
 	signer                  signer.Signer
 	evmChainID              *big.Int
@@ -50,8 +50,8 @@ type destinationClient struct {
 	maxBaseFee              *big.Int
 	maxPriorityFeePerGas    *big.Int
 	logger                  logging.Logger
-	poolTxsSemaphore        chan struct{}
 	txInclusionTimeout      time.Duration
+	numPendingTxs           int
 }
 
 func NewDestinationClient(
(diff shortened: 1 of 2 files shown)


bernard-avalabs (commented):
This is much easier to reason about than the previous semaphore-based solution.

vms/evm/destination_client.go
```diff
@@ -349,6 +326,12 @@ func (c *destinationClient) waitForReceipt(
 		)
 		return nil, err
 	}
+
+	c.nonceCond.L.Lock()
+	defer c.nonceCond.L.Unlock()
+	c.numPendingTxs--
```
iansuvak, line 332: [no action necessary] If we use `sync/atomic` for `c.numPendingTxs` we can decrement it without grabbing the lock. Not sure if there are downsides to doing the `Signal()` without holding the lock but it seems correct to me.

iansuvak (approved):
LGTM and I agree that the Cond implementation + doing startup handling separately without concurrency concerns is a lot cleaner to reason about.

vms/evm/destination_client.go
```diff
@@ -102,106 +102,128 @@ func NewDestinationClient(
...
-			zap.Error(err),
-		)
-		return nil, err
-	}
-
-	numPendingTxs := max(currentNonce-pendingNonce, 0)
-	if numPendingTxs > 0 {
-		// Defensively account for the case where the number of pending txs is greater than poolTxsPerAccount
-		overrunPendingTxs := uint64(max(int(numPendingTxs)-int(poolTxsPerAccount), 0))
+	// Block until all pending txs are accepted
```
geoff-vball, line 105: As implemented right now, the relayer is creating all destination sequentially, and blocking during the creating of each one until the pending and current nonces are synced. I don't think we want this behaviour.

vms/evm/destination_client.go
```diff
@@ -349,6 +326,12 @@ func (c *destinationClient) waitForReceipt(
 		)
 		return nil, err
 	}
+
+	c.nonceCond.L.Lock()
+	defer c.nonceCond.L.Unlock()
+	c.numPendingTxs--
+	c.nonceCond.Signal()
```
geoff-vball, line 333: This has the same issue as before, where if this function errors we never decrement `c.numPendingTxs`

vms/evm/destination_client.go
```diff
@@ -102,106 +102,128 @@ func NewDestinationClient(
...
+	gasTipCap *big.Int,
+	callData []byte,
+	signedMessage *avalancheWarp.Message,
+) (*types.Transaction, error) {
+	c.nonceCond.L.Lock()
+	defer c.nonceCond.L.Unlock()
+
+	for c.numPendingTxs >= poolTxsPerAccount {
+		c.nonceCond.Wait()
+	}
```
geoff-vball, line 173: I think this was true with the previous implementation as well, but we don't have a `fifo` property on the transactions that are waiting on the condition variable. If we already have 16 pending txs, and 4 more come along to be processed, the order they will get sent will be random. This miiight not be a problem? But I do think we should probably do our best to order messages sent to a particular chain if it's not too much hassle.

vms/evm/destination_client.go
cam-schultz: In the error case we mark the relayer as unhealthy and restart. It's ambiguous how to treat `numPendingTxs` if we error fetching the receipt, since we have no other way of knowing if the transaction was accepted.

vms/evm/destination_client.go
cam-schultz: Agreed, it's not optimal, but since it's only on startup, I may defer this to a later ticket.

vms/evm/destination_client.go
cam-schultz: The relayer makes no guarantees about transaction delivery order. We already concurrently process messages in an uncoordinated fashion. https://github.com/ava-labs/icm-services/blob/main/relayer/application_relayer.go#L173

vms/evm/destination_client.go
cam-schultz: To keep the synchronization logic simple, I'm going to keep `c.numPendingTxs` as a non-atomic type, and guard it using the `noncCond` lock. From the docs, 

> It is allowed but not required for the caller to hold c.L during the call [to `Signal`]

however, there's a readability benefit to using the same lock aquisition pattern for all calls to `Signal`

geoff-vball: @iansuvak I originally had the same thought, but there's a deadlock possibility if you decrement and signal between when a goroutine checks the cond var and when it waits. I don't think it would be an issue for us with a maximum of 16 pending txs, but becomes more of a problem as `poolTxsPerAccount` -> 1.

PR #771, opened 2025-05-06
Change usages of go:generate to automatically use version in go.mod file

geoff-vball:
## Why this should be merged
Removes the need for `generate.sh` and manually installing the correct version of `mockgen`.

## How this works
Using //go:generate go run go.uber.org/mock/mockgen instead of //go:generate mockgen will actually run the version of mockgen specified in the go.mod file so we don't have to jump through the extra hoops of parsing the correct version from the go.mod file and the using go install on that version.

## How is this documented


cam-schultz (approved):
Nice. IIUC, `//go:generate` just runs whatever command is provided in the environment of `go generate` call. I've never used `go run` to invoke a dependency directly, but it makes sense that it uses the `go.mod` in the same way as it would if we were using `go run` to build and run local code.

PR #773, opened 2025-05-07
PoC: multiple signers

geoff-vball:
## Why this should be merged
PoC

## How is this documented


iansuvak (commented):
My opinion isn't super strong but I do prefer this approach and eliminating the lock from the accountSigner makes it easier to reason about.

vms/evm/destination_client.go
```diff
@@ -239,29 +265,41 @@ func (c *destinationClient) SendTx(
...
+		}
+		cases = append(cases, reflect.SelectCase{
+			Dir:  reflect.SelectSend,
+			Chan: reflect.ValueOf(c.messageChans[i]),
+			Send: reflect.ValueOf(messageData),
+		})
+	}
+
+	// Select an available, eligible signer and acquire a txQueue slot
+	reflect.Select(cases)
```
cam-schultz, line 294: Can we preserve the debug logs for the eligible signers and the selected signer?

vms/evm/destination_client.go
```diff
@@ -144,17 +163,22 @@ func NewDestinationClient(
...
-					signer:       signer,
-					currentNonce: currentNonce,
-					lock:         &sync.Mutex{},
-					txQueue:      txQueue,
-				}
-				selectCases[i] = reflect.SelectCase{
-					Dir:  reflect.SelectSend,
-					Chan: reflect.ValueOf(txQueue),
-					Send: reflect.ValueOf(struct{}{}),
+					logger:            logger.With(zap.Stringer("address", signer.Address())),
```
cam-schultz, line 169:
```suggestion
					logger:            logger.With(zap.Stringer("senderAddress", signer.Address())),
```
To disambiguate from the called address

vms/evm/destination_client.go
```diff
@@ -270,146 +308,120 @@ func (c *destinationClient) SendTx(
...
-	// Select an available, eligible signer and acquire a txQueue slot
-	idx, _, _ := reflect.Select(cases)
-	if deliverers.Len() != 0 {
-		idx = idxs[idx]
+func (s *accountSigner) processIncomingTransactions() {
+	for {
+		// We can only only get to listen to readyChan if there is an open pending tx slot
+		s.txQueue <- struct{}{}
+		s.logger.Debug("Waiting for incoming transaction")
+		// TODO handle error
```
cam-schultz, line 317: Writing to `data.resultChan` seems reasonable here.

vms/evm/destination_client.go
```diff
@@ -173,18 +197,20 @@ func NewDestinationClient(
 		zap.Uint64("nonce", pendingNonce),
 	)
 
-	return &destinationClient{
+	destClient = destinationClient{
```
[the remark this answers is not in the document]
geoff-vball: I don't think this can be the case? We don't set any values on `destinationClient` until the end of the function, so the `accountSigner` would have none of the correct fields, but tests seem to be passing properly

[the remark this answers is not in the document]
geoff-vball: But for whatever reason my brain has dumped everything I knew about golang assignment lol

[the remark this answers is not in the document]
cam-schultz: Ah no you're right, the stack variable has a fixed address, and reassigning a new value won't move it.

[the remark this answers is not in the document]
cam-schultz: So the answer here is a bit more nuanced. Go uses "stack copying", meaning that the runtime can may be reallocated when necessary, moving all stack variables to new addresses.

However, the Go compiler also performs escape analysis to determine which stack variables are referenced outside the stack's scope, and instead allocated them on the heap. This ensures the stack variable is not changed in any way on stack copy.

I confirmed that the variable in question is escaped to the heap by running `go build -gcflags='-m' vms/evm/destination_client.go`.

vms/evm/destination_client.go
```diff
@@ -270,146 +302,114 @@ func (c *destinationClient) SendTx(
...
 
-func (c *destinationClient) waitForReceipt(
+func (s *accountSigner) waitForReceipt(
 	txHash common.Hash,
-	signerIdx int,
-) (*types.Receipt, error) {
+	resultChan chan messageResult,
+) {
+	// Release the txQueue slot once this function returns
+	defer func() { <-s.txQueue }()
```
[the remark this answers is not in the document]
iansuvak: True -- It's either still in the mempool, it got dropped, or got accepted and for some reason we can't reach our RPC endpoint. 

The simplest thing to do here is to fatal and hard restart but that's expensive. Might be acceptable though since `txInclusionTimeout` is configurable.

Alternatively can pass in the nonce here and check whether pending nonce is greater than or equal to the one we are waiting for still but doesn't seem worthwhile without an additional healing path and violating `txInclusionTimeout`

vms/evm/destination_client.go
```diff
@@ -239,29 +259,41 @@ func (c *destinationClient) SendTx(
...
-	)
-	if err != nil {
-		c.logger.Error(
-			"Failed to issue transaction",
-			zap.Stringer("messageID", signedMessage.ID()),
-			zap.Error(err),
-		)
-		return nil, err
+	resultChan := make(chan messageResult)
+	defer close(resultChan)
```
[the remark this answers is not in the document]
geoff-vball: I switched this to have the sender(s) close, but it feels kind of messy. I think I'd slightly lean towards having the receiver close and just comment it.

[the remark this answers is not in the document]
cam-schultz: Restricting channel close to the sender is more or less the language [standard](https://go.dev/tour/concurrency/4). I don't think our use case deviates from normal channel usage, so we should stick with this standard unless there's a specific reason not to.

vms/evm/destination_client.go
```diff
@@ -239,29 +265,46 @@ func (c *destinationClient) SendTx(
...
+
+	// Wait for the receipt to be returned
+	result := <-resultChan
+	if result.err != nil {
 		c.logger.Error(
 			"Failed to get transaction receipt",
-			zap.String("txID", signedTx.Hash().String()),
 			zap.Error(err),
 		)
 		return nil, err
```
iansuvak, line 310:
```suggestion
			zap.Error(result.err),
		)
		return nil, result.err
```

vms/evm/destination_client.go
```diff
@@ -239,177 +268,184 @@ func (c *destinationClient) SendTx(
...
+		}
+		c.logger.Debug(
+			"Signer eligible to deliver message",
+			zap.Any("address", signerAddress),
 		)
-		return nil, err
+		cases = append(cases, reflect.SelectCase{
+			Dir:  reflect.SelectSend,
+			Chan: reflect.ValueOf(c.messageChans[i]),
+			Send: reflect.ValueOf(messageData),
```
bernard-avalabs, line 299: It is definitely cleaner to have the next stage perform the signing and sending once a "slot" is available.

vms/evm/destination_client.go
```diff
@@ -239,177 +268,184 @@ func (c *destinationClient) SendTx(
...
+		})
 	}
 
-	receipt, err := c.waitForReceipt(signedTx.Hash(), signerIdx)
-	if err != nil {
+	// Select an available, eligible signer and acquire a txQueue slot
```
bernard-avalabs, line 303: Some of the comments still refer to txQueue even though it has been replaced with a queuedTxSemaphore.

May be helpful to further explain that, instead of just acquiring a slot, it sends the messagData to an available and eligible concurrentSigner thread here.

vms/evm/destination_client.go
```diff
@@ -239,177 +268,184 @@ func (c *destinationClient) SendTx(
...
-		// Only select from the signers that are in the deliverers set
-		var addresses []common.Address
-		for i, signer := range c.keys {
-			if deliverers.Contains(signer.signer.Address()) {
-				idxs[len(cases)] = i
-				cases = append(cases, c.selectCases[i])
-				addresses = append(addresses, signer.signer.Address())
+func (s *concurrentSigner) processIncomingTransactions() {
+	for {
+		// We can only only get to listen to readyChan if there is an open pending tx slot
```
bernard-avalabs, line 324:
```suggestion
		// We listen to readyChan only if there is an open pending tx slot
```

bernard-avalabs (commented):
I think this design where we have one goroutine per signer is definitely simpler and easier to reason about than what we had.

vms/evm/destination_client.go
```diff
@@ -239,177 +268,184 @@ func (c *destinationClient) SendTx(
...
-	idxs := make(map[int]int) // map of index in cases to index in c.keys
-	if deliverers.Len() != 0 {
-		// Only select from the signers that are in the deliverers set
-		var addresses []common.Address
-		for i, signer := range c.keys {
-			if deliverers.Contains(signer.signer.Address()) {
-				idxs[len(cases)] = i
-				cases = append(cases, c.selectCases[i])
-				addresses = append(addresses, signer.signer.Address())
+func (s *concurrentSigner) processIncomingTransactions() {
```
cam-schultz, line 322: Can you add a comment to this function that for any `MessageData` read from `s.messageChan`, a value is guaranteed to be written to `MessageData.resultChan`?

vms/evm/destination_client.go
```diff
@@ -239,177 +268,184 @@ func (c *destinationClient) SendTx(
...
+		})
 	}
 
-	receipt, err := c.waitForReceipt(signedTx.Hash(), signerIdx)
-	if err != nil {
+	// Select an available, eligible signer
+	reflect.Select(cases)
+
+	// Wait for the receipt or error to be returned
+	result := <-resultChan
```
cam-schultz, line 307: We should add a timeout context and corresponding select statement so that this is guaranteed to not hang indefinitely. The current implementation looks as if resultChan will always be written to within a finite amount of time, but a future change may not provide that guarantee.

cam-schultz, line 307: If the channel is closed before a value is written, then the result will have a nil error, since that is the 0 value of the channel's type. Again, this should never be the case given the current implementation, but that could change in the future.

To account for this, we should pull from the channel using `result, ok := <-resultChan` and handle the close case specifically.

vms/evm/destination_client.go
```diff
@@ -239,177 +268,184 @@ func (c *destinationClient) SendTx(
...
-	idx, _, _ := reflect.Select(cases)
-	if deliverers.Len() != 0 {
-		idx = idxs[idx]
 	}
+}
 
-	c.logger.Debug(
-		"Selected signer",
-		zap.Stringer("address", c.keys[idx].signer.Address()),
+func (s *concurrentSigner) issueTransaction(
```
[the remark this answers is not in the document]
cam-schultz: Let's also include the existing comment about synchronizing nonce access in the function comment in some capacity.

vms/evm/destination_client.go
```diff
@@ -239,177 +268,184 @@ func (c *destinationClient) SendTx(
...
+		gasLimit:      gasLimit,
+		gasFeeCap:     gasFeeCap,
+		gasTipCap:     gasTipCap,
+		callData:      callData,
+		signedMessage: signedMessage,
+		resultChan:    resultChan,
+	}
+
+	var cases []reflect.SelectCase
+	for i, signerAddress := range c.signerAddresses {
```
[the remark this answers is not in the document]
geoff-vball: I've implemented this... It does clean things up a bunch, though I'm not super happy with giving the destination client access to all of the fields of the `concurrentSigner`. Afaict, because they both need references to each other, they need to be in the same package so we can't keep anything private. I'll just document this

vms/evm/destination_client.go
```diff
@@ -56,28 +56,32 @@ type destinationClient struct {
 	txInclusionTimeout      time.Duration
 }
 
+// Type alias for the destinationClient to have access to the fields but not the methods of the concurrentSigner.
+type readonlyConcurrentSigner concurrentSigner
```
cam-schultz, line 60: [non-blocking] It may clean up the access control a bit if we transfer all `concurrentSigner` methods to `destinationClient`, and pass in the `concurrentSigner` as an argument instead. I see a few advantages:
- No need to compose `concurrentSigner` with a `*destinationClient`
- Easier to reason about the relationship between the types, since it would be pure one directional composition
- Easier to enforce the invariants on `processIncomingTransactions` and `issueTransaction` when it's internal to a single type, rather than function calls made between types.

iansuvak (approved):

PR #774, opened 2025-05-09
refactor to use filterLogs for catchup

iansuvak:
## Why this should be merged
Reduce the number of RPC calls made during catch-up to just a single log fetch per 200 blocks range instead of 2 per block if the block contains warp messages. 

## How this works
- Change listener to listen on a new channel that has pre-processed `WarpBlockInfo` instead of raw headers
- Pre-process both result of log filter queries during catchup and headers into the warpBlockInfo. 

## How this was tested
- Existing unit tests extended
- existing e2e passing

### TODO:

- Add explicit tests for types package including `NewWarpMessageInfo` and `LogsToBlocks`
- Extend the subscriber catch-up test to check for cases where icm messages exist in catchup blocks

## How is this documented
-updated comments


vms/evm/subscriber.go
```diff
@@ -182,11 +238,27 @@ func (s *subscriber) subscribe(retryTimeout time.Duration) error {
...
-func (s *subscriber) Headers() <-chan *types.Header {
-	return s.headers
+// blocksInfoFromHeaders listens to the header channel and converts the headers to [relayerTypes.WarpBlockInfo]
+// and writes them to the blocks channel consumed by the listener
+func (s *subscriber) blocksInfoFromHeaders() {
+	for header := range s.headers {
+		block, err := relayerTypes.NewWarpBlockInfo(s.logger, header, s.rpcClient)
+		if err != nil {
+			s.logger.Error("Failed to create Warp block info", zap.Error(err))
+			// TODO: handle error -- send to a different channel (?)
```
cam-schultz, line 253: Any errors emitted by the catch up mechanism should result in `false` being written to the `done` channel

vms/evm/subscriber.go
```diff
@@ -112,7 +120,7 @@ func (s *subscriber) processBlockRange(
 	fromBlock, toBlock *big.Int,
 ) error {
 	for i := fromBlock.Int64(); i <= toBlock.Int64(); i++ {
-		header, err := s.getHeaderByNumberRetryable(big.NewInt(i))
+		logs, err := s.getFilterLogsByBlockRangeRetryable(fromBlock, toBlock)
```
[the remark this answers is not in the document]
cam-schultz: In fact, I believe we can remove the outer loop altogether.

relayer/listener.go
```diff
@@ -192,14 +194,19 @@ func (lstnr *Listener) processLogs(ctx context.Context) error {
...
-				lstnr.ethClient,
 				errChan,
 			)
 		case err := <-lstnr.Subscriber.Err():
+			// If the error is due to failed to process logs rather than a subscription
+			if errors.Is(err, relayerTypes.ErrFailedToProcessLogs) {
+				lstnr.healthStatus.Store(false)
+				lstnr.logger.Error("Error processing logs. Relayer goroutine exiting")
+				return fmt.Errorf("error processing logs: %w", err)
+			}
```
cam-schultz, line 209: Any errors emitted by the catch up mechanism should result in `false `being written to the `done` channel

vms/evm/subscriber.go
```diff
@@ -121,11 +129,31 @@ func (s *subscriber) processBlockRange(
 			)
 			return err
 		}
-		s.headers <- header
+		blocks, err := relayerTypes.LogsToBlocks(logs)
```
cam-schultz, line 132:
```suggestion
		blocksWithICMMessages, err := relayerTypes.LogsToBlocks(logs)
```
To clarify that this is not expected to be a contiguous list

vms/evm/subscriber.go
```diff
@@ -121,11 +129,31 @@ func (s *subscriber) processBlockRange(
...
+		if err != nil {
+			s.logger.Error("Failed to convert logs to blocks", zap.Error(err))
+			return err
+		}
+		currentBlock := fromBlock.Uint64()
```
cam-schultz, line 137:
```suggestion
		// Write values for every block number between fromBlock and toBlock to the icmBlocks channel,
		// filling in blocks that don't have ICM messages with empty blocks
		currentBlock := fromBlock.Uint64()
```

vms/evm/subscriber.go
```diff
@@ -121,11 +129,31 @@ func (s *subscriber) processBlockRange(
...
+		for _, block := range blocks {
+			s.sendEmptyBlocks(currentBlock, block.BlockNumber-1)
+			s.icmBlocks <- block
+			currentBlock = block.BlockNumber + 1
+		}
+		// Send empty blocks for the blocks after the last block in the range
+		s.sendEmptyBlocks(currentBlock, toBlock.Uint64())
```
cam-schultz, line 144: I'd be in favor of moving this to a helper function that takes a discontinuous list of blocks, a start height, and an end height, and returns a contiguous list of blocks with empty blocks filled in as necessary. Then we send all values of that list to the channel.

That should make it easier to tell what is being sent when.

vms/evm/subscriber.go
```diff
@@ -30,8 +33,11 @@ type subscriber struct {
 	rpcClient    ethclient.Client
 	blockchainID ids.ID
 	headers      chan *types.Header
+	icmBlocks    chan *relayerTypes.WarpBlockInfo
 	sub          interfaces.Subscription
 
+	errChan chan error
```
cam-schultz, line 39: This is unused

types/types.go
```diff
@@ -111,3 +113,32 @@ func UnpackWarpMessage(unsignedMsgBytes []byte) (*avalancheWarp.UnsignedMessage,
...
+		if a.BlockNumber < b.BlockNumber {
+			return -1
+		}
+		if a.BlockNumber > b.BlockNumber {
+			return 1
+		}
+		return 0
+	})
+
+	blocks := make([]*WarpBlockInfo, 0)
```
cam-schultz, line 128: nit: slight preference toward `var blocks []*WarpBlockInfo` to avoid the common mix up (at least for me) between size and cap arguments in `make`

types/types.go
```diff
@@ -111,3 +113,32 @@ func UnpackWarpMessage(unsignedMsgBytes []byte) (*avalancheWarp.UnsignedMessage,
...
+	for _, log := range logs {
+		// If the slice is empty or the last block is not the current block, add a new block
```
cam-schultz, line 130: This comment may be more explanatory if rephrased to "If this log's block hasn't been seen yet, append a new entry to the slice"

vms/evm/subscriber.go
iansuvak: This is an error from the live subscription instead of the catchup. Before they were self-contained in the processBlock step since that's where the `NewWarpBlockInfo` RPC call was being made. Now we need an alternate way to communicate to the listener hence the TODO comment. 

The easiest solution is to expose another separate channel here but doesn't seem clean to have both `Err()` and this new channel which will only really error out on RPC issues or if the logs are malformed which is tons less likely.

types/types.go
iansuvak: Yeah my instinct is to err on the side of make so I don't need to distinguish between maps and slices but don't actually prefer one over the other -- made the change

types/types.go
```diff
@@ -111,3 +113,32 @@ func UnpackWarpMessage(unsignedMsgBytes []byte) (*avalancheWarp.UnsignedMessage,
...
+func LogsToBlocks(logs []types.Log) ([]*WarpBlockInfo, error) {
+	slices.SortFunc(logs, func(a, b types.Log) int {
+		if a.BlockNumber < b.BlockNumber {
+			return -1
+		}
+		if a.BlockNumber > b.BlockNumber {
+			return 1
+		}
+		return 0
+	})
```
geoff-vball, line 126: `sort.SliceStable` which uses a boolean sortfunc is way more concise

vms/evm/subscriber.go
iansuvak: I've made the change but not sure that it's the cleanest. Happy to keep, revert or accept additional suggestions. 

I think that extracting the for loop that makes the empty blocks was a big readability win but I feel less strongly about this one. The nice thing about the for loop actually sending blocks on the channel was that we had less assignments and slice manipulation.

types/types.go
```diff
@@ -111,3 +113,32 @@ func UnpackWarpMessage(unsignedMsgBytes []byte) (*avalancheWarp.UnsignedMessage,
...
+
+	var blocks []*WarpBlockInfo
+	for _, log := range logs {
+		// If this log's block hasn't been seen yet, append a new entry to the slice
+		if len(blocks) == 0 || blocks[len(blocks)-1].BlockNumber != log.BlockNumber {
+			blocks = append(blocks, &WarpBlockInfo{
+				BlockNumber: log.BlockNumber,
+				Messages:    make([]*WarpMessageInfo, 0),
+			})
+		}
```
geoff-vball, line 136: This step seems really awkward. Could we do something like, create a `map[int][]*WarpMessageInfo`. Then for each iteration we just append the `warpMessageInfo` entry at that block number. Then, when we have the full map, we can iterate over the entries creating a `[]*WarpBlockInfo`, then sort.

vms/evm/subscriber.go
iansuvak: I'm using it now but not thrilled about listening on two different error channels but any "merging" strategies seem worse.

vms/evm/subscriber.go
```diff
@@ -182,17 +206,66 @@ func (s *subscriber) subscribe(retryTimeout time.Duration) error {
...
 func (s *subscriber) Cancel() {
 	// Nothing to do here, the ethclient manages both the log and err channels
 }
+
+// Takes a block range and a slice of blocks containing ICM messages and fills in the missing blocks with empty blocks
+func fillBlockRange(fromBlock, toBlock uint64, icmBlocks []*relayerTypes.WarpBlockInfo) ([]*relayerTypes.WarpBlockInfo, error) {
+	var filledBlocks []*relayerTypes.WarpBlockInfo
+	currentBlock := fromBlock
+	for _, block := range icmBlocks {
+		filledBlocks = append(filledBlocks, makeEmptyBlocks(currentBlock, block.BlockNumber-1)...)
```
geoff-vball, line 252: What the behaviour here if `currentBlock` = `block.BlockNumber-1`. Or do we append `nil`? Or is it a no-op

vms/evm/subscriber.go
```diff
@@ -182,17 +206,66 @@ func (s *subscriber) subscribe(retryTimeout time.Duration) error {
...
+
 func (s *subscriber) Cancel() {
 	// Nothing to do here, the ethclient manages both the log and err channels
 }
+
+// Takes a block range and a slice of blocks containing ICM messages and fills in the missing blocks with empty blocks
+func fillBlockRange(fromBlock, toBlock uint64, icmBlocks []*relayerTypes.WarpBlockInfo) ([]*relayerTypes.WarpBlockInfo, error) {
+	var filledBlocks []*relayerTypes.WarpBlockInfo
+	currentBlock := fromBlock
+	for _, block := range icmBlocks {
```
geoff-vball, line 251: It would be slightly less efficient (actually, maybe not because of all the appending we do), but much more straightforward if we just created an array of empty blocks for the whole range, and then overwrote the entries with the `icmBlocks`

types/types.go
iansuvak: Yeah that's the way I had it originally but swapped because of a IDE hint saying that slices.SortFunc is faster/more readable.. Not sure that I agree on the more readable.

vms/evm/subscriber.go
```diff
@@ -182,17 +206,66 @@ func (s *subscriber) subscribe(retryTimeout time.Duration) error {
 		return err
 	}
 	s.sub = sub
+
+	go s.blocksInfoFromHeaders()
```
geoff-vball, line 210: Making a physical note to come back here

types/types.go
geoff-vball: lol I feel like you're being gaslit 😛

types/types.go
iansuvak: I just opted for a full simplification where this function returns the map itself and then we no longer need to do sorting or anything. I think it's a lot cleaner and requires fewer `append` allocations except for re-appending to the `WarpMessageInfo.Messages` slice

cam-schultz (approved):

cam-schultz (commented):
Let's add the latest block height to the log [here](https://github.com/ava-labs/icm-services/blob/2acac16a435f56a37831c1dd5779f77173974f43/vms/evm/subscriber.go#L71), and an info log within the loop [here](https://github.com/ava-labs/icm-services/blob/2acac16a435f56a37831c1dd5779f77173974f43/vms/evm/subscriber.go#L98) to track catch-up progress

geoff-vball (commented):
Just one minor concern, otherwise lgtm

vms/evm/subscriber.go
```diff
@@ -182,17 +209,40 @@ func (s *subscriber) subscribe(retryTimeout time.Duration) error {
...
```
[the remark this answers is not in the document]
iansuvak: Thanks for the catch! Moved it to `NewSubscriber` so it only gets called once and is guaranteed to get called when the subscriber is started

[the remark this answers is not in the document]
cam-schultz: Great call out. We should definitely address this, but just want to point out this isn't a safety concern, since only one goroutine can read a given value from `s.headers`, and `s.icmBlocks` makes no guarantee on block order.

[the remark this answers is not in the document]
iansuvak: Yeah the test does assume that they will be in order but it passes since we don't resubscribe during it

geoff-vball (approved):

PR #775, opened 2025-05-13
Override default http client

iansuvak:
## Why this should be merged

Overwrites global http.DefaultClient with some hopefully sane defaults until client structs in avago accept clients directly. 

## How is this documented


geoff-vball (approved):

relayer/main/main.go
```diff
@@ -87,6 +87,23 @@ func main() {
...
+	// Modify the default http.DefaultClient globally
+	// TODO: Remove this temporary fix once the RPC clients used by the relayer
+	// start accepting custom underlying http clients.
+	{
+		// Set the timeout conservatively to catch any potential cases where the context is not used
+		// and the request hangs indefinitely.
+		http.DefaultClient.Timeout = 2 * utils.DefaultRPCTimeout
+		maxConns := 10_000
+		http.DefaultClient.Transport = &http.Transport{
+			MaxConnsPerHost:     maxConns,
```
cam-schultz, line 100: If we want to relax this client as much as possible, should we set this to 0 (no limit)?

cam-schultz (approved):

PR #779, opened 2025-05-23
replace panics and fix version flag

### LINKED ISSUE #778: Replace panics with fatals where appropriate
**Context and scope**
We currently panic at app start ups even though it's not always appropriate and can cause confusion in logs. We should re-evaluate instances where we do and replace them with fatals. 

**Open questions**
Things to consider are here are that panics will execute defers but fatals will call `os.Exit()` which won't and panics will output stack traces unlike fatals.

iansuvak:
## Why this should be merged
Closes #778 
## How this works
-- replaces panics during config parsing with calls to `log.Fatalf` which is equivalent to `printLn` and followed by `os.Exit(1)`. 
- fixes calls to --version flag by changing the `const` to a `var` in both `main.go` files and modifies the build scripts to match the correct prefix to override it. 
## How this was tested

Tested manually that output is human readable and works correctly and that the version flag works as intended. 

## How is this documented

### DIFF
--- a/relayer/main/main.go
+++ b/relayer/main/main.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log"
 	"net/http"
 	"os"
 	"runtime"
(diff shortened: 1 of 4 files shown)


bernard-avalabs (commented):
Generally looks good to me. My only concern is the elimination of stack traces in a log.fatal compared to a panic.

iansuvak:
> Generally looks good to me. My only concern is the elimination of stack traces in a log.fatal compared to a panic.

Yeah I left the runtime panics where we actually want to crash, leave a stack trace and execute defers. These are just startup config parsing issues so stack traces shouldn't be meaningful

geoff-vball (approved):

cam-schultz (approved):
LGTM

PR #795, opened 2025-06-09
Increment request IDs on retries

iansuvak:
## Why this should be merged
Correctly handles retries on the retries of the signature requests by incrementing the request ID for each individual app request made to the validators. 

Also logs the requestID in a wrapped logger so that any request related logs have it can easily be filtered on when debugging. 
## How this was tested
CI
## How is this documented


geoff-vball (approved):

cam-schultz (approved):
LGTM.

Thinking aloud about what happens currently without this change - on retry, we'd resend out app requests to the validator set, which cannot change between attempts. If we receive an app response from a node that was responding to the prior attempt, then the node's response could be double-counted in the current attempt if it responds to that app request as well. This is definitely a bug, and unique IDs per app request should address it.

bernard-avalabs (approved):

cam-schultz (approved):

PR #799, opened 2025-06-11
Increase log fetch timeout to handle EVM header/RPC race condition

ylg-avalabs:
## Why this should be merged
This PR addresses a race condition where the relayer may receive a block header via websocket before the block’s logs are available via RPC (https://github.com/ava-labs/icm-services/issues/785). This is a known behavior in EVM-based nodes due to asynchronous log and index processing after a block becomes canonical. Without this change, the relayer may fail to fetch logs for new blocks, resulting in missed or delayed Warp message processing.

## How this works
The timeout for fetching logs in `NewWarpBlockInfo` is increased by a factor of two `utils.DefaultRPCTimeout * 2` when calling `utils.WithRetriesTimeout`. This gives the RPC backend more time to index the block and make logs available, reducing the likelihood of encountering the race condition. The code comments clearly explain the rationale for this increased timeout.

## How is this documented


types/types.go
```diff
@@ -59,7 +59,13 @@ func NewWarpBlockInfo(logger logging.Logger, header *types.Header, ethClient eth
 			})
 			return err
 		}
-		err = utils.WithRetriesTimeout(logger, operation, utils.DefaultRPCTimeout, "get warp logs from block")
+
+		// We increase the timeout here to reduce the chance of hitting a race condition
+		// where the block header is received via websocket subscription before the block's
+		// logs are available via RPC. This is a known behavior in EVM nodes due to
+		// asynchronous log/index processing after a block becomes canonical.
+		timeout := utils.DefaultRPCTimeout * 2
```
geoff-vball, line 67: I would be okay setting this even higher. If we got the header, the logs should be available eventually, and we have no recovery mechanism either, so I think waiting somewhere on the order of ~30s would still be appropriate

iansuvak (approved):

PR #800, opened 2025-06-11
Improve Relayer Startup Logging and Diagnostics for Insufficient Stake Connectivity

ylg-avalabs:
## Why this should be merged
This PR significantly improves the diagnostics and troubleshooting capabilities of the relayer during startup, especially in scenarios where it fails to connect to sufficient stake. Enhanced logging provides better visibility into validator connectivity, stake distribution, and configuration, making it easier to identify and resolve startup issues. (https://github.com/ava-labs/icm-services/issues/730)

## How this works
- Adds detailed debug logging of each connected validator's nodeID and weight when the relayer fails to connect to sufficient stake.
- Logs additional context at key steps, including:
  - When fetching warp configs for each destination.
  - When attempting to retrieve connected validators.
- The warning log for insufficient stake now includes the number of connected peers (global or subnet-scoped, as available).

## How is this documented


iansuvak (commented):
Can we modify the success  `Info` message as well above on line 109 for consistency?

And make the changes to runtime failures to connect in the [SignatureAggregator.connectToQuorumValidators]?

relayer/network_utils.go
```diff
@@ -86,30 +95,47 @@ func checkSufficientConnectedStake(
...
+		if !utils.CheckStakeWeightExceedsThreshold(
 			big.NewInt(0).SetUint64(connectedValidators.ConnectedWeight),
 			connectedValidators.ValidatorSet.TotalWeight,
 			maxQuorumNumerator,
 		) {
-			logger.Info(
-				"Connected to sufficient stake",
+			// Log details of each connected validator (nodeID and weight).
+			// This is useful for troubleshooting startup issues when the relayer fails to connect to sufficient stake.
+			if logger.Enabled(logging.Debug) {
```
iansuvak, line 106: I would downgrade this one to be `verbose` level instead of `debug` since it's at runtime and on a per-request basis. But i'm also fine leaving it as is since we have had even louder debugs in the past.

bernard-avalabs (approved):

relayer/network_utils.go
```diff
@@ -86,30 +95,47 @@ func checkSufficientConnectedStake(
...
+					for _, nodeID := range vdr.NodeIDs {
+						logger.Verbo(
+							"Connected validator details",
+							zap.Stringer("subnetID", subnetID),
+							zap.String("nodeID", nodeID.String()),
+							zap.Uint64("weight", vdr.Weight),
+						)
+					}
+				}
+			}
```
cam-schultz, line 117: I think this would be much more useful at debug log level. At present, we only use verbo logs to trace high frequency, low level operations, such as DB writes and p2p requests. These logs however, target relayer connectivity only on startup.

Another way to frame it is that it's easy to imagine scenarios where we'd want to up our logging to debug on a live deployment, but bumping to verbo jwould likely generate a prohibitively large number of logs.

relayer/network_utils.go
```diff
@@ -86,30 +95,47 @@ func checkSufficientConnectedStake(
...
+		if !utils.CheckStakeWeightExceedsThreshold(
 			big.NewInt(0).SetUint64(connectedValidators.ConnectedWeight),
 			connectedValidators.ValidatorSet.TotalWeight,
 			maxQuorumNumerator,
 		) {
-			logger.Info(
-				"Connected to sufficient stake",
+			// Log details of each connected validator (nodeID and weight).
+			// This is useful for troubleshooting startup issues when the relayer fails to connect to sufficient stake.
+			if logger.Enabled(logging.Verbo) {
```
cam-schultz, line 106: Can we instead log the validator details unconditionally, rather than only in the case that we're not connected to sufficient stake? If we have a relayer that inconsistently connects to peers, it would be helpful to see the connected list across many runs, regardless of the level of connectivity.

ylg-avalabs: [Moved the log](https://github.com/ava-labs/icm-services/pull/800/commits/3f226a7ea92671f285fb1f26f46985191ac743ad#diff-37ffd39f2866baaf19af2ce5bba9578258a42bc2572e119919553def393cfb50R99-R110) out of the if statement so that it logs the validator info unconditionally.

geoff-vball (approved):

PR #817, opened 2025-07-09
Update go

geoff-vball:
## How is this documented


iansuvak:
Actually it looks like icm-contracts was bumped down here by accident as well.

michaelkaplan13 (approved):

PR #819, opened 2025-07-14
Reuse network in e2e test

ylg-avalabs:
## Why this should be merged
- Allow to reuse network in e2e test
- It saves e2e test time

## How this works
- Create `network-dir` flag to passing in the network directory storing network info
- Save the teleporter registry addresses and validator addresses when creating a new network in the tests
- When the `network-dir` flag is set, it loads the network from the directory and set the saved teleporter registry addresses and validator addresses

## How this was tested
- e2e tests

## How is this documented


geoff-vball:
I tried running this but couldn't get it to work. Where is the default location for `network-dir` that the network is saved to? I was getting an error saying "config.json" does not exist.

I would have expected to be able to set `network-dir` to a specific location and then:
- If a network exists there, it reuses it. 
- If a network doesn't exist there, it creates a new one from scratch and saves it there.

ylg-avalabs:
> I tried running this but couldn't get it to work. Where is the default location for `network-dir` that the network is saved to? I was getting an error saying "config.json" does not exist.
> 
> I would have expected to be able to set `network-dir` to a specific location and then:
> 
> * If a network exists there, it reuses it.
> * If a network doesn't exist there, it creates a new one from scratch and saves it there

When we run the test in normal case (`./scripts/e2e_test.sh`), we will see the following logs. In the log, we see the directory where the network is persisted. Then, in the next run, we can reuse this network: 
```
/scripts/e2e_test.sh --network-dir /Users/linguan.yang/.tmpnet/networks/20250724-144425.01028-icm-off-chain-services-eZe-test
```
Screenshot 2025-07-24 at 2 45 04 PM

scripts/e2e_test.sh
```diff
@@ -19,6 +30,7 @@ if [ "$HELP" = true ]; then
     echo "Run E2E tests for ICM Services."
     echo ""
     echo "Options:"
+    echo "  --network-dir                     Path to the network directory to reuse. If not provided, a new network will be created."å
```
geoff-vball, line 33:
```suggestion
    echo "  --network-dir                     Path to the network directory to reuse. If not provided, a new network will be created."
```

scripts/e2e_test.sh
```diff
@@ -58,6 +82,9 @@ go build -v -o tests/cmd/decider/decider ./tests/cmd/decider/
 # Run the tests
 echo "Running e2e tests $RUN_E2E"
 RUN_E2E=true LOG_LEVEL=${LOG_LEVEL} SIG_AGG_PATH=${SIG_AGG_PATH:-"$BASEDIR/icm-services/signature-aggregator"} ./tests/tests.test \
+  --root-network-dir=${root_dir} \
```
geoff-vball, line 85: What happens if we pass an empty string into this flag?

scripts/e2e_test.sh
```diff
@@ -34,6 +46,18 @@ source "$BASE_PATH"/scripts/versions.sh
 
 BASEDIR=${BASEDIR:-"$HOME/.teleporter-deps"}
 
+if [ -n "$reuse_network_dir" ]; then
+    if [ -d "$reuse_network_dir" ] && [ "$(ls -A "$reuse_network_dir")" ]; then
```
geoff-vball, line 50:
```suggestion
    if [ -d "$reuse_network_dir" ] && [ "$(ls -a "$reuse_network_dir")" ]; then
```

`ls -A` will always return something because it includes `.` and `..`
`ls -a` will omit these

tests/basic_relay.go
```diff
@@ -167,8 +167,8 @@ func BasicRelay(network *network.LocalNetwork, teleporter utils.TeleporterTestIn
 
 	// We should not receive a new block on subnet B, since the relayer should have
 	// seen the Teleporter message was already delivered.
-	log.Info("Waiting for 10s to ensure no new block confirmations on destination chain")
-	Consistently(newHeadsB, 10*time.Second, 500*time.Millisecond).ShouldNot(Receive())
+	// log.Info("Waiting for 10s to ensure no new block confirmations on destination chain")
+	// Consistently(newHeadsB, 10*time.Second, 500*time.Millisecond).ShouldNot(Receive())
```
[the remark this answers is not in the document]
geoff-vball: Does this still fail?

scripts/e2e_test.sh
ylg-avalabs: If the root-network-dir is empty string, it will use the [default root network dir](https://github.com/ava-labs/avalanchego/blob/master/tests/fixture/tmpnet/network.go#L298-L304).

tests/e2e_test.go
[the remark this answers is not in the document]
iansuvak: To test the L1 case. In general networks first get created as subnets and the concept of subnet still remains, i.e. they are still referred to by their `subnetID` even if they are L1s

PR #824, opened 2025-07-14
N-04 Code Simplification Suggestions

ylg-avalabs:
## Why this should be merged
Throughout the codebase, multiple opportunities for code simplification were identified. These recommendations span various components and functionalities, addressing both specific code segments and broader architectural patterns:
  - Within the [initializeTrackedSubnets](https://github.com/ava-labs/icm-services/blob/11c6cafa6d5a8572a2f48364e850491db5c86b80/relayer/config/config.go#L248) function, the warp configuration per destination chain can be accessed through `destinationBlockchain.warpConfig` instead of through the 1GetWarpConfig` function.
  - The `req.Message` could be sanitized with [SanitizeHexString](https://github.com/ava-labs/icm-services/blob/11c6cafa6d5a8572a2f48364e850491db5c86b80/utils/utils.go#L129), just like `req.Justification`.
  - The [accumulatedSignatureWeight](https://github.com/ava-labs/icm-services/blob/11c6cafa6d5a8572a2f48364e850491db5c86b80/signature-aggregator/aggregator/aggregator.go#L501) value is emitted twice in the warning.
  - In the relayer's `main.go`, the decider connection could be closed on `defer` instead of [setting a finalizer](https://github.com/ava-labs/icm-services/blob/11c6cafa6d5a8572a2f48364e850491db5c86b80/relayer/main/main.go#L596).
  - Every time `writeToDatabase` is invoked, the [database is read](https://github.com/ava-labs/icm-services/blob/11c6cafa6d5a8572a2f48364e850491db5c86b80/relayer/checkpoint/checkpoint.go#L67-L78) to check whether the committed height exceeds the stored height. This repeated check can be avoided by moving this logic into the `NewCheckpointManager` function by setting `committedHeight` to the bigger value of `startingHeight` and `storedHeight`.

## How is this documented

### DIFF
--- a/relayer/checkpoint/checkpoint.go
+++ b/relayer/checkpoint/checkpoint.go
@@ -5,6 +5,7 @@ package checkpoint
 
 import (
 	"container/heap"
+	"math"
 	"strconv"
 	"sync"
 
@@ -26,6 +27,8 @@ type CheckpointManager struct {
 	committedHeight uint64
 	lock            *sync.RWMutex
 	pendingCommits  *utils.UInt64Heap
+	// Update the dirty flag when committedHeight is updated
+	dirty bool
 }
 
 func NewCheckpointManager(
@@ -43,7 +46,6 @@ func NewCheckpointManager(
 		zap.Uint64("startingHeight", startingHeight),
 	)
 
-	committedHeight := startingHeight
 	storedHeight, err := database.GetLatestProcessedBlockHeight(db, relayerID)
 	if err != nil && !database.IsKeyNotFoundError(err) {
 		logger.Error(
@@ -54,16 +56,14 @@ func NewCheckpointManager(
 		return nil
 	}
 
-	if storedHeight > committedHeight {
-		committedHeight = storedHeight
-	}
+	committedHeight := math.Max(float64(storedHeight), float64(startingHeight))
 
 	return &CheckpointManager{
 		logger:          logger,
 		database:        db,
 		writeSignal:     writeSignal,
 		relayerID:       relayerID,
-		committedHeight: committedHeight,
+		committedHeight: uint64(committedHeight),
 		lock:            &sync.RWMutex{},
 		pendingCommits:  h,
 	}
@@ -77,7 +77,8 @@ func (cm *CheckpointManager) writeToDatabase() {
 	cm.lock.RLock()
 	defer cm.lock.RUnlock()
 	// Defensively ensure we're not writing the default value
-	if cm.committedHeight == 0 {
+	// If the committedHeight is not change, we skip the write
+	if cm.committedHeight == 0 || !cm.dirty {
 		return
 	}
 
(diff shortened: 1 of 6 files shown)


geoff-vball:
Don't merge this yet... I want to double-check the logic on the database operations

relayer/checkpoint/checkpoint.go
```diff
@@ -42,12 +42,28 @@ func NewCheckpointManager(
...
+			"Failed to get latest processed block height",
+			zap.Error(err),
+			zap.String("relayerID", relayerID.ID.String()),
+		)
+		return nil
+	}
+
+	if storedHeight > committedHeight {
+		committedHeight = storedHeight
+	}
```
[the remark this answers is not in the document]
geoff-vball: I'm pretty sure there aren't any cases where `storedHeight` may be higher than the `startingHeight`, so I think in terms of correctness this should all be fine... 

Though I'm wondering if we should set some kind of `dirty` flag when `committedHeight` is updated. `writeToDatabase()` is triggered on a timer, and if we don't have a new value right now we'll still write anyway.

iansuvak:
I do think that preventing unnecessary writes like Geoff mentioned would be good here

relayer/checkpoint/checkpoint.go
```diff
@@ -43,30 +47,27 @@ func NewCheckpointManager(
...
 			zap.String("relayerID", relayerID.ID.String()),
 		)
-		return nil
+		return nil, fmt.Errorf("failed to get the latest processed block height: %v", err)
 	}
 
-	if storedHeight > committedHeight {
-		committedHeight = storedHeight
-	}
+	committedHeight := math.Max(float64(storedHeight), float64(startingHeight))
```
geoff-vball, line 60: Whoops, sorry I forgot `math.Max()` only took floats... I don't think we want to cast here. We should just write a helper function for uint64s (or a generic function if you're feeling ambitious).

relayer/checkpoint/checkpoint.go
```diff
@@ -77,7 +78,8 @@ func (cm *CheckpointManager) writeToDatabase() {
 	cm.lock.RLock()
 	defer cm.lock.RUnlock()
 	// Defensively ensure we're not writing the default value
-	if cm.committedHeight == 0 {
+	// If the committedHeight is not change, we skip the write
```
geoff-vball, line 81:
```suggestion
	// If committedHeight is not changed, we can skip the write
```

relayer/checkpoint/checkpoint.go
```diff
@@ -77,7 +78,8 @@ func (cm *CheckpointManager) writeToDatabase() {
 	cm.lock.RLock()
 	defer cm.lock.RUnlock()
 	// Defensively ensure we're not writing the default value
-	if cm.committedHeight == 0 {
```
geoff-vball, line 64: I think we should change the lock acquisition above to just regular lock/unlock since we're writing to the `dirty` variable now.

relayer/checkpoint/checkpoint.go
ylg-avalabs: Oh yes, we can directly use built-in function `max()` starting Go 1.21.

geoff-vball (approved):

PR #827, opened 2025-07-15
N-01 Unbuffered Error Channel Leads to Potential Goroutine Leaks

ylg-avalabs:
## Why this should be merged
In the [processLogs](https://github.com/ava-labs/icm-services/blob/11c6cafa6d5a8572a2f48364e850491db5c86b80/relayer/listener.go#L161) function, downstream errors from [block processing](https://github.com/ava-labs/icm-services/blob/11c6cafa6d5a8572a2f48364e850491db5c86b80/relayer/application_relayer.go#L184) are sent to an unbuffered channel, errChan. This channel is consumed by the main select loop, which processes one event at a time. While this works for sequential errors, it creates a race condition in scenarios with concurrent failures.

If multiple `ProcessBlock` goroutines encounter an error at the same time, they will all attempt to send over the unbuffered channel. While the first one will succeed, the others will block indefinitely, waiting for a receiver. This is particularly problematic because the `processLogs` loop can terminate for reasons unrelated to the errChan, such as a
WebSocket subscription error or context cancellation. If the loop exits while other goroutines are still blocked waiting to send their errors, those goroutines will be orphaned and leak forever. Although, in the current state of the application, the program would exit after the first error in the listener.

Consider replacing the unbuffered channel with a buffered channel to report multiple errors without blocking. This will allow the system to accommodate the reporting of multiple errors concurrently. Making the code more future-proof, even during a cascade of failures or an abrupt shutdown of the listener loop.

## How is this documented

### DIFF
--- a/relayer/listener.go
+++ b/relayer/listener.go
@@ -38,6 +38,7 @@ type Listener struct {
 	healthStatus       *atomic.Bool
 	ethClient          ethclient.Client
 	messageCoordinator *MessageCoordinator
+	maxConcurrentMsg   uint64
 }
 
 // RunListener creates a Listener instance and the ApplicationRelayers for a subnet.
@@ -50,6 +51,7 @@ func RunListener(
 	relayerHealth *atomic.Bool,
 	startingHeight uint64,
 	messageCoordinator *MessageCoordinator,
+	maxConcurrentMsg uint64,
 ) error {
 	// Create the Listener
 	listener, err := newListener(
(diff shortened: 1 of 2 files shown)


geoff-vball (commented):
I think making this a config variable would burden users with having to understand something they shouldn't really care about. I think we should just set some const somewhere around maybe 100?

michaelkaplan13:
> I think making this a config variable would burden users with having to understand something they shouldn't really care about. I think we should just set some const somewhere around maybe 100?

This is actually already a [configuration option](https://github.com/ava-labs/icm-services/blob/main/relayer/config/config.go#L76), with a default of 250. I think we're just re-using the existing configuration value here to set the maximum number of items in the channel.

michaelkaplan13 (approved):

iansuvak (approved):
What @michaelkaplan13 said -- this is re-using an existing config option

PR #828, opened 2025-07-15
L-02 Ensure Complete Network Shutdown in appRequestNetwork

ylg-avalabs:
## Why this should be merged
The initialization process of the P2P networking stack within the [NewNetwork](https://github.com/ava-labs/icm-services/blob/11c6cafa6d5a8572a2f48364e850491db5c86b80/peers/app_request_network.go#L111) function of app_request_network.go involves the creation of two critical background goroutines:
1. A goroutine to run the main network event loop ( `testNetwork.Dispatch()`).
2. A goroutine to periodically refresh the validator sets for all tracked subnets([startUpdateValidators](https://github.com/ava-labs/icm-services/blob/11c6cafa6d5a8572a2f48364e850491db5c86b80/peers/app_request_network.go#L331)).

The `appRequestNetwork` struct provides a `Shutdown` method, which is intended to gracefully terminate these background processes and release their resources. However, it does not provide any mechanism to stop the `startUpdateValidators` goroutine. While the program only shuts down on exit, this leak can become more severe in future refactors.

Consider passing a context into `startUpdateValidators` that can be cancelled from the `Shutdown` function.

## How is this documented


peers/app_request_network.go
```diff
@@ -325,21 +333,28 @@ func (n *appRequestNetwork) TrackSubnet(subnetID ids.ID) {
...
+				n.trackedSubnetsLock.RUnlock()
+
+				n.logger.Debug(
+					"Fetching validators for subnets",
+					zap.Any("subnetIDs", append([]ids.ID{constants.PrimaryNetworkID}, subnets...)),
+				)
+
+				n.updateValidatorSet(context.Background(), constants.PrimaryNetworkID)
+				for _, subnet := range subnets {
+					n.updateValidatorSet(context.Background(), subnet)
```
geoff-vball, line 353: I think we should pass in the `appRequestNetwork`'s `ctx` to these functions

peers/app_request_network.go
```diff
@@ -273,6 +277,7 @@ func NewNetwork(
 	}
 	vdrsCache := cache.NewTTLCache[ids.ID, avalancheWarp.CanonicalValidatorSet](canonicalValidatorSetCacheTTL)
 
+	ctx, cancel := context.WithCancel(context.Background())
```
geoff-vball, line 280: We might also want to consider having `NewNetwork` take a `ctx` parameter that we can use here instead of `context.Background()`

signature-aggregator/main/main.go
```diff
@@ -105,7 +108,10 @@ func main() {
 		logger.Fatal("Failed to create app request network", zap.Error(err))
 		os.Exit(1)
 	}
-	defer network.Shutdown()
+	defer func() {
```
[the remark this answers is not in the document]
ylg-avalabs: I think the [order of the defer functions are LIFO](https://go.dev/blog/defer-panic-and-recover). If we call defer cancel() right after we create the context, the network.shutdown() will be called first and then cancel(). In this case, network.shutdown() will be blocked, since it wait for cancel signal to stop the goroutine.

relayer/main/main.go
```diff
@@ -176,7 +178,10 @@ func main() {
 		logger.Fatal("Failed to create app request network", zap.Error(err))
 		os.Exit(1)
 	}
-	defer network.Shutdown()
+	defer func() {
+		cancel()
+		network.Shutdown()
```
geoff-vball, line 183: If we want to wait for individual portions of the code to shutdown, I think we should propogate a `waitgroup` to them. That way we can wait on the waitgroup for all the services to exit, and we don't need odd things like this

signature-aggregator/main/main.go
```diff
@@ -151,11 +155,27 @@ func main() {
...
-		logger.Fatal("Server error", zap.Error(err))
+	errGroup.Go(func() error {
+		httpServer := &http.Server{
+			Addr: fmt.Sprintf(":%d", cfg.APIPort),
+		}
+		// Handle Graceful shutshown
+		go func() {
+			<-ctx.Done()
+			if err := httpServer.Shutdown(ctx); err != nil {
+				logger.Error("Failed to shutdown healthcheck server", zap.Error(err))
```
iansuvak, line 166: This isn't just the healthcheck server, right? It's the whole server including the main `aggregate-signatures` endpoint. 

Otherwise these changes look better

iansuvak (approved):
LGTM but want @geoff-vball to take another look

signature-aggregator/main/main.go
```diff
@@ -151,11 +155,27 @@ func main() {
...
-	err = http.ListenAndServe(fmt.Sprintf(":%d", cfg.APIPort), nil)
-	if errors.Is(err, http.ErrServerClosed) {
-		logger.Info("Server closed")
-	} else if err != nil {
-		logger.Fatal("Server error", zap.Error(err))
+	errGroup.Go(func() error {
+		httpServer := &http.Server{
+			Addr: fmt.Sprintf(":%d", cfg.APIPort),
+		}
+		// Handle Graceful shutshown
```
geoff-vball, line 162:
```suggestion
		// Handle Graceful shutdown
```

peers/app_request_network.go
```diff
@@ -398,6 +412,10 @@ func (n *appRequestNetwork) updateValidatorSet(
 
 func (n *appRequestNetwork) Shutdown() {
 	n.network.StartClose()
+	// Wait until startUpdateValidators goroutine stop.
+	n.logger.Info("Waiting for updating validator process to stop...")
+	n.wg.Wait()
```
geoff-vball, line 417: Sorry... I think we can get rid of this waitgroup altogether. I think we can just say that in any case where we are shutting down, the top-level context should have already been cancelled, so we don't have to worry about waiting for `startUpdateValidators` to exit, so there's no need to wait here for the goroutine to exit

peers/app_request_network.go
```diff
@@ -325,22 +330,31 @@ func (n *appRequestNetwork) TrackSubnet(subnetID ids.ID) {
 }
 
 func (n *appRequestNetwork) startUpdateValidators() {
+	n.wg.Add(1)
```
geoff-vball, line 333: a nit unrelated to this PR... If the entire function is wrapped in `go func() {}`, it would probably be easier just to call it as `go arNetwork.startUpdateValidators()` and not have the extra level of indentation

peers/app_request_network.go
ylg-avalabs: ok, I think wait for the goroutine to stop before return is more graceful shutdown.

PR #829, opened 2025-07-16
L-05 Lack of Graceful Shutdown for OS Signals

ylg-avalabs:
## Why this should be merged
Both `main.go` [[1](https://github.com/ava-labs/icm-services/blob/11c6cafa6d5a8572a2f48364e850491db5c86b80/relayer/main/main.go), [2](https://github.com/ava-labs/icm-services/blob/11c6cafa6d5a8572a2f48364e850491db5c86b80/signature-aggregator/main/main.go)] applications follow a similar initialization pattern. They set up all required resources, such as the P2P network, database connections, and HTTP servers, and then conclude with a blocking call ( `errgroup.Wait()` for the relayer, `http.ListenAndServe` for the aggregator) that keeps the process alive. Cleanup functions, like network.Shutdown(), are scheduled using defer.

This application structure does not handle OS signals like `SIGINT` or `SIGTERM`. When the process receives one of these signals, it terminates immediately and ungracefully. The blocking call at the end of main is interrupted, and the process exits before the deferred cleanup functions have a chance to run. The Redis database connection is not explicitly closed either. Furthermore, any in-flight HTTP requests are cut off. Lastly, the latest processed block height from the `CheckpointManager` may not be written to the database, causing the relayer to re-process blocks on restart.

Consider implementing a standard graceful shutdown mechanism in both `main` functions. This involves listening for OS signals on a channel and coordinating a clean, ordered shutdown of all components.

## How is this documented

### DIFF
--- a/relayer/main/main.go
+++ b/relayer/main/main.go
@@ -11,7 +11,6 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
-	"time"
 
 	"github.com/ava-labs/avalanchego/api/info"
 	"github.com/ava-labs/avalanchego/ids"
@@ -59,7 +58,12 @@ const (
 func main() {
 	cfg := buildConfig()
 
-	errGroup, ctx := errgroup.WithContext(context.Background())
+	// Create parent context with cancel function
+	parentCtx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// Create errgroup with parent context
+	errGroup, ctx := errgroup.WithContext(parentCtx)
 
 	// Modify the default http.DefaultClient globally
 	// TODO: Remove this temporary fix once the RPC clients used by the relayer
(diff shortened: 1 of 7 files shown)


relayer/main/main.go
```diff
@@ -313,9 +321,30 @@ func main() {
...
+		// No error for graceful shutdown
+		return nil
+	})
+
 	logger.Info("Initialization complete")
-	err = errGroup.Wait()
-	logger.Error("Relayer exiting.", zap.Error(err))
+	if err := errGroup.Wait(); err != nil {
+		logger.Fatal("Relayer exiting with error.", zap.Error(err))
+		os.Exit(1)
```
iansuvak, line 344: Generally LGTM but do we need these os.Exit(1)? Both here and sig-agg It seems like the only benefit is to return a non-zero exit code but downside is that we need to ensure that all defers have been executed already by the time we call this.

PR #831, opened 2025-07-16
L-01 Enhancements for Configuration Robustness

iansuvak:
## Why this should be merged
In the configuration setup for both the relayer and signature-aggregation services, multiple
areas have been identified where improvements could significantly enhance the robustness
and reliability of the configuration handling:
1. Currently, the configuration structs for both the relayer and signature-aggregation
services are populated using the `Unmarshal` function from a `Viper` instance. To
ensure a stricter adherence to the defined configuration schema and prevent unintended
fields from being processed, it is recommended to switch to the `UnmarshalExact`
function. This change will enforce a more precise mapping between the configuration file
and the corresponding struct, thereby eliminating the risk of processing extraneous or
misnamed configuration parameters.
2. The `Validate` function of the signature-aggregator service currently does not enforce
the presence of at least one `TrackedSubnetIDs` or `ManuallyTrackedPeers`.
Incorporating this check will ensure that the service is configured with a minimum
operational context, thereby preventing scenarios where the service starts without any
actionable configuration, leading to potential operational inefficiencies.
3. Both the signature-aggregator and relayer services could benefit from enhanced error
reporting within their respective `Validate` functions. By wrapping the returned errors
with additional context, developers and operators can more easily diagnose and rectify
configuration issues.

Consider applying the above suggestions for a more properly built configuration to tighten the
validation process and ensuring that the configuration accurately reflects the intended setup.
This will help reduce the potential for runtime errors and misconfigurations.

## How is this documented

### DIFF
--- a/relayer/config/viper.go
+++ b/relayer/config/viper.go
@@ -33,10 +33,16 @@ func BuildViper(fs *pflag.FlagSet) (*viper.Viper, error) {
 	// Map flag names to env var names. Flags are capitalized, and hyphens are replaced with underscores.
 	v.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
 
-	filename, err := fs.GetString(ConfigFileKey)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get config file flag: %w", err)
+	var filename string
+	var err error
+	filename = os.Getenv(ConfigFileEnvKey)
+	if filename == "" {
+		filename, err = fs.GetString(ConfigFileKey)
+		if err != nil {
+			return nil, fmt.Errorf("config file not set via flag or environment variable: %w", err)
+		}
 	}
+
 	v.SetConfigFile(filename)
 	v.SetConfigType("json")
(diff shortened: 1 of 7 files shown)


relayer/config/viper.go
```diff
@@ -32,17 +32,11 @@ func BuildViper(fs *pflag.FlagSet) (*viper.Viper, error) {
...
 	v.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
-	if err := v.BindPFlags(fs); err != nil {
-		return nil, err
-	}
 
-	// Verify that required flags are set
-	if !v.IsSet(ConfigFileKey) {
-		DisplayUsageText()
-		return nil, fmt.Errorf("config file not set")
+	filename, err := fs.GetString(ConfigFileKey)
```
iansuvak, line 36: this is the only usage of the passed in flagset variable here which means it should probably be pulled out. Didn't do it just yet because it would require unit test changes but I can

geoff-vball (approved):

PR #832, opened 2025-07-16
N-02 Loss of Component-Level Metrics Compromises Observability

ylg-avalabs:
## Why this should be merged
Across various system components, metrics are either inadvertently discarded or intentionally ignored. This is done by the initialization of instances with their own isolated Prometheus registries, which are subsequently never exposed for observation.

This pattern is evident in the initialization of the AdaptiveTimeoutManager within [external_handler.go](https://github.com/ava-labs/icm-services/blob/11c6cafa6d5a8572a2f48364e850491db5c86b80/peers/external_handler.go#L56), as well as the message.Creator in both the [relayer](https://github.com/ava-labs/icm-services/blob/11c6cafa6d5a8572a2f48364e850491db5c86b80/relayer/main/main.go#L183) and [signature-aggregator](https://github.com/ava-labs/icm-services/blob/11c6cafa6d5a8572a2f48364e850491db5c86b80/signature-aggregator/main/main.go#L108) main functions, where new Prometheus registries are created but not made accessible for monitoring purposes. This approach, particularly the deliberate decision to isolate the metrics of the message.Creator, results in a trade-off between reducing potential metric noise and sacrificing valuable insights into the system's operational state.

Consider reevaluating the current strategy regarding metric collection and exposure. Specifically, integrating all component level metrics into a unified, centrally exposed Prometheus registry would enable a holistic view of the system's performance and health.

## How is this documented

### DIFF
--- a/relayer/checkpoint/checkpoint.go
+++ b/relayer/checkpoint/checkpoint.go
@@ -5,6 +5,7 @@ package checkpoint
 
 import (
 	"container/heap"
+	"fmt"
 	"strconv"
 	"sync"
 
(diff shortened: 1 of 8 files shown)


iansuvak (approved):
This LGTM. The context we should be aware of here is https://github.com/ava-labs/icm-services/pull/748 where the changes were made to avoid panics when importing sig-agg / network would cause panics since we were hardcoding passing defaultRegisterer then. This doesn't run into this issue and I've ran it manually and confirmed that the new time out manager metrics appear in at the `/metrics` endpoint correctly

PR #840, opened 2025-07-23
Use GetCurrentValidators API

ylg-avalabs:
## Why this should be merged
- Currently, icm-services dependents on a [patch branch of ava-labs/avalanchego](https://github.com/ava-labs/avalanchego/tree/icm-services-patch-v1.13.2). 
- The patch contains [GetCurrentL1Validators](https://github.com/ava-labs/avalanchego/compare/master...icm-services-patch-v1.13.2#diff-07adbc0edb62e0f0bd7cbcfc65a4e6d8be518002233ca81a8410b0e0ba5bbbbcR336) API. However, the current L1 validators can be get from [GetCurrentValidators](https://github.com/ava-labs/avalanchego/blob/v1.13.2/vms/platformvm/client.go#L70)
- If the validator is L1 validator, the L1 validator info is store in [ClientPermissionlessValidator.ClientL1Validator](https://github.com/ava-labs/avalanchego/blob/ffd7af93c0614696dcfe5fba1adf8e19d254a6fa/vms/platformvm/client_permissionless_validator.go#L117-L122)
- As a result, we can filter the L1 validator from the validator list returned from `GetCurrentL1Validators`

## How is this documented

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.23.11
 
 require (
 	github.com/alexliesenfeld/health v0.8.1
-	github.com/ava-labs/avalanchego v1.13.3-0.20250701190537-839ace23368a // branch icm-services-patch-v1.13.2
+	github.com/ava-labs/avalanchego v1.13.3-0.20250701190514-843a2b9bb3d8 // branch icm-services-patch-v1.13.2
 	github.com/ava-labs/icm-contracts v1.0.9-0.20250701204604-b34a75d8c753
 	github.com/ava-labs/libevm v1.13.14-0.3.0.rc.1
 	github.com/ava-labs/subnet-evm v0.7.5
@@ -65,7 +65,7 @@ require (
 	github.com/Masterminds/semver/v3 v3.3.0 // indirect
 	github.com/Microsoft/go-winio v0.6.2 // indirect
 	github.com/OpenPeeDeeP/depguard/v2 v2.2.1 // indirect
-	github.com/StephenButtolph/canoto v0.15.0 // indirect
+	github.com/StephenButtolph/canoto v0.17.1 // indirect
 	github.com/VictoriaMetrics/fastcache v1.12.1 // indirect
 	github.com/alecthomas/go-check-sumtype v0.3.1 // indirect
 	github.com/alexkohler/nakedret/v2 v2.0.5 // indirect
@@ -75,7 +75,7 @@ require (
 	github.com/antlr4-go/antlr/v4 v4.13.1 // indirect
 	github.com/ashanbrown/forbidigo v1.6.0 // indirect
 	github.com/ashanbrown/makezero v1.2.0 // indirect
-	github.com/ava-labs/coreth v0.15.2-rc.0.0.20250620163936-0058a7ec6d9d // indirect
+	github.com/ava-labs/coreth v0.15.2 // indirect
 	github.com/aws/aws-sdk-go-v2/credentials v1.17.70 // indirect
 	github.com/aws/aws-sdk-go-v2/feature/ec2/imds v1.16.32 // indirect
 	github.com/aws/aws-sdk-go-v2/internal/configsources v1.3.36 // indirect
@@ -148,7 +148,7 @@ require (
 	github.com/getsentry/sentry-go v0.18.0 // indirect
 	github.com/ghostiam/protogetter v0.3.9 // indirect
 	github.com/go-chi/chi/v5 v5.2.2 // indirect
-	github.com/go-cmd/cmd v1.4.1 // indirect
+	github.com/go-cmd/cmd v1.4.3 // indirect
 	github.com/go-critic/go-critic v0.12.0 // indirect
 	github.com/go-logr/logr v1.4.2 // indirect
 	github.com/go-logr/stdr v1.2.2 // indirect
--- a/signature-aggregator/aggregator/aggregator.go
+++ b/signature-aggregator/aggregator/aggregator.go
@@ -23,6 +23,7 @@ import (
 	"github.com/ava-labs/avalanchego/subnets"
 	"github.com/ava-labs/avalanchego/utils/constants"
 	"github.com/ava-labs/avalanchego/utils/crypto/bls"
+	"github.com/ava-labs/avalanchego/utils/json"
 	"github.com/ava-labs/avalanchego/utils/logging"
 	"github.com/ava-labs/avalanchego/utils/rpc"
 	"github.com/ava-labs/avalanchego/utils/set"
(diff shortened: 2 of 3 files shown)


signature-aggregator/aggregator/aggregator.go
```diff
@@ -249,7 +250,40 @@ func (s *SignatureAggregator) CreateSignedMessage(
...
+			if err != nil {
+				log.Error(
+					"Failed to fetch current L1 validators",
+					zap.String("subnetID", subnetID.String()),
+					zap.Error(err),
+				)
+				return nil, fmt.Errorf("failed to fetch current L1 validators: %w", err)
+			}
+
+			l1Validators := make([]platformvmapi.APIL1Validator, 0, len(validators))
```
iansuvak, line 263: we don't have tests for this functionality yet, but we should at the very least add unit tests for it. Ideally E2E but creating underfunded nodes in could be tricky with tmpnet.

tests/e2e_test.go
```diff
@@ -133,15 +133,22 @@ var _ = ginkgo.BeforeSuite(func() {
...
+
 	// Convert the subnets to sovereign L1s
 	for _, subnet := range localNetworkInstance.GetL1Infos() {
 		localNetworkInstance.ConvertSubnet(
 			networkStartCtx,
 			subnet,
 			teleporterTestUtils.PoAValidatorManager,
-			[]uint64{units.Schmeckle, units.Schmeckle},
+			[]uint64{units.Schmeckle, units.Schmeckle, units.Schmeckle, units.Schmeckle},
+			[]uint64{fundedBalance, fundedBalance, fundedBalance, underfundedBalance},
```
iansuvak, line 148: This does test it but very implicitly. I would document the reason for why we are adding a validator and perhaps add an explicit test that using signature aggregator that explicitly confirms that underfunded node is not a member of the aggregate signature.

go.mod
```diff
@@ -4,8 +4,8 @@ go 1.23.11
 
 require (
 	github.com/alexliesenfeld/health v0.8.1
-	github.com/ava-labs/avalanchego v1.13.3-0.20250701190537-839ace23368a // branch icm-services-patch-v1.13.2
-	github.com/ava-labs/icm-contracts v1.0.9-0.20250701204604-b34a75d8c753
+	github.com/ava-labs/avalanchego v1.13.3-0.20250701190514-843a2b9bb3d8 // branch icm-services-patch-v1.13.2
```
michaelkaplan13, line 7: To confirm my understanding, after this change ICM services will still depend on a branch of AvalancheGo, but just not the `GetCurrentL1Validators` API added in that branch anymore? And that's why we're reverting the dependency to an older commit on the same branch?

signature-aggregator/aggregator/aggregator.go
```diff
@@ -179,6 +178,115 @@ func (s *SignatureAggregator) connectToQuorumValidators(
...
+				log.Debug(
+					"Skipping non-L1 validator",
+					zap.String("nodeID", v.NodeID.String()),
+				)
+				continue
+			}
+
+			l1Validator := v.ClientL1Validator
+
+			if l1Validator.Balance == nil {
```
geoff-vball, line 212: Do we know what cases the Balance would be nil? And shouldn't these nodes get added to the underfunded list?

signature-aggregator/aggregator/aggregator.go
```diff
@@ -179,6 +178,115 @@ func (s *SignatureAggregator) connectToQuorumValidators(
...
+			zap.String("signingSubnetID", signingSubnet.String()),
+			zap.Error(err),
+		)
+		return nil, err
+	}
+
+	return underfundedL1Nodes, nil
+}
+
+func (s *SignatureAggregator) getExcludedValidators(
```
geoff-vball, line 248: Would it make more sense to cache the set of `excluded` validators instead of the set of `underfunded` validators?

signature-aggregator/aggregator/aggregator.go
```diff
@@ -812,6 +876,10 @@ func (s *SignatureAggregator) aggregateSignatures(
 			return nil, set.Bits{}, fmt.Errorf("%s: %w", msg, err)
 		}
 		signatures = append(signatures, sig)
+		log.Info(
+			"Adding signature to aggregate",
+			zap.Int("index", i),
+		)
```
geoff-vball, line 882: It might make more sense just to have one info log of the bit-set once we're done filling it

tests/e2e_test.go
```diff
@@ -133,15 +133,22 @@ var _ = ginkgo.BeforeSuite(func() {
 		teleporterInfo.DeployTeleporterRegistry(subnet, fundedKey)
 	}
 
+	// Default balance for funded accounts
+	fundedBalance := units.Avax * 100
+	// Underfunded balance for validators
+	underfundedBalance := 2048*units.NanoAvax - 1
```
geoff-vball, line 139: This should probably be a constant somewhere

signature-aggregator/aggregator/aggregator.go
```diff
@@ -178,6 +179,124 @@ func (s *SignatureAggregator) connectToQuorumValidators(
...
+	// Set of underfunded L1 validator nodes
+	underfundedNodes := set.NewSet[ids.NodeID](0)
+	for _, validator := range l1Validators {
+		if validator.Balance == nil {
+			underfundedNodes.Add(validator.NodeID)
+			log.Warn(
+				"Skipping L1 validator with nil balance",
+				zap.String("nodeID", validator.NodeID.String()),
+			)
+			continue
```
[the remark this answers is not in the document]
ylg-avalabs: We added the validator to the underfundedNodes set. I will updated the log message.

signature-aggregator/aggregator/aggregator.go
ylg-avalabs: This is used for debug. I will remove this.

signature-aggregator/aggregator/aggregator.go
geoff-vball: Ah, I see we get the Validator set from a different API. That makes sense then :)

signature-aggregator/aggregator/aggregator_test.go
```diff
@@ -479,3 +479,196 @@ func TestUnmarshalResponse(t *testing.T) {
...
+					ClientStaker: platformvm.ClientStaker{NodeID: nodeID1},
+					ClientL1Validator: platformvm.ClientL1Validator{
+						ValidationID: &validationID1,
+						Balance:      &funded},
+				},
+				{
+					ClientStaker: platformvm.ClientStaker{NodeID: nodeID2},
+					ClientL1Validator: platformvm.ClientL1Validator{
+						ValidationID: &validationID2,
+						Balance:      &funded},
```
michaelkaplan13, line 615: May make sense to have one of the validation IDs be `underfunded` here, to confirm that having at least 1 "funded" validation ID is sufficient

michaelkaplan13 (approved):
I left a few more minor suggestions, but all of them are just for tests. Otherwise LGTM

iansuvak (approved):

PR #850, opened 2025-07-29
Signature aggregator requestnum

anviimishra:
## Why this should be merged
In the version before, even requests to the signature aggregator would return back after a timeout. This would cause the aggregator to have a large amount of signing retries and take up more time. This version eliminates the number of retries by fixing this issue.

## How this works
The change made ensures that only odd numbered requests are made to the signature aggregator. This is important because the p2p network that the aggregator spins out accepts only odd numbered requests.

## How this was tested
It was tested through the ICM load simulation tool that I've been working. There was a noticeable drop in relaying or signing times and no retries. 

## How is this documented
There are in line comments explaining how the request number is first generated and then incremented.


geoff-vball:
You probably just have to fix the unit test here https://github.com/ava-labs/icm-services/blob/remove-patch/signature-aggregator/aggregator/aggregator_test.go#L242

iansuvak (approved):

michaelkaplan13 (approved):
🙌

geoff-vball (approved):

PR #869, opened 2025-08-13
update subnet-evm/avalanche go and interface changes

iansuvak:
## Why this should be merged

Brings avalanchego dependency up to date with v1.13.4 with the patch still on top of it. 
## How this works

Standard update but also defines a new interface since the pchain client is now a struct instead of an interface so defines a new minimal interface in order to still be able to mock it out for testing. 

## How this was tested

CI

## How is this documented


geoff-vball (approved):
Should we also update the `icm-contracts` submodule to the latest version that also uses the updated deps?

geoff-vball (approved):

PR #939, opened 2025-09-29
Support ACP181

ylg-avalabs:
## Why this should be merged
This PR refactors the validator caching architecture to prepare for ACP-181 "P-Chain Epoched Views" implementation. 

## How this works
API Changes:
- GetCanonicalValidators() now takes pchainHeight parameter
- CreateSignedMessage() updated to accept P-Chain height instead of destination blockchain ID
- GetCurrentCanonicalValidatorSet() enhanced to fetch validators at specified height
- API defaults to ProposedHeight when height not specified

Height-Based Logic:
- Current Validators: pchainHeight = pchainapi.ProposedHeight → Latest validators
- Epoched Validators: pchainHeight = specificHeight → Historical validators at that height
- Cache Selection: Automatic based on height type

Integration:
- Relayer: Added getPChainHeightForValidatorSet() for Granite detection

## How this was tested
- Unit tests and e2e tests

## How is this documented

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,8 +6,8 @@ require (
 	connectrpc.com/connect v1.19.1
 	github.com/alexliesenfeld/health v0.8.1
 	github.com/ava-labs/avalanchego v1.13.6-rc.1.0.20251013184242-946ea96c2844 // branch icm-services-patch-v1.13.6
-	github.com/ava-labs/icm-contracts v1.0.10-0.20251013183254-551a111edd7c
-	github.com/ava-labs/libevm v1.13.15-0.20251002164226-35926db4d661
+	github.com/ava-labs/icm-contracts v1.0.10-0.20251016045813-a42672a4e254
+	github.com/ava-labs/libevm v1.13.15-0.20251003195629-910e897c54c3
 	github.com/ava-labs/subnet-evm v0.7.10-0.20251013155523-47a1ce400863
 	github.com/aws/aws-sdk-go-v2 v1.39.2
 	github.com/aws/aws-sdk-go-v2/config v1.31.12
--- a/go.sum
+++ b/go.sum
@@ -108,10 +108,10 @@ github.com/ava-labs/coreth v0.15.4-rc.4 h1:ze7/IwDptWG1u2d32uUZz9Ix9ycVUtlB8Jufu
 github.com/ava-labs/coreth v0.15.4-rc.4/go.mod h1:yVwuMyPkZ48xzZ0y2OdIwaoUqvSsgPYoodyX9BZJ2uo=
 github.com/ava-labs/firewood-go-ethhash/ffi v0.0.12 h1:aMcrLbpJ/dyu2kZDf/Di/4JIWsUcYPyTDKymiHpejt0=
 github.com/ava-labs/firewood-go-ethhash/ffi v0.0.12/go.mod h1:cq89ua3iiZ5wPBALTEQS5eG8DIZcs7ov6OiL4YR1BVY=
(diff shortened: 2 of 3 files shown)


peers/app_request_network.go
```diff
@@ -424,24 +433,82 @@ func (c *CanonicalValidators) GetValidator(nodeID ids.NodeID) (*warp.Validator,
...
+	n.logger.Debug("Getting validator set at P-Chain height",
+		zap.Stringer("subnetID", subnetID),
+		zap.Uint64("pchainHeight", pchainHeight),
+		zap.Bool("isProposedHeight", pchainHeight == pchainapi.ProposedHeight),
+	)
+
+	var validatorSet avalancheWarp.CanonicalValidatorSet
+	var err error
+
+	if pchainHeight == pchainapi.ProposedHeight {
```
michaelkaplan13, line 453: We shouldn't ever use the `pchainapi.ProposedHeight` constant as a key in the cache, because the "proposed" height is constantly updating to represent new actual P-Chain heights.

cache/lru_cache.go
```diff
@@ -0,0 +1,55 @@
...
+			return value, nil
+		}
+		c.lock.RUnlock()
+	}
+
+	// Fetch new value
+	newValue, err := fetchFunc(key)
+	if err != nil {
+		var zero V
+		return zero, err
```
geoff-vball:
```suggestion
		return *new(V), err
```

relayer/application_relayer.go
```diff
@@ -317,6 +373,38 @@ func (r *ApplicationRelayer) RelayerID() database.RelayerID {
 	return r.relayerID
 }
 
+// getPChainHeightForValidatorSet determines the appropriate P-Chain height for validator set selection
+// Returns ProposedHeight for current validators if Granite is not activated, or the epoch P-Chain height if activated
+func (r *ApplicationRelayer) getPChainHeightForValidatorSet(ctx context.Context) (uint64, error) {
```
geoff-vball: This name confused me a little at first. I would probably prefer something like `getPChainHeightForDestination`

peers/config.go
```diff
@@ -18,4 +19,5 @@ type Config interface {
 	GetAllowPrivateIPs() bool
 	GetTrackedSubnets() set.Set[ids.ID]
 	GetTLSCert() *tls.Certificate
+	GetUpgradeConfig() *upgrade.Config
```
geoff-vball: So I think it might make more sense to instead have `*upgrade.Config` as a field in the `AppRequestNetwork` object, which would make it accessible everywhere. We can take the initialize func that's currently on the config and just call it in the initialization of `AppRequestNetwork` to get the config once. That way we also don't have to remember to call `Initialize` on the config, which I'm not a huge fan of.

tests/e2e_test.go
```diff
@@ -39,8 +40,16 @@ var (
...
+
+	e2eFlags *e2e.FlagVars
 )
 
+func TestMain(m *testing.M) {
+	e2eFlags = e2e.RegisterFlags()
+	flag.Parse()
+	os.Exit(m.Run())
+}
+
```
geoff-vball: Is this used somewhere/somehow?

geoff-vball (commented):
Just a few small comments. Otherwise LGTM :)

tests/e2e_test.go
iansuvak: Yes -- I do believe this is necessary to be able to pass custom flags such as `--activate-granite` to ginkgo test run.

geoff-vball: Oh interesting, yeah I see how it's used in subnet-evm now. I had the directory excluded from my searches

PR #951, opened 2025-10-08
Fetch network upgrade info at relayer config initialize time

iansuvak:
## Why this should be merged

Fetches network upgrade information at config building time that allows for switching behavior based on network upgrade information without hardcoding of upgrade times like was done previously. 

## How is this documented

### DIFF
--- a/relayer/config/config.go
+++ b/relayer/config/config.go
@@ -11,12 +11,15 @@ import (
 	"net/url"
 	"reflect"
 	"strings"
+	"time"
 
 	basecfg "github.com/ava-labs/icm-services/config"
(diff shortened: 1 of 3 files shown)


relayer/config/config.go
```diff
@@ -322,6 +340,10 @@ func (c *Config) LogSafeField() zap.Field {
 	return zap.Any("config", c.sanitizeForLogging())
 }
 
+func (c *Config) IsGraniteActivated() bool {
```
iansuvak, line 343: Should consider just exposing the `upgrade.Config` itself instead of this?

PR #958, opened 2025-10-16
Formatting, deprecated import

geoff-vball:
## Why this should be merged
Mostly formatting changes I stumbled through while writing the sig-agg PR

## How is this documented

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,8 +6,8 @@ require (
 	connectrpc.com/connect v1.19.1
 	github.com/alexliesenfeld/health v0.8.1
 	github.com/ava-labs/avalanchego v1.13.6-rc.1.0.20251013184242-946ea96c2844 // branch icm-services-patch-v1.13.6
-	github.com/ava-labs/icm-contracts v1.0.10-0.20251013183254-551a111edd7c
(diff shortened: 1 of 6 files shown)


peers/validators/canonical_validator_client.go
```diff
@@ -34,7 +34,7 @@ type CanonicalValidatorState interface {
 		ctx context.Context,
 		subnetID ids.ID,
 		pchainHeight uint64,
-	) (avalancheWarp.CanonicalValidatorSet, error)
+	) (validators.WarpSet, error)
```
michaelkaplan13: Were the old types left in place in AvalancheGo? I'm surprised this was building at all previously...

michaelkaplan13 (approved):
Thanks for splitting this out. Would be great for these to be caught/enforced by a linter in the future. We could look into incorporating AvalancheGo's full linting scripts to keep it standardized.

peers/validators/canonical_validator_client.go
geoff-vball: Left as a type-alias, but might be removed later https://github.com/ava-labs/avalanchego/blob/578d0a915b83cd9ba1243591085ce192d93cdd06/vms/platformvm/warp/validator.go#L33

iansuvak (approved):

PR #975, opened 2025-10-22
Properly add options to info api

geoff-vball:
## How is this documented


iansuvak (approved):

peers/info_client.go
```diff
@@ -74,5 +74,5 @@ func (i *InfoAPI) Uptime(ctx context.Context) (*info.UptimeResponse, error) {
 }
 
 func (i *InfoAPI) Upgrades(ctx context.Context, options ...rpc.Option) (*upgrade.Config, error) {
```
iansuvak: We should merge these together with the ones on the client or remove from interface

PR #1011, opened 2025-11-11
Remove all fetches to `GetAllValidatorsAt` with `ProposedHeight` 

iansuvak:
## Why this should be merged

Idle relayers (with no messages to relay) would never request validator sets for a cacheable height and would therefore still fetch proposed heights at health-check time. 

## How this works

- Healthcheck logs and returns no errors in case there is no cached height
- Periodic validator refresh fetches latest PChain height instead of proposedHeight

## How is this documented


geoff-vball (commented):
This all LGTM

PR #1014, opened 2025-11-11
Cache current epoch

### LINKED ISSUE #994: Cache current epoch for the epoch duration
**Context and scope**
Currently we fetch the current epoch on a per message basis. We should consider reducing the number of these calls since the value is static over the epoch duration. 

These values are cached on the server so it's not a big deal either way but there is no reason to require the round trip to complete in order to relay the message. 

One option here would be to have a TTL cache for the duration of the epoch which is easily calculable with `epoch.StartTime + epochDuration`. This doesn't guarantee that the new epoch has started if the sealing block hasn't been accepted but at that point we can keep making requests for each message until the endpoint returns a new epoch value which can then be cached. 

**Open questions**

ylg-avalabs:
## Why this should be merged
- Closes #994 
- Cache current epoch for the epoch duration

## How is this documented

### DIFF
--- a/vms/evm/destination_client.go
+++ b/vms/evm/destination_client.go
@@ -40,6 +40,10 @@ const (
 	poolTxsPerAccount             = 16
 	pendingTxRefreshInterval      = 2 * time.Second
 	defaultBlockAcceptanceTimeout = 30 * time.Second
+
+	// epochCacheKey is the singleflight key for epoch caching
+	// Since we only cache one epoch per destination, the key is constant
+	epochCacheKey = "epoch"
 )
 
 // Client interface wraps the ethclient.Client interface for mocking purposes.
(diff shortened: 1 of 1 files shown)


cache/ttl_cache.go
```diff
@@ -75,6 +76,65 @@ func (c *TTLCache[K, V]) Get(key K, fetchFunc func(K) (V, error), invalidate boo
...
+			}
+			if !isExpired {
+				return item.value, nil
+			}
+		}
+	}
+
+	keyStr := keyToString(key)
+
+	v, err, _ := c.sfGroup.Do(keyStr, func() (interface{}, error) {
```
geoff-vball: This func should return (V, error), then we also don't have to cast it on line 135

vms/evm/destination_client.go
```diff
@@ -205,6 +219,27 @@ func NewDestinationClient(
...
+		return nil, err
+	}
+
+	baseURL := fmt.Sprintf("%s://%s", endpoint.Scheme, endpoint.Host)
+	blockchainID := destinationBlockchain.BlockchainID
+	proposerClient := peers.NewProposerVMAPI(baseURL, blockchainID, &destinationBlockchain.RPCEndpoint)
+
+	logger.Info("Initialized ProposerVM client for destination chain",
+		zap.Stringer("destinationBlockchainID", destinationID),
+	)
```
geoff-vball: I don't think this log provides any value, `NewProposerVMAPI` can't fail

vms/evm/destination_client.go
```diff
@@ -521,3 +548,89 @@ func (c *destinationClient) BlockGasLimit() uint64 {
...
+				zap.Error(err),
+			)
+			return pchainapi.ProposedHeight, nil
+		}
+		return height, nil
+	}
+
+	if c.proposerClient == nil {
+		return 0, fmt.Errorf("proposerClient is required when Granite is activated")
+	}
```
geoff-vball: I don't think this check gets us anything. We're always setting `proposerClient` to be non-nil. Whether or not it's a valid endpoint is another story 🙃

PR #1034, opened 2025-11-19
Clean up listener and subscriber logging

geoff-vball:
## Why this should be merged
Uses `logger.With()` to pass the blockchain ID to the logger for the listener and subscriber so that we don't need to add it to every log line. 

Consolidates some duplicate logging.

## How is this documented


vms/evm/subscriber.go
```diff
@@ -195,7 +181,7 @@ func (s *subscriber) Subscribe(retryTimeout time.Duration) error {
 
 	err := s.subscribe(retryTimeout)
 	if err != nil {
-		return errors.New("failed to subscribe to node")
+		return err
```
geoff-vball: Not sure why we weren't propagating the error here.

ylg-avalabs (approved):

PR #1035, opened 2025-11-19
Cleanup logging in application relayer

geoff-vball:
## Why this should be merged
Cleans up logging in the application relayer. Also makes sure that all the context from the message handler (warp message ID etc) are logged in all logs in the application relayer.

## How is this documented


mdelle1 (commented):
LGTM

PR #1036, opened 2025-11-19
Remove logger from message handler factory

geoff-vball:
## Why this should be merged
The message handler factory doesn't need a logger. Whatever service wants to create a message handler can pass their own logger when creating a message handler.

## How is this documented

### DIFF
--- a/database/database.go
+++ b/database/database.go
@@ -6,11 +6,12 @@
 package database
 
 import (
+	"fmt"
+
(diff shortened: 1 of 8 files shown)


iansuvak (approved):
Thanks for also removing the innermost logs and letting the caller log the error.

iansuvak (approved):

PR #1048, opened 2025-11-24
Convert AppRequestNetwork from interface to concrete struct

ylg-avalabs:
## Why this should be merged

The `AppRequestNetwork` struct had grown organically over time and contained methods that didn't logically belong together. This PR refactors the code to improve separation of concerns by:

1. **Extracting validator management logic** into a dedicated `ValidatorManager` struct
2. **Creating `NetworkInfo`** to handle network metadata and upgrade configuration
3. **Converting interfaces to concrete structs** where interfaces weren't needed (removed unnecessary abstraction)
4. **Improving method organization** by making `GetNetworkHealthFunc` a method instead of a standalone function

This makes the codebase more maintainable, easier to test, and clearer in its responsibilities.

### New Structure

**`ValidatorManager`** (`peers/validator_manager.go`):
- Manages all validator-related operations: caching, fetching, and updating validator sets
- Handles P-Chain height tracking and lookback limits
- Methods: `GetAllValidatorSets`, `GetValidatorSet`, `UpdateTrackedValidatorSet`, `GetLatestSyncedPChainHeight`, `StartCacheValidatorSets`, etc.

**`NetworkInfo`** (`peers/network_info.go`):
- Provides network metadata and upgrade information
- Methods: `IsGraniteActivated`, `GetGraniteEpochDuration`

**`AppRequestNetwork`** (refactored):
- Now focused on P2P network operations: sending messages, tracking subnets, registering requests
- Embeds `ValidatorManager` and `NetworkInfo` for validator and network metadata functionality
- `GetNetworkHealthFunc` converted from standalone function to method

### Key Changes

1. **Removed unnecessary interfaces**: `AppRequestNetwork` is now a concrete struct since it's only used internally
2. **Better separation**: Validator logic, network info, and P2P operations are clearly separated
3. **Improved testability**: Created `validator_manager_test.go` with dedicated tests for validator logic

## How this was tested
- Unit tests & E2E tests

## How is this documented

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/ava-labs/subnet-evm v0.8.1-0.20251111182129-7fc05124d976
 	github.com/aws/aws-sdk-go-v2 v1.40.0
 	github.com/aws/aws-sdk-go-v2/config v1.32.2
-	github.com/aws/aws-sdk-go-v2/service/kms v1.49.0
+	github.com/aws/aws-sdk-go-v2/service/kms v1.49.1
 	github.com/cenkalti/backoff/v4 v4.3.0
 	github.com/hashicorp/golang-lru/v2 v2.0.7
 	github.com/onsi/ginkgo/v2 v2.27.2
--- a/go.sum
+++ b/go.sum
@@ -136,8 +136,8 @@ github.com/aws/aws-sdk-go-v2/service/internal/accept-encoding v1.13.3 h1:x2Ibm/A
 github.com/aws/aws-sdk-go-v2/service/internal/accept-encoding v1.13.3/go.mod h1:IW1jwyrQgMdhisceG8fQLmQIydcT/jWY21rFhzgaKwo=
 github.com/aws/aws-sdk-go-v2/service/internal/presigned-url v1.13.14 h1:FIouAnCE46kyYqyhs0XEBDFFSREtdnr8HQuLPQPLCrY=
 github.com/aws/aws-sdk-go-v2/service/internal/presigned-url v1.13.14/go.mod h1:UTwDc5COa5+guonQU8qBikJo1ZJ4ln2r1MkF7Dqag1E=
-github.com/aws/aws-sdk-go-v2/service/kms v1.49.0 h1:QbztCKLI2qnjFZ/QYq3hZ8SW7SnTwB5h0NjREtKXIGo=
-github.com/aws/aws-sdk-go-v2/service/kms v1.49.0/go.mod h1:NZo9WJqQ0sxQ1Yqu1IwCHQFQunTms2MlVgejg16S1rY=
+github.com/aws/aws-sdk-go-v2/service/kms v1.49.1 h1:U0asSZ3ifpuIehDPkRI2rxHbmFUMplDA2VeR9Uogrmw=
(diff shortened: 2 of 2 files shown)


geoff-vball (commented):
Just a few things. Not sure if we want to merge this or https://github.com/ava-labs/icm-services/pull/1045 first. @ylg-avalabs it would be great if you could review that PR

peers/validator_manager.go
```diff
@@ -0,0 +1,289 @@
...
+
+func (v *ValidatorManager) GetAllValidatorSets(
+	ctx context.Context,
+	pchainHeight uint64,
+) (map[ids.ID]snowVdrs.WarpSet, error) {
+	// If we're getting the proposed height, bypass the cache and get the latest data
+	// We can't cache this call because we don't know the actual P-Chain height being returned.
+	if pchainHeight == pchainapi.ProposedHeight {
+		return v.validatorClient.GetAllValidatorSets(ctx, pchainHeight)
+	}
```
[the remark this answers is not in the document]
geoff-vball: Using `GetLatestHeight` isn't as bad IMO because we actually know what height we're requesting. I'd be in favour of removing support for `ProposedHeight`, or have it grab the epoched height for that chain. It's the requesting an unknown height that I think we should get rid of.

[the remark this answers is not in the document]
ylg-avalabs: If it is called with `ProposedHeight`, can we just directly call [GetLatestValidatorSets](https://github.com/ava-labs/icm-services/pull/1048/files#diff-4c7914b27cc3f773824d15594daeb37caa71e2a369e68d0ec64148401ad20921R80)? This function will get the latest height then call GetAllValidatorSets with the latest height.

signature-aggregator/aggregator/aggregator.go
```diff
@@ -64,7 +64,7 @@ var (
 )
 
 type SignatureAggregator struct {
-	network                peers.AppRequestNetwork
+	network                peers.AppRequestNetworkTestHelper
```
iansuvak: nit: I know that this is temporary but I don't think that we should be merging anything named `TestHelper` into main in the critical code path. If we are temporarily keeping the interface I would still call it something else.

peers/validator_manager_test.go
```diff
@@ -0,0 +1,174 @@
...
+	snowVdrs "github.com/ava-labs/avalanchego/snow/validators"
+	"github.com/ava-labs/avalanchego/utils/logging"
+	pchainapi "github.com/ava-labs/avalanchego/vms/platformvm/api"
+	"github.com/ava-labs/icm-services/cache"
+	validator_mocks "github.com/ava-labs/icm-services/peers/clients/mocks"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+)
+
+func TestGetLatestSyncedPChainHeight(t *testing.T) {
```
geoff-vball: I personally believe unit tests like this have negative value. They're really complicated code testing very simple functionality. Perhaps if we encapsulated the syncing logic in it's own struct it would be simpler? Definitely open to other opinions here though

ylg-avalabs: Agree! And it is already covered in the higher level when we test the `AppRequestNetwork`. Removed it.

geoff-vball (approved):

PR #1056, opened 2025-11-26
Graft #2: Do graft

geoff-vball:
## Why this should be merged
Actually grafts `icm-contracts` into this repo, keeping all commit history.

## How is this documented


iansuvak (commented):
Confirmed that the graft of the commits is present here and that they match but we should re-sync when ready to merge since there are a few commits on top since then. Also the latest merge commit is unverified right now

PR #1058, opened 2025-11-26
Graft #4: Consolidate contracts, abi-bindings

geoff-vball:
## Why this should be merged
This is probably the largest PR in this chain. 

1. Consolidates the contracts into one place. 
2. Moves the abi-bindings to the top level.
3. Consolidates a few shell scripts.

## How is this documented


batconjurer (approved):

scripts/versions.sh
```diff
@@ -26,10 +28,19 @@ function extract_commit() {
   echo "$version"
 }
 
+# ICM_SERVICES_VERSION is currently needed for the contracts E2E tests but is not a direct dependency
+# since that would create a circular dependency. We should refactor the code until this is no longer the case.
+# ICM_SERVICES_VERSION=${ICM_SERVICES_VERSION:-'signature-aggregator-v1.0.0-rc.0'}
+ICM_SERVICES_VERSION=${ICM_SERVICES_VERSION:-'9564f00d296c7daeffbf26c4cc4866b3b6e98185'}
```
iansuvak: We shouldn't need this any longer. I'm fine waiting until the merge is complete to remove it

.golangci.yml
```diff
@@ -16,7 +16,6 @@ issues:
 linters:
   disable-all: true
   enable:
-    - goconst
```
iansuvak: what breaks if we don't remove this? Should we have an issue to put it back in a separate PR? I see that we have it in icm-contracts specific `.golangci.yml`?

PR #1061, opened 2025-11-28
Graft #7: Remove circular dependency on sig-agg

geoff-vball:
## Why this should be merged
Removes the circular dependency in `icm-contracts` that would download the `sig-agg` binary. We now just build it in-place.

## How is this documented


iansuvak (approved):
🚀

batconjurer (approved):

PR #1071, opened 2025-12-03
Update solc to 0.8.30

geoff-vball:
## Why this should be merged
Updates solc to `v0.8.30`.

## How this works
The compiler complains about `pragma solidity ^0.8.25` not matching semver requirements. This is fixed by changing to `pragma solidity ^0.8.25.0` - I think because the actual versions include patches.

## How this was tested
I confirmed the change does not let the contracts compile with solidity `v0.8.24`

## How is this documented

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/ava-labs/icm-services
 
-go 1.24.9
+go 1.24.11
 
 tool (
 	github.com/bufbuild/buf/cmd/buf


batconjurer:
> Why are we not switching pragmas to `0.8.30` directly? IMO there is benefit in a specific tag/commit being guaranteed to have been compiled/tested/audited against a specific version.

I would second this

PR #1112, opened 2025-12-04
Log TxIDs in destination_client

iansuvak:
## Why this should be merged

This will ensure that all logs related to sending, or waiting for a confirmation include the txID of the relevant transaction. 

## How this works

I didn't propagate the newly wrapped logger to `waitForReceipt` since that function returns via a channel and I wanted the log in `SendTx` which is the authoritative final log about the failure to also include the TxID. I'm open to changing it back if that's preferred. 

## How is this documented


vms/evm/destination_client.go
```diff
@@ -434,22 +433,26 @@ func (s *concurrentSigner) issueTransaction(
 	sendTxCtx, sendTxCtxCancel := context.WithTimeout(context.Background(), utils.DefaultRPCTimeout)
 	defer sendTxCtxCancel()
 
-	s.logger.Info(
+	log := s.logger.With(
+		zap.Stringer("txID", signedTx.Hash()))
```
geoff-vball: We might as well add the nonce and the gas values to this `logger.With` for the rest of the log calls in this function

PR #1120, opened 2025-12-08
Consolidate loggers

geoff-vball:
## Why this should be merged
Removes `libevm` and go `log` loggers. Creates a global version of the `avalanchego` logger.

## How is this documented


batconjurer (approved):

iansuvak (approved):
🙌

PR #1131, opened 2025-12-11
Validator set updater

ylg-avalabs:
## Why this should be merged
- Create validator set updater in icm-service to update the validator set to AvalancheValidatorSetRegistry contract
- Create external evm destination chain client to send the message to evm chain

TODO:
- Refactor to extract the common part of L1 and external evm destination client.  

## How is this documented

### DIFF
--- a/abi-bindings/go/AvalancheValidatorSetRegistry/AvalancheValidatorSetRegistry.go
+++ b/abi-bindings/go/AvalancheValidatorSetRegistry/AvalancheValidatorSetRegistry.go
@@ -558,6 +558,27 @@ func (_AvalancheValidatorSetRegistry *AvalancheValidatorSetRegistryTransactorSes
 	return _AvalancheValidatorSetRegistry.Contract.UpdateValidatorSet(&_AvalancheValidatorSetRegistry.TransactOpts, validatorSetID, message, validatorBytes)
 }
 
+// UpdateValidatorSetWithDiff is a paid mutator transaction binding the contract method.
+//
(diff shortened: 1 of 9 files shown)


ethereum/docker/Dockerfile.geth
```diff
@@ -0,0 +1,30 @@
...
+    apt-get install -y wget ca-certificates psmisc && \
+    rm -rf /var/lib/apt/lists/*
+
+# Set Geth version with commit hash
+# The "12b4131f" is the commit hash for the official v1.16.1 release of geth.
+# See: https://geth.ethereum.org/downloads/ and https://github.com/ethereum/go-ethereum/releases/tag/v1.16.1
+ARG GETH_VERSION=v1.16.1-12b4131f
+
+# Download and install geth
+RUN wget https://gethstore.blob.core.windows.net/builds/geth-linux-amd64-${GETH_VERSION#v}.tar.gz && \
```
geoff-vball: What does the `#v` in the variables do?

icm-contracts/tests/flows/services/validator_set_updater.go
```diff
@@ -0,0 +1,401 @@
...
+						// - Higher P-chain height
+						updatedValidatorSet, err := registry.GetValidatorSet(callOpts, big.NewInt(0))
+						if err != nil {
+							log.Warn("Failed to query registry for update check", zap.Error(err))
+							continue
+						}
+
+						// Check if the validator set has been updated with more validators
+						if len(updatedValidatorSet.Validators) > firstValidatorCount &&
+							updatedValidatorSet.PChainHeight > firstPChainHeight {
```
geoff-vball: Same thing here as above. If we hit this condition we should break from the loop and execute the rest of the code below once, so all this doesn't have to be nested.

PR #1137, opened 2025-12-15
Add warp interface to teleporter

batconjurer:
## Why this should be merged
In order to re-use Teleporter contracts on external EVM chains, these contracts should not rely on having access to the Warp pre-compile. This PR replaces the hard-coded pre-compile address with an interface. This interface also extends the methods available to the Teleporter contracts. In particular, Teleporter contracts can no receive ICM messages directly rather than looking them up from storage slots.
## How is this documented


icm-contracts/avalanche/teleporter/TeleporterMessenger.sol
```diff
@@ -108,6 +112,16 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
             => mapping(address feeTokenContract => uint256 redeemableRewardAmount)
     ) internal _relayerRewardAmounts;
 
+    /*
+     * @notice Register the address of the contract to be used for sending/receiving
+     * Warp messages.
+     * @dev This function is a delayed constructor to allow using Nick's method
+     * to deploy this contract
+     */
+    function initialize(address warpContract) external initializer {
```
michaelkaplan13:
```
     * @dev This function is a delayed constructor to allow using Nick's method
     * to deploy this contract
```

I think this comment/rationale is incorrect. Nick's method txs can still provide constructor arguments for contract deployments.

icm-contracts/avalanche/teleporter/TeleporterMessenger.sol
michaelkaplan13: Taking a step back, I don't think this is the correct approach for trying to extend `TeleporterMessenger` to work on external chains. 

There aren't any contracts or precompile that implement both `IWarpMessenger` and the extension added here in `IWarpExt`, which means that `TeleporterMessenger` would need to know which function of `IWarpExt` to call based on what implementation of that interface it is provided, which completely breaks that abstraction.

I haven't fully thought it through, but think it would be better to have structure such as:
```
IWarpMessenger -> interface the warp precompile implements
IExternalMessenger -> interface external messaging contracts implement for sending/verifying
ITeleporterMessenger -> (modified) interface definition for the Teleporter protocol (i.e. message format and processing. This interface will now not include `receiveCrossChainMessage`, since it is not applicable for all chains.
IAvalancheTeleporterMessenger -> for sending & receiving within Avalanche chains
IExternalTeleporterMessenger -> for sending & receiving to/from external chains
```

Lets discuss on the next planning call

batconjurer: While we can certainly discuss this further, I do want to clarify that `NativeWarp.sol` does implement `IWarpExt` and thus `IWarpMessenger`. The current aspect which I think can be improved is relayers currently need to know which TeleporterMessenger method to call depending on the context. This could probably be abstracted by making the bulk of the TeleportMessenger logic an interface and overriding the `receiveCrossChainMessage` method.

batconjurer: Which I'm now realizing is your point.

PR #1138, opened 2025-12-16
Consolidate Teleporter setup in e2e tests

geoff-vball:
## How is this documented


batconjurer (approved):
This PR is fine for now, but we still need to figure out how we will test changes to teleporter contracts in the future.

PR #1151, opened 2026-01-06
First draft TV2

geoff-vball:
## How is this documented


icm-contracts/avalanche/teleporter/TeleporterMessenger.sol
```diff
@@ -108,6 +107,14 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
             => mapping(address feeTokenContract => uint256 redeemableRewardAmount)
     ) internal _relayerRewardAmounts;
 
+    constructor(
+        address verifierSender
+    ) {
+        blockchainID = bytes32(0); // TODO decide how to set this for external chains
```
mdelle1: This is the destination blockchain ID right? If so should this be renamed to indicate that?

icm-contracts/avalanche/teleporter/TeleporterMessenger.sol
```diff
@@ -239,41 +244,23 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
         require(
-            teleporterMessage.destinationBlockchainID == blockchainID_,
+            teleporterMessage.destinationBlockchainID == blockchainID,
             "TeleporterMessenger: invalid destination chain ID"
         );
 
         // Calculate the message ID of the message given the source blockchain ID and message nonce.
         bytes32 messageID = calculateMessageID(
-            warpMessage.sourceChainID, blockchainID_, teleporterMessage.messageNonce
+            message.sourceBlockchainID, blockchainID, teleporterMessage.messageNonce
```
mdelle1: Would it make sense for `calculateMessageID` to also take in the raw message, and then use the message hash to determine the messageID?
Otherwise, different instances of the Teleporter contract from the same source to destination blockchain will have colliding message IDs for entirely different messages. Using the message hash incorporates global uniqueness of messages across different Teleporter instances (from the same source to destination chains).

icm-contracts/avalanche/teleporter/TeleporterMessenger.sol
mdelle1: But on [this](https://github.com/ava-labs/icm-services/pull/1151/files#diff-3dd91ce12c3028dbbe7d1983fdc277a61a6b9f82c2acd91bb51723995f043c25R257) line we assert `teleporterMessage.destinationBlockchainID == blockchainID` right?

icm-contracts/avalanche/teleporter/ITeleporterMessenger.sol
```diff
@@ -12,6 +12,15 @@ struct TeleporterMessageReceipt {
     address relayerRewardAddress;
 }
 
+struct ICMMessage {
+	TeleporterMessage unsignedMessage;
```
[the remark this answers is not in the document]
geoff-vball: Yeah, good point

PR #1156, opened 2026-01-13
Add `SimpleSerialize` utility library

mdelle1:
This PR introduces a lightweight library for verifying SSZ (Simple Serialize) Merkle Proofs, which allows traversing the SSZ binary tree structure used by the Ethereum Consensus Layer. This logic is a prerequisite for the `ZKVerifier` implementation, enabling the contract to verify Beacon State Roots against Beacon Block Roots. 

This PR is based off of the SSZ library implementation by Succinct Labs which can be found [here](https://github.com/succinctlabs/telepathy-contracts/blob/main/src/libraries/SimpleSerialize.sol) and uses their provided test vector.


batconjurer (commented):
Some small nits. I find the name Simple Serialize counter intuitive. Also, I think this contract makes more sense in under `common` rather than `avalanche`

geoff-vball (commented):
Needs one small tweak I think, but otherwise this PR lgtm

PR #1157, opened 2026-01-14
Replace subnet-evm clients with libevm clients

### LINKED ISSUE #1073: Allow use of latest lib-evm for abi-gen
**Context and scope**

Currently the abigen script is hardcoded to use an old version of subnet-evm because abigen was removed from it in favor of uplifted one to abigen. The code produced by libevm abigen doesn't work right now due to lacking methods. These should be added to the libevm and hardcoding should be undone to use the latest libevm version of the abigen. 

**Open questions**

iansuvak:
## Why this should be merged

Closes #1073

## How this works

In order to replace the underlying client library the interfaces used are split across the 3 main use-cases:

- Source clients, subscribing to and listening to new messages
- Destination clients delivering the messages
- Full contract deployment used in E2E testing flows

The first 3 commits are just changing the scripts, abi-bindings and files that only had import lines changed while the commits 4 and 5 contain functional changes and testing changes respectively. 

## How this was tested

E2E and unit testing

## How is this documented

### DIFF
--- a/relayer/listener.go
+++ b/relayer/listener.go
@@ -32,11 +32,11 @@ type Listener struct {
 	currentRequestID   uint32
 	logger             logging.Logger
 	sourceBlockchain   config.SourceBlockchain
-	catchUpResultChan  chan bool
 	healthStatus       *atomic.Bool
 	ethClient          *ethclient.Client
 	messageCoordinator *MessageCoordinator
 	maxConcurrentMsg   uint64
+	errChan            chan error
 }
 
 // RunListener creates a Listener instance and the ApplicationRelayers for a subnet.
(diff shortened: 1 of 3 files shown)


vms/evm/ava_destination_client.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/ava-labs/libevm/ethclient"
+	"github.com/ava-labs/libevm/rpc"
+)
+
+var _ DestinationRPCClient = (*avaDestinationClient)(nil)
+
+// avaDestinationClient wraps libevm's ethclient.Client and implements DestinationRPCClient.
+// It delegates to the underlying ethclient.Client for methods that are available,
+// and uses the rpcClient for methods that are avalanche specific.
+type avaDestinationClient struct {
```
iansuvak: I dislike this name and am accepting suggestions. The only function here that actually needs rpcClient and raw fetching is `EstimateBaseFee` which is possibly replicatable by just fetching the latest block and getting fee from there in eth semantics but for now I wanted to keep the functionality the same.

geoff-vball (approved):

PR #1159, opened 2026-01-15
Fix subscriber reconnections to not miss blocks

geoff-vball:
## Why this should be merged
Closes https://github.com/ava-labs/icm-services/issues/1155

## How this works
Fires off a goroutine to catch up on blocks on startup and whenever the subscriber reconnects.

## How is this documented

### DIFF
--- a/relayer/listener.go
+++ b/relayer/listener.go
@@ -28,15 +28,15 @@ const (
 
 // Listener handles all messages sent from a given source chain
 type Listener struct {
-	Subscriber         *evm.Subscriber
-	currentRequestID   uint32
-	logger             logging.Logger
-	sourceBlockchain   config.SourceBlockchain
-	healthStatus       *atomic.Bool
-	ethClient          *ethclient.Client
-	messageCoordinator *MessageCoordinator
-	maxConcurrentMsg   uint64
-	errChan            chan error
+	Subscriber                   *evm.Subscriber
+	currentRequestID             uint32
+	logger                       logging.Logger
+	sourceBlockchain             config.SourceBlockchain
+	healthStatus                 *atomic.Bool
+	ethClient                    *ethclient.Client
+	messageCoordinator           *MessageCoordinator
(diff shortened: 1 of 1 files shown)


relayer/listener.go
```diff
@@ -134,53 +128,41 @@ func newListener(
...
-
-			// Mark the relayer as unhealthy if the catch-up process fails or if the catch-up channel is unexpectedly closed.
-			if !ok {
-				lstnr.healthStatus.Store(false)
-				lstnr.logger.Error("Catch-up channel unexpectedly closed. Exiting listener goroutine.")
-				return fmt.Errorf("catch-up channel unexpectedly closed")
+			lstnr.logger.Error("Received error", zap.Error(err))
+		case icmBlockInfo := <-lstnr.Subscriber.ICMBlocks():
+			// If this is the first subscriber block either or startup or after a reconnect.
+			if catchingUp {
```
iansuvak: we aren't distinguishing here whether the block that we've just received has come from subscription or from a catchup process. Can we handle this by adding that info to the `icmBlockInfo`?

iansuvak: Also what happens here if we have multiple catchups running simultaneously? Before the mental model was simpler because we only ever had a single catchup goroutine.

icm-contracts/tests/utils/services_utils.go
```diff
@@ -532,6 +532,8 @@ func TriggerProcessMissedBlocks(
 	defer startupCancel()
 	WaitForChannelClose(startupCtx, readyChan)
 
+	// Send a native transfer to trigger block production
+	SendNativeTransfer(ctx, sourceL1Info, fundedKey, fundedAddress, big.NewInt(1))
```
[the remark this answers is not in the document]
iansuvak: I don't know if I would call it cleaner but there is `IssueTxsToAdvanceChain` in `utils/chain.go` in my #1157 PR.

PR #1175, opened 2026-01-29
Validator set registry

batconjurer:
## Why this should be merged
Adds a validator set registry contract and tests. The actual registry is split into two layers: a logical and data layer. This logic layer handles the sharding of validator set updates. The data layer stores, parses, and applies raw data. This current data layer implementation uses naive sharding, but this can be swapped out for something smarter, like diffs.
## How is this documented


icm-contracts/ethereum/interfaces/IAvalancheValidatorSetManager.sol
```diff
@@ -0,0 +1,96 @@
...
+     * @notice Set a partial validator set as most recently registered
+     */
+    function setPartialValidatorSet(
+        bytes32 avalancheBlockchainID,
+        PartialValidatorSet memory partialValidatorSet
+    ) external;
+
+    /**
+     * @notice  Validate and apply a shard to a partial validator set. If the set is completed by this shard, copy
+     * it over to the `_validatorSets` mapping.
```
iansuvak: This interface doc comment references an implementation variable. It should be kept general

foundry.toml
```diff
@@ -18,6 +18,7 @@ evm_version = 'shanghai'
 src = 'icm-contracts/ethereum/'
 solc_version = '0.8.30'
 evm_version = 'prague'
+via_ir = true
```
geoff-vball: Just curious why you applied this here

icm-contracts/ethereum/AvalancheValidatorSetManager.sol
```diff
@@ -0,0 +1,156 @@
...
+    // Used to allow updating validator sets across multiple transactions
+    mapping(bytes32 => PartialValidatorSet) internal _partialValidatorSets;
+    /**
+     * @notice Set the **complete** validator set most recently registered
+     */
+
+    function setValidatorSet(
+        bytes32 avalancheBlockchainID,
+        ValidatorSet memory validatorSet
+    ) external {
```
mdelle1: I'm not sure I understand the intent here -- any attacker can just arbitrarily update the validator set by calling this function...

icm-contracts/ethereum/AvalancheValidatorSetManager.sol
```diff
@@ -0,0 +1,156 @@
...
+     * @dev Applies a set of validators to partial to a set that has been registered.
+     */
+    function applyShard(ValidatorSetShard calldata shard, bytes memory shardBytes) external {
+        bytes32 avalancheBlockchainID = shard.avalancheBlockchainID;
+        // Parse the validators.
+        (Validator[] memory validators, uint64 validatorWeight) =
+            ValidatorSets.parseValidators(shardBytes);
+        require(validators.length > 0, "Validator set cannot be empty");
+        require(validatorWeight > 0, "Total weight must be greater than 0");
+
```
mdelle1: Doesn't the logic below break if the shards arrive out of order? For example, if there are 15 validators and we have shards of length 5, if the second shard `[val_5, ..., val_9] `arrives first, then it would insert these into the first 5 indices of `_partialValidatorSets[avalancheBlockchainID].validators` which is incorrect.

mdelle1: Edit: I see that you make the call `require(shardsReceived + 1 == shard.shardNumber, "Received shard out of order");` in `updateValidatorSet` -- I still think a check of this nature should be placed inside `applyShard` since critically this function is `external` and calling it out of order would break the logic of the contract.

icm-contracts/ethereum/tests/AvalancheValidatorSetRegistry.t.sol
```diff
@@ -3,17 +3,563 @@ pragma solidity ^0.8.30;
...
+contract AvalancheValidatorSetRegistryInitialization is AvalancheValidatorSetRegistryCommon {
     AvalancheValidatorSetRegistry private _registry;
+    AvalancheValidatorSetManager private _manager;
 
     function setUp() public {
-        _registry = new AvalancheValidatorSetRegistry(_NETWORK_ID);
+        (ValidatorSet memory validatorSet, bytes32 validatorSetHash) = dummyPChainValidatorSet();
+        bytes32[] memory shardHashes = new bytes32[](1);
+        shardHashes[0] = validatorSetHash;
+        ValidatorSetMetadata memory initialValidatorSetData = ValidatorSetMetadata({
```
mdelle1:
```suggestion
        ValidatorSetMetadata memory initialValidatorSetMetaData = ValidatorSetMetadata({
```

icm-contracts/ethereum/AvalancheValidatorSetManager.sol
batconjurer: good catch, I forgot to add a check there

icm-contracts/ethereum/AvalancheValidatorSetManager.sol
```diff
@@ -0,0 +1,156 @@
...
+        // update the partial validator set
+        uint256 offset = _partialValidatorSets[avalancheBlockchainID].numValidators;
+        for (uint256 i = 0; i < validators.length; i++) {
+            _partialValidatorSets[avalancheBlockchainID].validators[offset + i] = validators[i];
```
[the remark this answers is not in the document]
batconjurer: I'm not sure that is correct. These vectors are pre-allocated when `register` gets called. My understanding is that Solidity actually populates the array with 0 versions of the struct rather than simply reserving space. As such pushing would add new validators after all the 0 entries instead of overwriting the reserved slot. We do the pre-allocation to put the gas cost of reserving storage slots onto the register call instead of any update call.

icm-contracts/ethereum/AvalancheValidatorSetManager.sol
batconjurer: So the registry is meant to be the logical layer, which is why the check is there. The manager is just meant to store data and the only validation is does is checking the data is well formed. This function should not be called by any contract other than the registry, although I had forgotten to enforce this (you caught this above). With the ownership pattern in place, this should be secure now and the separation of responsibilities shouldn't be an issue

icm-contracts/ethereum/interfaces/IAvalancheValidatorSetManager.sol
```diff
@@ -0,0 +1,107 @@
...
+} from "../utils/ValidatorSets.sol";
+
+/**
+ * @title IAvalancheValidatorManager
+ * @notice Interface for an Avalanche Validator Manager contract
+ * @dev This interface defines the data layer for managing the validator sets and updating them
+ */
+interface IAvalancheValidatorSetManager {
+    /**
+     * @notice This contract control data that is owned by another contract. We
```
geoff-vball:
```suggestion
     * @notice This contract controls data that is owned by another contract. We
```

icm-contracts/ethereum/AvalancheValidatorSetManager.sol
```diff
@@ -0,0 +1,188 @@
...
+     */
+    function initialize(
+        address owner
+    ) external initializer {
+        _owner = owner;
+    }
+
+    /**
+     * @notice Set the **complete** validator set most recently registered
+     */
```
geoff-vball: If these docs are defined in the interface file, we should just point there, instead of having them duplicated since they might accidentally diverge.

icm-contracts/ethereum/AvalancheValidatorSetRegistry.sol
```diff
@@ -2,21 +2,275 @@
...
+            IAvalancheValidatorSetManager(validatorSetManagerContract).setPartialValidatorSet(
+                avalancheBlockchainID, partialValidatorSet
+            );
+            if (!isRegistered(message.message.sourceBlockchainID)) {
+                // pre-allocate storage for the completed validator set
+                ValidatorSet memory validatorSet = ValidatorSet({
+                    avalancheBlockchainID: avalancheBlockchainID,
+                    pChainHeight: validatorSetMetadata.pChainHeight,
+                    pChainTimestamp: validatorSetMetadata.pChainTimestamp,
+                    validators: new Validator[](validatorSetMetadata.totalValidators),
```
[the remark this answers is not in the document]
batconjurer: Indeed I am.  In general, every validator set needs two allocations. One for completed sets and one were we are buffering updates until we have a new completed set. I'm warming up all those storage slots at once.

foundry.toml
batconjurer: I had "stack too deep" issues when compiling. I tried removing local variables, but after doing that and still having the error several times, it got tiresome.

icm-contracts/ethereum/AvalancheValidatorSetManager.sol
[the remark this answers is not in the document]
mdelle1: The gas of pre-allocating the validator set [here](https://github.com/ava-labs/icm-services/pull/1175/changes#diff-0fd0cbcc47619e1707fb73cbfa008330485145e3748f38ba6ca2710ef040d6ddR130)

PR #1188, opened 2026-02-17
Update go to 1.25.7

geoff-vball:
## How is this documented


iansuvak:
Actually should be able to remove gomaxprocs stuff as well but can do it in a separate PR as well.

PR #1211, opened 2026-03-02
Consolidate `AvalancheValidatorSetRegistry` Unit Tests 

mdelle1:
Progress towards issue https://github.com/ava-labs/icm-services/issues/1201 . The PR is not yet completed, but I wanted to get opinions before I proceed in this fashion with the 2 remaining contracts. 

The greater goal is that if we add more implementations of the registry contract, it shouldn't be cumbersome to update the unit tests for this new implementation.


icm-contracts/ethereum/AvalancheValidatorSetRegistry.sol
```diff
@@ -122,11 +122,8 @@ contract AvalancheValidatorSetRegistry is IAvalancheValidatorSetRegistry {
             );
 
             if (!isRegistered(message.sourceBlockchainID)) {
```
mdelle1: I removed these lines because I don't think that the stored height and timestamp should immediately be updated, rather they should be updated after all shards have been applied, so this update should occur in the `applyShard` implementation

PR #1212, opened 2026-03-02
Remove `ByteSlicer` for gas-optimized unpacking in `ValidatorSets.sol`

mdelle1:
Addresses the issue https://github.com/ava-labs/icm-services/issues/1190 .


geoff-vball (commented):
Did you do any gas testing on these changes? Intuitively to me, I would think that XORing the bytes n times for an n-sized array would be less efficient than allocating an addition array in memory like the byte slicer was doing. 

I'm pretty sure the only way to make this operation more efficient is to have the `bytes` passed into the function be `calldata` instead of `memory`, in which case we can use the built-in array slicing syntax `x[start:end]`. When/where you can use calldata can be tricky, but I feel like we should be able to make it work here.

batconjurer (changes requested):
This seems to mostly inline the ByteSlicer library calls. Are we sure this would give a lot of gas improvement? If we can, we should try to use assembly to get rid of the loops over the individual bytes. At the very least, we should turn off the checks on incrementing the loop counters

PR #1246, opened 2026-03-24
Doc Update: 24.03.2026

batconjurer:
## Why this should be merged
An update to the docs to keep them in sync with the actual implementation. Most notable is that the decision to use Nick's method all the way down is now presented as the canonical solution.
## How is this documented


docs/external-interop/icm_message_authentication.md
```diff
@@ -1,28 +1,18 @@
...
+ `TeleporterMessengerV2` will be deployed via Nick's method; its constructor taking the address of a contract implementing the  `IAdapter` interface. Note that this requires the `IAdapter` contract to also be deployed via Nick's method. To support multiple verification schemes, multiple `TeleporterMessengerV2` contracts may be deployed. To summarize:
+ * The same implementations of an `IAdapter` contract are deployed to the same address on each chain.
+ * All `TeleporterV2` contracts implementing the same scheme are deployed to the same address on each chain.
 
-The first option is that a `TeleporterV2` contract has both `IMessageVerifier` and `IMessageSender` implementations hard-coded as part of its state. To support multiple verification schemes, multiple `TeleporterV2` contracts may be deployed.  
+Two `TeleporterV2` contracts may talk to one another if they use the same `IAdapter` contract. When a TeleporterV2 contract receives a message, it checks if it originated from the same contract address as its own. If so, it knows that it is part of the same scheme and can verify the message using the hard-coded `IAdapater` implementation.
```
geoff-vball:
```suggestion
Two `TeleporterV2` contracts may talk to one another if they are deployed at the same address. This means that they have been deployed using `Nick's Method` and use the same `IAdapter` contract. When a TeleporterV2 contract receives a message, it checks if it originated from the same contract address as its own. If so, it knows that it is part of the same scheme and can verify the message using the hard-coded `IAdapater` implementation.
```

PR #1272, opened 2026-04-13
1 listener per protocol address

geoff-vball:
## Why this should be merged
Creates 1 listener/subscriber per protocol address, and the listener will filter events based on the protocol address.

## How is this documented

### DIFF
--- a/relayer/main/main.go
+++ b/relayer/main/main.go
@@ -327,25 +327,32 @@ func main() {
 
 	// Create listeners for each of the subnets configured as a source
 	for _, sourceBlockchain := range cfg.SourceBlockchains {
-		// errgroup will cancel the context when the first goroutine returns an error
-		errGroup.Go(func() error {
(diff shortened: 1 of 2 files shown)


relayer/main/main.go
```diff
@@ -327,25 +327,23 @@ func main() {
...
-				minHeights[sourceBlockchain.GetBlockchainID()],
-				messageCoordinator,
-				cfg.MaxConcurrentMessages,
-			)
-			if err != nil {
-				log.Error("error running listener", zap.Error(err))
-			}
-			return err
-		})
+		for _, protocolAddr := range sourceBlockchain.ProtocolAddresses() {
```
mdelle1, line 330: Why is the scoped logger `logger.With(...` dropped?

geoff-vball: We actually call `logger.With()` right inside of `RunListener` with more parameters

mdelle1 (approved):

batconjurer (approved):

iansuvak (approved):

PR #1273, opened 2026-04-14
Implement EventFilterForProtocol

geoff-vball:
## Why this should be merged
Passes the Protocol type to the application relayer/listener so that it can decide how to listen to events. The protocol will also be used to parse/dispatch messages separately for TeleporterV2 in a followup PR

## How is this documented

### DIFF
--- a/relayer/config/types.go
+++ b/relayer/config/types.go
@@ -3,6 +3,14 @@
 
 package config
 
+import "github.com/ava-labs/libevm/common"
+
+// Protocol pairs a contract address with the message protocol it implements.
+type Protocol struct {
+	Address  common.Address
+	Protocol MessageProtocol
+}
+
 // Supported Message Protocols
 type MessageProtocol int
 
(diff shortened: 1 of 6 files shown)


relayer/config/types.go
```diff
@@ -3,6 +3,14 @@
 
 package config
 
+import "github.com/ava-labs/libevm/common"
+
+// Protocol pairs a contract address with the message protocol it implements.
+type Protocol struct {
```
mdelle1, line 9: I'd suggest to avoid making the `Protocol` struct name the same as the field `Protocol`. Perhaps renaming the `Protocol` field to `Type`, or something similar, e.g., `Protocol.Type`.

relayer/listener.go
```diff
@@ -28,6 +28,25 @@ const (
 	retryResubscribeTimeout = 10 * time.Second
 )
 
+// EventFilterForProtocol returns the ethereum log filter topics for the given protocol and
+// contract address. Returns nil for protocols that do not use a listener.
+func EventFilterForProtocol(
```
mdelle1: Shouldn't this just take in the entire Protocol struct as a parameter instead of the 2 fields separately?

mdelle1 (dismissed):

batconjurer (dismissed):

iansuvak (dismissed):

batconjurer (approved):

PR #1278, opened 2026-04-17
Make packing more generic

### LINKED ISSUE #1295: IAdapters should be able to serialize/deserialize their own attestations
**Context and scope**
Currently `abi-bindings/go/TeleporterMessengerV2/packing.go` is hardcoded to pack the attestation for the WarpAdapter. The packing method should take an input attestation, and we should have different implementations for each IAdapter to pack their own attestation type.

geoff-vball:
## Why this should be merged
Closes https://github.com/ava-labs/icm-services/issues/1295

Makes `PackReceiveCrossChainMessageV2` changeable based on what the Adapter contract is. This is necessary because different adapter contracts will need their `attestation`s packed differently.

## How is this documented


abi-bindings/go/TeleporterMessengerV2/packing.go
```diff
@@ -77,35 +86,21 @@ func PackRetryMessageExecution(sourceBlockchainID ids.ID, message TeleporterMess
...
 	return tabi.Pack("receiveCrossChainMessage", TeleporterICMMessage{
 		Message:            teleporterMessage,
 		SourceBlockchainID: sourceBlockChainID,
 		Attestation:        attestation,
 	}, relayerRewardAddress)
 }
 
-// PackReceiveCrossChainMessage packs a ReceiveCrossChainMessageInput to form
-// a call to the receiveCrossChainMessage function
-func PackReceiveCrossChainMessage(messageIndex uint32, relayerRewardAddress common.Address) ([]byte, error) {
```
geoff-vball, line 100: We have `PackReceiveCrossChainMessageV2` above, we don't need this.

icm-contracts/tests/utils/teleporter.go
```diff
@@ -111,16 +113,31 @@ func (t TeleporterTestInfo) TeleporterRegistryAddress(blockchainID ids.ID) commo
...
 }
 
 func (t TeleporterTestInfo) SetTeleporterV2(address common.Address, blockchainID ids.ID) {
 	info := t[blockchainID]
 	info.teleporterMessengerAddress = address
 	info.teleporterProtocol = TELEPORTER_V2
+	info.packReceiveCrossChainMessage = unimplementedPackReceiveCrossChainMessage
+}
+
+func (t TeleporterTestInfo) SetTeleporterV2WarpAdapter(address common.Address, blockchainID ids.ID) {
```
batconjurer, line 136: nit: Since this is supposed to add packing support for all IAdapters, should the name actually include `Warp`?

geoff-vball: So, this particular function sets the `packReceiveCrossChainMessage` to the one that is specific to the `WarpAdapter`. For any distinct `IAdapter`, we need a different `packReceiveCrossChainMessage` because the `attestation` will be different.

icm-contracts/tests/utils/teleporter.go
```diff
@@ -194,17 +211,14 @@ func (t TeleporterTestInfo) PackReceiveCrossChainMessage(
...
-	switch t.TeleporterProtocol(destination.BlockchainID) {
-	case TELEPORTER:
-		return t.PackReceiveCrossChainMessageV1(signedMessage, senderKey, destination)
-	case TELEPORTER_V2:
-		return t.PackReceiveCrossChainMessageV2(signedMessage, senderKey, destination)
-	default:
-		panic("unsupported teleporter protocol")
+	fn := t[destination.BlockchainID].packReceiveCrossChainMessage
+	if fn == nil {
+		panic("no PackReceiveCrossChainMessage function set for this chain")
```
mdelle1: Should we add the chain ID to this error log?

icm-contracts/tests/utils/teleporter.go
iansuvak: Here in test files it is fine since it's explicitly setting the `PackReceiveChainMessageWarpAdapter`

iansuvak (approved):

icm-contracts/tests/utils/teleporter.go
```diff
@@ -111,16 +113,31 @@ func (t TeleporterTestInfo) TeleporterRegistryAddress(blockchainID ids.ID) commo
...
 	info.teleporterMessengerAddress = address
 	info.teleporterProtocol = TELEPORTER
+	info.packReceiveCrossChainMessage = PackReceiveCrossChainMessageV1
 }
 
 func (t TeleporterTestInfo) SetTeleporterV2(address common.Address, blockchainID ids.ID) {
 	info := t[blockchainID]
 	info.teleporterMessengerAddress = address
 	info.teleporterProtocol = TELEPORTER_V2
+	info.packReceiveCrossChainMessage = unimplementedPackReceiveCrossChainMessage
```
mdelle1, line 133: The function `SetTeleporterV2` will automatically panic at runtime. I think we should either remove it and force callers to pick a specific adapter, or perhaps comment it out and leave an explicit TODO?

mdelle1:
Minor comments but otherwise LGTM

batconjurer (approved):

icm-contracts/tests/utils/teleporter.go
geoff-vball: `SetTeleporterV2` won't panic at runtime. If we call `SetTeleporterV2` and then someone calls `packReceiveCrossChainMessage` it will panic, which is the point. `SetTeleporterV2` is only used in `avalanche_validator_set_registry.go` which does not call `packReceiveCrossChainMessage`. I think we build the `ReceiveCrossChainMessage` manually. We could probably move that implementation to here if we wanted and make it `SetTeleporterV2EcdsaVerifier`

icm-contracts/tests/utils/teleporter.go
geoff-vball: added

mdelle1 (approved):

PR #1281, opened 2026-04-30
Add nightly integration test for Boundless ZK proof indexing

mdelle1:
Adds a nightly integration test that sends a real message on Sepolia via an `ECDSAVerifier` contract, then polls a Boundless subgraph until a ZK consensus proof covering the transaction is available. All Node.js scripts and npm dependencies run inside a Docker containers to mitigate supply-chain risk. The test runs daily via github actions and is the first step in a full end-to-end ZK verification flow in a follow-up PR.

The `ECDSAVerifier` event emission will be swapped out with `DiffUpdater` once `sendMessage` is implemented for this contract.


scripts/tools/fixture-gen/send_sepolia_message.mts
```diff
@@ -0,0 +1,102 @@
...
+const SENDER_CONTRACT = process.env.SENDER_CONTRACT!;
+
+if (!ETH_RPC_URL || !SENDER_PRIVATE_KEY || !SENDER_CONTRACT) {
+  console.error("Required env vars: ETH_RPC_URL, SENDER_PRIVATE_KEY, SENDER_CONTRACT");
+  process.exit(1);
+}
+
+// From ABI Go bindings for ECDSAVerifier
+// TODO: Swap out with DiffUpdater ABI once sendMessage implementation is complete. Issue: https://github.com/ava-labs/icm-services/issues/1282
+const ECDSA_VERIFIER_ABI = [
```
batconjurer, line 34: is there not a way to get this from the generated abi bindings directly? If that contract every gets updated, I can totally see myself forgetting to update this.

batconjurer (approved):

scripts/tools/fixture-gen/send_sepolia_message.mts
mdelle1: Good point, but changing the contract would require re-deploying the contract manually on Sepolia anyway, at which point the ABI and script would both need updating

scripts/tools/fixture-gen/send_sepolia_message.mts
```diff
@@ -0,0 +1,102 @@
...
+
+  // Find the ECDSAVerifierSendMessage event log index
+  const logIndex = receipt.logs.findIndex(
+    (log: ethers.providers.Log) =>
+      log.address.toLowerCase() === SENDER_CONTRACT.toLowerCase() &&
+      log.topics[0] === EVENT_TOPIC
+  );
+
+  if (logIndex === -1) {
+    throw new Error("ECDSAVerifierSendMessage event not found in receipt");
```
geoff-vball:
```suggestion
    throw new Error("ECDSAVerifierSendMessage event not found in receipt. This will fail if EVENT_TOPIC has changed");
```

scripts/tools/fixture-gen/send_sepolia_message.mts
```diff
@@ -0,0 +1,102 @@
...
+if (!ETH_RPC_URL || !SENDER_PRIVATE_KEY || !SENDER_CONTRACT) {
+  console.error("Required env vars: ETH_RPC_URL, SENDER_PRIVATE_KEY, SENDER_CONTRACT");
+  process.exit(1);
+}
+
+// From ABI Go bindings for ECDSAVerifier
+// TODO: Swap out with DiffUpdater ABI once sendMessage implementation is complete. Issue: https://github.com/ava-labs/icm-services/issues/1282
+const ECDSA_VERIFIER_ABI = [
+  "function sendMessage(tuple(uint256 messageNonce, address originSenderAddress, address originTeleporterAddress, bytes32 destinationBlockchainID, address destinationAddress, uint256 requiredGasLimit, address[] allowedRelayerAddresses, tuple(uint256 receivedMessageNonce, address relayerRewardAddress)[] receipts, bytes message) message) external",
+  "event ECDSAVerifierSendMessage(tuple(uint256 messageNonce, address originSenderAddress, address originTeleporterAddress, bytes32 destinationBlockchainID, address destinationAddress, uint256 requiredGasLimit, address[] allowedRelayerAddresses, tuple(uint256 receivedMessageNonce, address relayerRewardAddress)[] receipts, bytes message) message)"
```
geoff-vball: We don't use this

scripts/tools/fixture-gen/poll_boundless_proofs.mts
```diff
@@ -0,0 +1,167 @@
...
+
+// How long to wait for a proof before giving up (3 hours)
+const MAX_WAIT_MS = 180 * 60 * 1000; 
+// How often to poll the subgraph (1 minute)
+const POLL_INTERVAL_MS = 60 * 1000;
+
+if (!SUBGRAPH_URL || !ETH_RPC_URL) {
+  console.error("Required env vars: SUBGRAPH_URL, ETH_RPC_URL");
+  process.exit(1);
+}
```
geoff-vball: I'm pretty sure these cases are covered by the `!`s when we get the vars from the env on lines 22 and 23

scripts/tools/fixture-gen/send_sepolia_message.mts
```diff
@@ -0,0 +1,102 @@
...
+const __dirname = path.dirname(__filename);
+
+const ETH_RPC_URL = process.env.ETH_RPC_URL!;
+const SENDER_PRIVATE_KEY = process.env.SENDER_PRIVATE_KEY!;
+const SENDER_CONTRACT = process.env.SENDER_CONTRACT!;
+
+if (!ETH_RPC_URL || !SENDER_PRIVATE_KEY || !SENDER_CONTRACT) {
+  console.error("Required env vars: ETH_RPC_URL, SENDER_PRIVATE_KEY, SENDER_CONTRACT");
+  process.exit(1);
+}
```
geoff-vball, line 29: These should also already be checked

scripts/tools/fixture-gen/send_sepolia_message.mts
geoff-vball: If we change the contract and this needs to be updated, we should leave a comment on the part of the code that will fail that this should probably be updated.

batconjurer (approved):

geoff-vball (approved):

PR #1291, opened 2026-05-05
Add reset logic to diff updater contract

ylg-avalabs:
## How is this documented


icm-contracts/ethereum/AvalancheValidatorSetRegistry.sol
```diff
@@ -129,21 +129,46 @@ contract AvalancheValidatorSetRegistry is IAvalancheValidatorSetRegistry, IAdapt
...
         }
 
-        // First registration for this payload chain: P-chain signs. Updates: that chain's validators sign.
-        if (!isRegistered(payloadBlockchainID)) {
+        // Signature routing:
+        //   - First registration (!isRegistered): signed by the P-chain
+        //     validator set initialized on this contract at deploy time.
+        //   - Any subsequent registration, including a "reset" that
+        //     wholesale replaces the registered set, is signed by the
+        //     registered L1 validator set. The contract cannot trust a
```
batconjurer: I don't think this portion of the comment is accurate. The P-chain validators are not static and definitely can be trusted; it is simply that the power to update a validator set here should belong to the L1, not the p-chain.

icm-contracts/ethereum/AvalancheValidatorSetRegistry.sol
```diff
@@ -129,21 +129,46 @@ contract AvalancheValidatorSetRegistry is IAvalancheValidatorSetRegistry, IAdapt
...
+        //     current L1 set and the registered set to achieve quorum.
+        bool wasAlreadyRegistered = isRegistered(payloadBlockchainID);
+        if (!wasAlreadyRegistered) {
             verifyICMMessage(message, pChainID);
         } else {
             verifyICMMessage(message, payloadBlockchainID);
         }
 
-        // Parse and validate the validator set payload
+        // Parse and validate the validator set payload. For DiffUpdater the
```
batconjurer: This is being extremely nit-picky, but in general I don't like parent contracts being aware of its children since that should be an abstraction barrier. I like the comment, but I would either make it generic to not refer to specific children, or split it up and move it the children directly.

icm-contracts/ethereum/AvalancheValidatorSetRegistry.sol
```diff
@@ -129,21 +129,46 @@ contract AvalancheValidatorSetRegistry is IAvalancheValidatorSetRegistry, IAdapt
...
+
         bytes32 avalancheBlockchainID = validatorSetMetadata.avalancheBlockchainID;
 
+        // Only treat the payload as a true "reset" event (emitting the
+        // dedicated event and instructing multi-shard processing to skip
+        // anchor checks) when the chain was already registered. For a
+        // fresh registration the prev=0 / prev=0 markers are just the
+        // natural encoding of "no prior state" and should emit the normal
+        // `ValidatorSetRegistered` event.
+        bool isActualReset = isReset && wasAlreadyRegistered;
```
batconjurer: The false positive on the `isReset` is an artifact of the current DiffUpdater implementation. In that function, we should change the `isReset` to additionally check if the chain was already registered. The parent contract should simply assume that the `isReset` variable is accurate.

icm-contracts/ethereum/DiffUpdater.sol
```diff
@@ -42,20 +54,24 @@ contract DiffUpdater is AvalancheValidatorSetRegistry {
...
-        require(
-            diff.previousTimestamp == currentValidatorSet.pChainTimestamp,
-            "Diff anchor timestamp mismatch"
-        );
-        require(diff.currentHeight > currentValidatorSet.pChainHeight, "P-Chain height too low");
-        require(
-            diff.currentTimestamp > currentValidatorSet.pChainTimestamp, "P-Chain timestamp too low"
-        );
+
+        if (!currentPartialValSet.isReset) {
```
batconjurer, line 63: I'm not sure I understand the reasoning for skipping these checks on reset. Same for the `parseValidatorSetMetadata` function

icm-contracts/ethereum/DiffUpdater.sol
```diff
@@ -106,13 +121,27 @@ contract DiffUpdater is AvalancheValidatorSetRegistry {
     }
 
     /**
-     * @dev The returned validator set if the result of applying the first diff to the currently registered
-     * set. The weight returned is the weight of the resulting set.
+     * @dev Returns the validator set that results from applying the first diff.
+     *
+     * The diff is interpreted as a reset when it carries `previousHeight == 0`
+     * and `previousTimestamp == 0`. A reset diff describes the entire current
```
batconjurer: Iterating my above comment, a reset here should be `previousHeight == 0` and `previousTimestamp == 0` and it was previously registered.

icm-contracts/ethereum/tests/AvalancheValidatorSetRegistry.t.sol
```diff
@@ -1211,6 +1212,176 @@ contract AvalancheValidatorSetRegistryPostInitialization is AvalancheValidatorSe
...
+        _registerAndApplyInitialL1Set();
+
+        (Validator[] memory newValidators, ICMMessage memory message, bytes memory shardBytes) =
+            _buildResetFixtureSingleShard(10, 10, 6);
+
+        assertTrue(_diffRegistry.isRegistered(L1_BLOCKCHAIN_ID));
+        assertFalse(_diffRegistry.isRegistrationInProgress(L1_BLOCKCHAIN_ID));
+
+        vm.expectEmit(true, false, false, false, address(_diffRegistry));
+        emit IAvalancheValidatorSetRegistry.ValidatorSetReset(L1_BLOCKCHAIN_ID);
```
batconjurer, line 1232: aren't you trying to assert that _diffRegistry.registerValidatorSet(message, shardBytes); emits this event? If so, won't this line mess that up?

icm-contracts/ethereum/tests/AvalancheValidatorSetRegistry.t.sol
```diff
@@ -1211,6 +1212,176 @@ contract AvalancheValidatorSetRegistryPostInitialization is AvalancheValidatorSe
...
+        // ValidatorSetReset event yet, finalization happens on the last
+        // shard).
+        _diffRegistry.registerValidatorSet(message, shards[0]);
+        assertTrue(_diffRegistry.isRegistrationInProgress(L1_BLOCKCHAIN_ID));
+
+        // Shard 2 finalizes the reset and must emit ValidatorSetReset.
+        ValidatorSetShard memory shard =
+            ValidatorSetShard({shardNumber: 2, avalancheBlockchainID: L1_BLOCKCHAIN_ID});
+        vm.expectEmit(true, false, false, false, address(_diffRegistry));
+        emit IAvalancheValidatorSetRegistry.ValidatorSetReset(L1_BLOCKCHAIN_ID);
```
batconjurer, line 1270: same question here

icm-contracts/ethereum/interfaces/IAvalancheValidatorSetRegistry.sol
```diff
@@ -17,6 +17,15 @@ interface IAvalancheValidatorSetRegistry {
 
     event ValidatorSetUpdated(bytes32 indexed avalancheBlockchainID);
 
+    /// @notice Emitted when a previously registered validator set is reset.
+    /// A reset occurs when the registered P-chain height for a chain has
+    /// advanced past the P-chain's minimum retained height so that signatures
+    /// anchored at the registered height can no longer be obtained. In that
+    /// case, the relayer submits a fresh P-chain-signed payload containing
```
batconjurer: Is it p-chain signed?  The rest of the code/comments seem that it is the L1 that signs it

batconjurer (changes requested):

icm-contracts/ethereum/interfaces/IAvalancheValidatorSetRegistry.sol
```diff
@@ -17,6 +17,15 @@ interface IAvalancheValidatorSetRegistry {
 
     event ValidatorSetUpdated(bytes32 indexed avalancheBlockchainID);
 
+    /// @notice Emitted when a previously registered validator set is reset.
+    /// A reset occurs when the registered P-chain height for a chain has
+    /// advanced past the P-chain's minimum retained height so that signatures
+    /// anchored at the registered height can no longer be obtained. In that
```
geoff-vball, line 23: Can this only trigger if more than 1/3 of the validator set has churned? What if 30% of the set has churned, but another 10% is just offline?

icm-contracts/ethereum/tests/AvalancheValidatorSetRegistry.t.sol
geoff-vball: This is just how `expectEmit` works. Here's the documentation for it

```
/// Prepare an expected log with (bool checkTopic1, bool checkTopic2, bool checkTopic3, bool checkData.).
/// Call this function, then emit an event, then call a function. Internally after the call, we check if
/// logs were emitted in the expected order with the expected topics and data (as specified by the booleans).
```

icm-contracts/ethereum/AvalancheValidatorSetRegistry.sol
geoff-vball: To piggyback here. Why can't we update the p-chain state? Shouldn't that be the only way we can trust a reset update?

PR #1306, opened 2026-05-07
Implements message verification for `MerkleValidatorSetRegistry`

mdelle1:
Introduces `MerkleValidatorSetRegistry` as a more gas-efficient alternative to `AvalancheValidatorSetRegistry`. Rather than storing the full validator set on-chain, this registry stores Merkle commitments to validator sets. The leaves are sorted in lexicographic order by public key encoded as `sha256(blsPublicKey || weight)`. 

Notably, this PR adds a `verifyMessage` function, which takes in a `TeleporterICMMessage`, and using its attestation field, verifies each of the signers are in the validator set Merkle commitment and uses those public keys to verify the aggregate BLS signature in the attestation. Registration cost goes from linear to constant-time in `SSTORE`s, and per-message verification trades off the per-signer cold `SLOAD`s to per-signer calldata and a single Merkle multi-proof verification cost, with the savings growing significantly as validator set size increases.

This PR adds:

- `MerkleValidatorSetRegistry` contract with `verifyMessage` functionality, currently only supported against the pChain validator set 
- `IMerkleValidatorSetRegistry` interface 
- Data types and parsers to `ValidatorSets.sol`, including functions `parseMerkleAttestation`, `serializeMerkleAttestation`, and `verifyMerkleAttestation`
- Tests covering the `verifyMessage` flow and round-trips to serialize/parse validator set Merkle attestations
- Bump OZ contracts to `v5.1.0` so `MerkleProof.processMultiProof` can use a custom hash function, e.g, `sha256` instead of the keccak default


icm-contracts/ethereum/utils/ValidatorSets.sol
```diff
@@ -212,6 +234,72 @@ library ValidatorSets {
...
+        // Perform Merkle multi-inclusion proof verification against the stored root
+        // TODO: Switch to multiProofVerifyCalldata once parseMerkleAttestation returns calldata slice offsets instead of a memory struct for additional gas savings
+        if (
+            !MerkleProof.multiProofVerify(att.proof, att.proofFlags, comm.root, leaves, _sha256Pair)
+        ) {
+            return false;
+        }
+
+        // Enforce uniqueness of signers
+        for (uint256 i = 0; i + 1 < numSigners;) {
```
batconjurer: I actually think the merkle multi-proof fails if there are duplicate signers and also if the signers are in the right order.

icm-contracts/ethereum/utils/ValidatorSets.sol
```diff
@@ -212,6 +234,72 @@ library ValidatorSets {
...
+                "Signers not strictly increasing"
+            );
+            unchecked {
+                ++i;
+            }
+        }
+
+        // Stake-weighted quorum threshold check
+        uint64 signerWeight = 0;
+        for (uint256 i = 0; i < numSigners; ++i) {
```
batconjurer, line 274: I'm not sure what's cheaper gas wise: a single for loop over all the signers to collect all this info or this approach where we can short circuit after each check.

icm-contracts/ethereum/tests/MerkleValidatorSetRegistry.t.sol
```diff
@@ -0,0 +1,277 @@
...
+    function dummyPChainValidatorSetSecretKeys() public pure returns (uint256[] memory) {
+        uint256[] memory secretKeys = new uint256[](4);
+        secretKeys[0] = 2;
+        secretKeys[1] = 3;
+        secretKeys[2] = 4;
+        secretKeys[3] = 5;
+        return secretKeys;
+    }
+}
+/**
```
batconjurer, line 33: nit: line break here

icm-contracts/ethereum/tests/MerkleValidatorSetRegistry.t.sol
```diff
@@ -0,0 +1,277 @@
...
+ * validator set commitments under arbitrary blockchain IDs, which is currently not implemented.
+ */
+
+contract MerkleValidatorSetRegistryVerifyMessageTest is MerkleValidatorSetRegistryCommon {
+    MerkleValidatorSetRegistry private _registry;
+    Validator[] private _validators;
+
+    function setUp() public {
+        uint256[] memory secretKeys = dummyPChainValidatorSetSecretKeys();
+        // Build a 5-validator set
```
batconjurer:
```suggestion
        // Build a 4-validator set
```

batconjurer (changes requested):
Generally looks good. One nit and a couple of questions on optimizing the verify flow.

remappings.txt
```diff
@@ -1,5 +1,7 @@
 @openzeppelin/contracts@5.0.2=lib/openzeppelin-contracts-upgradeable/lib/openzeppelin-contracts/contracts
+@openzeppelin/contracts@5.1.0=lib/openzeppelin-contracts-upgradeable/lib/openzeppelin-contracts/contracts
```
geoff-vball, line 2: So, we have two different remappings now pointing to the same thing. We should either update all the existing contracts to use `5.1.0`, or we should have a second submodule, with the remappings pointing to the appropriate one. Right now contracts with `@5.0.2` are pointing at `5.1.0` and that's confusing

geoff-vball: What I would probably do is make a standalone PR updating everything to `5.1.0`, and then submit the rest of this PR on top of that

icm-contracts/ethereum/utils/ValidatorSets.sol
mdelle1: You're right that the multi-proof would fail with duplicate or out-of-order signers, but OZ's `processMultiProof` doesn't have explicit duplicate or ordering checks, it just walks the leaves array, combining each leaf with either another leaf/computed-hash or a proof element according to proofFlags, and then reconstructs a candidate root. If we remove this uniqueness check, then if there are duplicate signers it would just show up as proof failed which is a less precise error log.

batconjurer: Right now, the proof would fail first, so we don't get the precise error log anyways. So at minimum, I guess we should re-arrange the checks. But we it's probably worth discussing gas-cost vs. helpful error messages for this call.

mdelle1: Ah yes, the check should occur beforehand, thanks for calling that out

mdelle1: As discussed I'll remove this check for gas savings

remappings.txt
mdelle1: Here is a PR for this: https://github.com/ava-labs/icm-services/pull/1312

icm-contracts/ethereum/utils/ValidatorSets.sol
mdelle1: The short-circuit approach should always be at least as good. By combining everything into a single loop, you would only save loop overhead which is on the order of 10s of gas, whereas the gas savings you get by skipping potential sha256 hashes or BLS G1 additions in short circuiting is much greater.

batconjurer (dismissed):

batconjurer (approved):

geoff-vball (approved):

PR #1311, opened 2026-05-11
Allow relayer to update the Merkle set validator registry

batconjurer:
## Why this should be merged
Built on #1314 

Given that the registering/applying of validators is not implemented in the MerkleValidatorSetRegistry yet, this should wait until that is done to be merged.
## How is this documented


icm-contracts/ethereum/utils/ValidatorSets.sol
```diff
@@ -314,6 +402,40 @@ library ValidatorSets {
         return serialized;
     }
 
+    /*
+    * @dev Serializes a ValidatorSetMerkleAttestation in the format expected by parseMerkleAttestation.
+    */
+    function serializeMerkleAttestation(
+        ValidatorSetMerkleAttestation memory att
+    ) internal pure returns (bytes memory) {
+        bytes2 codec = bytes2(0);
```
geoff-vball: This should probably be extracted as a `const` somewhere

relayer/validatorupdater/validator_set_merkle_commitment.go
```diff
@@ -0,0 +1,288 @@
...
+	// Pad to the next power of two so the tree is balanced and the resulting
+	// multi-proof is compatible with OpenZeppelin's MerkleProof.multiProofVerify.
+	n := nextPow2(len(validators))
+	signers := make([]*Validator, 0)
+	layer := make([]Node, n)
+	for i, v := range validators {
+		byteIdx := i / 8
+		bitIdx := uint(i % 8)
+		hash := validatorHash(validators[i])
+		if byteIdx < len(bitSetSig.Signers) && bitSetSig.Signers[byteIdx]&(1<<bitIdx) != 0 {
```
mdelle1: I ran the roundtrip ICM message with n=10 validators but it breaks due to a decoding issue here. `NewValidatorSetMerkleAttestation` decodes `BitSetSignature.Signers` bytes using little-endian byte order but avalanchego encodes the bitset as big-endian, so I think using avalanchego's own bitset `set.BitsFromBytes(...).Contains(i)` method may be the way to go.

batconjurer: Nice catch. Will fix

icm-contracts/tests/flows/services/merkle_updater.go
```diff
@@ -0,0 +1,488 @@
...
+func MerkleUpdater(
+	ctx context.Context,
+	log logging.Logger,
+	avalancheNetwork *network.LocalAvalancheNetwork,
+	ethereumNetwork *network.LocalEthereumNetwork,
+	teleporter utils.TeleporterTestInfo,
+) {
+	log.Info("Starting MerkleUpdater e2e test")
+
+	l1Info := avalancheNetwork.GetL1Infos()[2]
```
geoff-vball, line 53: Why do we grab index 2 here?

icm-contracts/tests/flows/services/merkle_updater.go
```diff
@@ -0,0 +1,488 @@
...
+		zap.Stringer("subnetID", l1Info.SubnetID),
+		zap.Uint32("networkID", networkID),
+		zap.Uint64("maxUpdateIntervalSeconds", merkleMaxUpdateIntervalSeconds),
+	)
+
+	ethClient := ethereumNetwork.EthClient
+	_, ethFundedKey := ethereumNetwork.GetFundedAccountInfo()
+	chainID := ethereumNetwork.ChainID
+	fundedAddress, fundedKey := avalancheNetwork.GetFundedAccountInfo()
+	_ = chainID
```
geoff-vball: Do we need this?

icm-contracts/tests/flows/services/merkle_updater.go
```diff
@@ -0,0 +1,488 @@
...
+	Expect(err).Should(BeNil())
+	Expect(pChainInitialized).Should(BeTrue(), "P-chain should be initialized by the constructor")
+
+	isRegistered, err := contract.IsRegistered(callOpts, l1BlockchainID)
+	Expect(err).Should(BeNil())
+	Expect(isRegistered).Should(BeFalse(), "L1 validator set should not be registered yet")
+
+	initialCommitment, err := contract.GetValidatorSetCommitment(callOpts, l1BlockchainID)
+	Expect(err).Should(BeNil())
+	Expect(initialCommitment.TotalWeight).Should(Equal(uint64(0)),
```
geoff-vball:
```suggestion
	Expect(initialCommitment.TotalWeight).Should(BeZero(),
```

icm-contracts/tests/flows/services/merkle_updater.go
```diff
@@ -0,0 +1,488 @@
...
+	defer rootChangeCancel()
+
+	rootChangeTicker := time.NewTicker(2 * time.Second)
+	defer rootChangeTicker.Stop()
+
+	var secondPChainHeight uint64
+	var secondTotalWeight uint64
+	var secondRoot [32]byte
+
+	for {
```
geoff-vball: This loop uses a slightly different idiom from the one on line 207. I think I prefer this one, but I think we might just want to refactor the functionality out to a helper function.

relayer/validatorupdater/merkle_validator_updater.go
```diff
@@ -0,0 +1,417 @@
...
+	contractAddress     common.Address
+	txOpts              *bind.TransactOpts
+
+	networkID    uint32
+	blockchainID ids.ID
+	subnetID     ids.ID
+	pollInterval time.Duration
+
+	maxUpdateInterval time.Duration
+
```
geoff-vball, line 42:
```suggestion
	subnetID     ids.ID
	
	pollInterval time.Duration
	maxUpdateInterval time.Duration

```

geoff-vball: Just a formatting nit

relayer/validatorupdater/merkle_validator_updater.go
```diff
@@ -0,0 +1,417 @@
...
+	"go.uber.org/zap"
+)
+
+type MerkleSetUpdater struct {
+	logger              logging.Logger
+	pChainClient        clients.CanonicalValidatorState
+	signatureAggregator *aggregator.SignatureAggregator
+	ethClient           *ethclient.Client
+	contract            *merklevalidatorsetregistry.MerkleValidatorSetRegistry
+	contractAddress     common.Address
```
geoff-vball, line 33: We don't use this anywhere

relayer/validatorupdater/merkle_validator_updater.go
```diff
@@ -0,0 +1,417 @@
...
+	txOpts *bind.TransactOpts,
+	networkID uint32,
+	blockchainID ids.ID,
+	subnetID ids.ID,
+	pollInterval time.Duration,
+	maxUpdateInterval time.Duration,
+) *MerkleSetUpdater {
+	if pollInterval == 0 {
+		pollInterval = defaultPollInterval
+	}
```
geoff-vball: I think we should do this logic outside of this function

relayer/validatorupdater/merkle_validator_updater.go
```diff
@@ -0,0 +1,417 @@
...
+		zap.Uint64("localPChainHeight", s.localPChainHeight),
+		zap.Uint64("pChainHeight", pChainHeight),
+		zap.String("newCommitmentRoot", fmt.Sprintf("0x%x", validatorSetUpdate.RootHash[:])),
+		zap.Bool("stale", s.isStale()),
+	)
+
+	if err := s.performUpdate(ctx, s.localPChainHeight, validatorSetUpdate, false); err != nil {
+		s.logger.Warn("Merkle root update failed, re-syncing from contract on next tick", zap.Error(err))
+		s.initialized = false
+		return err
```
geoff-vball: So this is a slightly weird case. We're logging a warning here but then returning an error which then gets logged as an error. We might just want to return `nil` here?

relayer/validatorupdater/merkle_validator_updater.go
```diff
@@ -0,0 +1,417 @@
...
+	ctx context.Context,
+	signedMsg *avalancheWarp.Message,
+	isFirstRegistration bool,
+) error {
+	var attestationValidators []*Validator
+	if isFirstRegistration {
+		// First registration is verified against the P-chain Merkle root stored in the
+		// contract constructor. Fetch the primary network validators used to build that
+		// root so the attestation proof is computed against the correct ordered set.
+		pChainWarpSet, err := s.pChainClient.GetProposedValidators(ctx, ids.Empty)
```
geoff-vball: This grabs the most recent p-chain validator set, but I think we need to get the p-chain validator at the height that was used in the contract constructor?

relayer/validatorupdater/merkle_validator_updater.go
```diff
@@ -0,0 +1,417 @@
...
+
+// ---------------------------------------------------------------------------
+// Shared helpers
+// ---------------------------------------------------------------------------
+
+func (s *MerkleSetUpdater) fetchSortedValidators(
+	ctx context.Context,
+	pChainHeight uint64,
+) ([]*Validator, error) {
+	allValidatorSets, err := s.pChainClient.GetAllValidatorSets(ctx, pChainHeight)
```
geoff-vball: We should add a function `GetValidatorsAt()` to the `CanonicalValidatorClient` that calls `v.client.GetValidatorsAt()` so we can fetch the validator set for one subnet at a time instead of fetching all validator for all subnets and throwing away everything but the one we want here

relayer/validatorupdater/validator_set_merkle_commitment.go
```diff
@@ -0,0 +1,291 @@
...
+		ProofFlags:         proofFlags,
+		AggregateSignature: [192]byte(aggregateSig),
+	}, nil
+}
+
+func (v *ValidatorSetMerkleAttestation) Bytes() []byte {
+	numFlags := len(v.ProofFlags)
+	flagBytesLen := (numFlags + 7) / 8
+
+	buf := make([]byte, 0, 2+4+len(v.Signers)*(96+8)+4+len(v.Proof)*32+4+flagBytesLen+192)
```
geoff-vball: We should make variables for all of these magic numbers

icm-contracts/tests/flows/services/merkle_updater.go
batconjurer: There are three L1s used for the subset updater, diff updater, and merkle updater tests. This is to keep the tests from conflicting with each other. Since merkle was added last, it got index 2.

icm-contracts/tests/flows/services/merkle_updater.go
batconjurer: oh, nice catch.

relayer/validatorupdater/merkle_validator_updater.go
batconjurer: Yeah, I think that makes sense

relayer/validatorupdater/merkle_validator_updater.go
batconjurer: Good catch.

icm-contracts/tests/flows/services/merkle_updater.go
```diff
@@ -0,0 +1,467 @@
...
+	)
+
+	// =========================================================================
+	// Phase 2: Wait for staleness-forced update.
+	// No new validators are added; after MaxUpdateIntervalSeconds the relayer
+	// refreshes the commitment to a newer P-chain height.
+	// =========================================================================
+	log.Info("Phase 2: Waiting for staleness-forced update (no validator changes)...")
+
+	// Issue a transaction toAdvance the P-chain by one block so the relayer has
```
mdelle1: typo toAdvance -> to advance

icm-contracts/tests/flows/services/merkle_updater.go
```diff
@@ -0,0 +1,467 @@
...
+	// a new height to commit to when its staleness timer fires.
+	_, err = avalancheNetwork.GetPChainWallet().IssueBaseTx(nil)
+	Expect(err).Should(BeNil())
+
+	elapsed := time.Since(firstRegistrationTime)
```
mdelle1: I think this should be `secondUpdateTime`

mdelle1 (approved):

geoff-vball (approved):

PR #1327, opened 2026-05-22
Add domain separators to Merkle hashing

### LINKED ISSUE #1319: Validator merkle trees should hash leaves differently than internal nodes
This is a standard best practice to prevent things like extension attacks

batconjurer:
## Why this should be merged
Closes #1319
## How is this documented


geoff-vball (dismissed):

icm-contracts/ethereum/utils/ValidatorSets.sol
```diff
@@ -922,7 +924,15 @@ library ValidatorSets {
...
-    function _sha256Pair(bytes32 a, bytes32 b) internal pure returns (bytes32) {
-        return a < b ? sha256(abi.encodePacked(a, b)) : sha256(abi.encodePacked(b, a));
+    function sha256InternalPair(bytes32 a, bytes32 b) internal pure returns (bytes32) {
+        return a < b
+            ? sha256(abi.encodePacked(uint256(0), a, b))
+            : sha256(abi.encodePacked(uint256(0), b, a));
+    }
+
+    function sha256Validator(
+        Validator memory val
```
mdelle1, line 940: Probably just rename to `hashValidator`? If so maybe we want to rename `validatorHash` to `hashValidator` on the Go-side on line 291 as well

mdelle1: I'd also leave a comment here for why we are pre-pending a 1 for second preimage resistance, and that callers should use this function in place of sha256 directly to hash validators

mdelle1 (dismissed):

geoff-vball (approved):

mdelle1 (approved):

PR #1329, opened 2026-05-22
Add max-gas-price gating to merkle validator updater

### LINKED ISSUE #1323: Add gas configurability to Relayer for `MerkleValidatorSetRegistry`
Add configuration to the Merkle Relayer to decline updating the validator set if the gas cost is too high

ylg-avalabs:
## Why this should be merged
The merkle updater submits a tx whenever the validator set changes or the staleness window elapses, regardless of network conditions. On destinations like Ethereum that gets expensive during fee spikes for updates that aren't time-critical. This adds an optional cap so the relayer can wait for gas to drop.

## How this works
New optional field on `ExternalEVMDestination`: `max-gas-price-gwei`. `0` (default) keeps existing behavior.

When set, `MerkleSetUpdater` checks `SuggestGasPrice` before each submission. If the suggested price is above the cap, it logs and returns without advancing local state, so the update is retried on the next poll. Check runs before signature aggregation to avoid wasted work. Both steady-state updates and first registration are gated. Only the merkle updater is wired up — subset/diff are untouched.

## How is this documented


relayer/config/external_evm_destination.go
```diff
@@ -29,4 +29,10 @@ type ExternalEVMDestination struct {
 	// weight change is below the threshold, an update is forced after this
 	// interval. 0 means no staleness cap (legacy behavior).
 	MaxUpdateIntervalSeconds uint64 `mapstructure:"max-update-interval-seconds" json:"max-update-interval-seconds,omitempty"` //nolint:lll
+	// Maximum suggested gas price (in gwei) on the destination chain at which
+	// the relayer will submit a validator-set update transaction. When the
+	// network's suggested gas price exceeds this threshold the update is
+	// deferred and retried on the next poll. 0 disables gas-price gating
+	// (legacy behavior). Currently honored by the merkle updater only.
```
geoff-vball: I don't think we need to describe legacy behaviour in these PRs as we've never shipped this functionality yet

relayer/validatorupdater/merkle_validator_updater.go
```diff
@@ -60,7 +69,12 @@ func NewMerkleSetUpdater(
 	subnetID ids.ID,
 	pollInterval time.Duration,
 	maxUpdateInterval time.Duration,
+	maxGasPriceGwei uint64,
 ) *MerkleSetUpdater {
+	var maxGasPriceWei *big.Int
+	if maxGasPriceGwei > 0 {
+		maxGasPriceWei = new(big.Int).Mul(new(big.Int).SetUint64(maxGasPriceGwei), weiPerGwei)
+	}
```
geoff-vball: I think if `maxGasPriceGwei uint64` then we should set the `*big.Int` to 0 instead of leaving it nil. That actually simplifies the logic here, and we don't risk dereferencing a nil pointer elsewhere.

relayer/validatorupdater/merkle_validator_updater.go
```diff
@@ -146,6 +158,16 @@ func (s *MerkleSetUpdater) checkAndUpdate(ctx context.Context) error {
 		zap.Bool("stale", s.isStale()),
 	)
 
+	ok, err := s.shouldSubmit(ctx)
```
mdelle1: Should we call this somewhere at the top of this function, e.g., after the P chain height/timestamp checks?

batconjurer (dismissed):

geoff-vball (dismissed):

relayer/validatorupdater/merkle_validator_updater.go
```diff
@@ -227,6 +257,33 @@ func (s *MerkleSetUpdater) isStale() bool {
 	return time.Since(s.lastUpdateTime) >= s.maxUpdateInterval
 }
 
+// shouldSubmit returns true when an update transaction may be submitted to
+// the destination chain. When a max-gas-price threshold is configured, this
+// queries the network's suggested gas price and returns false if it exceeds
+// the threshold. Callers should treat a false result as "defer until the
+// next poll" and must not advance local state.
+func (s *MerkleSetUpdater) shouldSubmit(ctx context.Context) (bool, error) {
```
geoff-vball: I would rename this function something like `gasPriceWithinBounds()`, at least for now while we're only using one dimension to decide if we should send or not. It's just a little more clear what we're checking in the calling function

geoff-vball: I would rename this function something like `gasPriceWithinBounds()`, at least for now while we're only using one dimension to decide if we should send or not. It's just a little more clear what we're checking in the calling function

geoff-vball (approved):

geoff-vball (approved):

PR #1337, opened 2026-05-26
Relayer: Implements P-chain sign fallback for L1 val set updates

### LINKED ISSUE #1324: Add disaster recovery mechanism for `MerkleValidatorSetRegistry` Relayer
Allow P-chain to sign off on updates to L1 validator sets. This would prevent the following scenario. Suppose there is a chain of validator set updates:
V1 --> V2 --> V3
If the on-chain update to V2 is skipped (e.g., if the gas threshold is exceeded), but there is no longer a quorum of V1 validators online, then it would be impossible to gather enough signatures to register V3 on-chain. In this scenario, we add the option to have the P-chain validator set, which is the source of truth, to sign off on the V3 update.

mdelle1:
Closes https://github.com/ava-labs/icm-services/issues/1324
This PR attempts to sign L1 validator set updates for the `MerkleValidatorSetRegistry` contract with the P-chain as a fallback.


relayer/validatorupdater/merkle_validator_updater.go
```diff
@@ -146,10 +146,19 @@ func (s *MerkleSetUpdater) checkAndUpdate(ctx context.Context) error {
...
+	// Try signing with the L1 first but fallback to the P-chain on failure
+	err = s.performUpdate(ctx, s.localPChainHeight, validatorSetUpdate, false, s.subnetID)
+	if err != nil {
+		s.logger.Warn("Self-signed update failed, retrying with P-Chain signature",
+			zap.Error(err))
+		err = s.performUpdate(ctx, s.localPChainHeight, validatorSetUpdate, false, constants.PrimaryNetworkID)
+		if err != nil {
+			s.logger.Warn("P-Chain fallback update also failed, re-syncing from contract on next tick",
+				zap.Error(err))
+			s.initialized = false
```
geoff-vball: Do we need to set `s.initialized = false` here anymore? If we fail to update with the subnet's validators and the primary network's validators, we've got some weird problem and we should probably retry with the subnet's validators first next time, I think

relayer/validatorupdater/merkle_validator_updater.go
```diff
@@ -243,14 +252,19 @@ func (s *MerkleSetUpdater) performUpdate(
 	onChainPChainHeight uint64,
 	validatorSetUpdate *ValidatorSetMerkleCommitment,
 	isFirstRegistration bool,
+	signingChain ids.ID,
```
batconjurer, line 330: I think we can simply the signature of this function. `isFirstRegistration` is now the same thing as the `signingChain` being ids.Empty

relayer/validatorupdater/merkle_validator_updater.go
```diff
@@ -286,20 +300,23 @@ func (s *MerkleSetUpdater) performUpdate(
...
 	ctx context.Context,
 	signedMsg *avalancheWarp.Message,
 	onChainPChainHeight uint64,
 	isFirstRegistration bool,
+	signingChain ids.ID,
 ) error {
+	// If the P-Chain signed the message in either initial registration or as a fallback,
+	// fetch the primary network validators used to build that root so the attestation
+	// proof is computed against the correct ordered set.
+	signedByPChain := isFirstRegistration || signingChain == constants.PrimaryNetworkID
```
batconjurer: same thing here, I think we can remove the `isFirstRegistration` flag

batconjurer (dismissed):

relayer/validatorupdater/merkle_validator_updater.go
```diff
@@ -167,10 +168,33 @@ func (s *MerkleSetUpdater) checkAndUpdate(ctx context.Context) error {
 		zap.Bool("stale", s.isStale()),
 	)
 
-	if err := s.performUpdate(ctx, s.localPChainHeight, validatorSetUpdate, false); err != nil {
-		s.logger.Warn("Merkle root update failed, re-syncing from contract on next tick", zap.Error(err))
-		s.initialized = false
-		return nil
```
mdelle1, line 173: Basically, we only retry if the chain is not the P-Chain and the fallback option is enabled. Otherwise, we re-tick.

relayer/validatorupdater/merkle_validator_updater.go
```diff
@@ -323,7 +348,6 @@ func (s *MerkleSetUpdater) performUpdate(
 	}
 
 	s.logger.Info("Signing new merkle root",
-		zap.Bool("isFirstRegistration", isFirstRegistration),
```
geoff-vball, line 326: The log here needs to be updated to `signingChain`

geoff-vball (approved):
Needs one small fix but otherwise LGTM

ylg-avalabs (approved):

geoff-vball (approved):

PR #1339, opened 2026-05-26
Contract: Add configurable P-chain fallback

### LINKED ISSUE #1338: Make P-chain signing fallbacks configurable for `MerkleValidatorSetRegistry`
As a follow up to https://github.com/ava-labs/icm-services/issues/1324 this PR gives L1s the option to only allow validator set updates from the L1's own set

mdelle1:
Closes https://github.com/ava-labs/icm-services/issues/1338


icm-contracts/ethereum/tests/MerkleValidatorSetRegistry.t.sol
```diff
@@ -611,6 +630,100 @@ contract MerkleValidatorSetRegistryRegisterUpdateTest is MerkleValidatorSetRegis
...
+        // Build L1 validator set
+        uint256[] memory l1SecretKeys = new uint256[](4); // 10, 11, 12,13
+        l1SecretKeys[0] = 11;
+        l1SecretKeys[1] = 12;
+        l1SecretKeys[2] = 13;
+        l1SecretKeys[3] = 10;
+        Validator[] memory l1Validators = new Validator[](4);
+        uint64 l1TotalWeight = 0;
+        bytes memory previousPublicKey = new bytes(BLST.BLS_UNCOMPRESSED_PUBLIC_KEY_INPUT_LENGTH);
+        for (uint256 i = 0; i < 4; i++) {
```
geoff-vball:
```suggestion
        for (uint256 i = 0; i < l1SecretKeys.length; i++) {
```

mdelle1:
Note this enables `allowPChainFallback` for the entire contract instance - configuring it for different chains within the same contract can be done, but we can discuss whether we want to take on the additional overhead in this thread, or if it is fine to deploy a new contract for the `allowPChainFallback = false` case.

icm-contracts/ethereum/tests/MerkleValidatorSetRegistry.t.sol
```diff
@@ -611,6 +630,100 @@ contract MerkleValidatorSetRegistryRegisterUpdateTest is MerkleValidatorSetRegis
...
+     *   3. The contract is updated.
+     */
+    function testL1SelfSignedUpdateSuccess() public {
+        // Build L1 validator set
+        uint256[] memory l1SecretKeys = new uint256[](4); // 10, 11, 12,13
+        l1SecretKeys[0] = 11;
+        l1SecretKeys[1] = 12;
+        l1SecretKeys[2] = 13;
+        l1SecretKeys[3] = 10;
+        Validator[] memory l1Validators = new Validator[](4);
```
geoff-vball:
```suggestion
        Validator[] memory l1Validators = new Validator[](l1SecretKeys.length);
```

icm-contracts/ethereum/tests/MerkleValidatorSetRegistry.t.sol
```diff
@@ -611,6 +630,100 @@ contract MerkleValidatorSetRegistryRegisterUpdateTest is MerkleValidatorSetRegis
...
 
+    /**
+     * @dev Happy path for the L1 self-signed update flow:
+     *   1. The P-Chain registers a new L1 with a Merkle root over a validator set.
+     *   2. The L1's own validators sign an update to the L1's commitment.
+     *   3. The contract is updated.
+     */
+    function testL1SelfSignedUpdateSuccess() public {
+        // Build L1 validator set
+        uint256[] memory l1SecretKeys = new uint256[](4); // 10, 11, 12,13
```
geoff-vball:
```suggestion
        uint256[] memory l1SecretKeys = new uint256[](4);
```

geoff-vball: I would generally advise against leaving comments like this when the code below speaks for itself. These types of comments often lie, like this one does right now ;)

icm-contracts/ethereum/tests/MerkleValidatorSetRegistry.t.sol
```diff
@@ -611,6 +630,100 @@ contract MerkleValidatorSetRegistryRegisterUpdateTest is MerkleValidatorSetRegis
...
+        ValidatorSetMerkleCommitment memory afterRegistration =
+            _registry.getValidatorSetCommitment(_NEW_CHAIN_ID);
+        assertEq(afterRegistration.root, l1Root);
+        assertEq(afterRegistration.totalWeight, l1TotalWeight);
+
+        // Construct the update signed by the L1's own validators
+        ValidatorSetMerkleCommitment memory updatedCommitment = ValidatorSetMerkleCommitment({
+            avalancheBlockchainID: _NEW_CHAIN_ID,
+            root: bytes32(uint256(0xababa)),
+            totalWeight: 10,
```
geoff-vball, line 692: Where does 10 come from?

icm-contracts/ethereum/tests/MerkleValidatorSetRegistry.t.sol
```diff
@@ -611,6 +630,100 @@ contract MerkleValidatorSetRegistryRegisterUpdateTest is MerkleValidatorSetRegis
...
+        bool[] memory proofFlags = new bool[](3);
+        proofFlags[0] = true;
+        proofFlags[1] = true;
+        proofFlags[2] = true;
+        ValidatorSetMerkleAttestation memory att = ValidatorSetMerkleAttestation({
+            signers: l1Validators,
+            proof: new bytes32[](0),
+            proofFlags: proofFlags,
+            aggregateBlsSig: aggregateSig
+        });
```
geoff-vball, line 709: Are we signing with 3/4 validators here? If so, why isn't `proofFlags` length 4, with `proofFlags[3] = false`?

icm-contracts/ethereum/MerkleValidatorSetRegistry.sol
```diff
@@ -74,13 +78,18 @@ contract MerkleValidatorSetRegistry is IMerkleValidatorSetRegistry, IAdapter {
...
+        // or the P-Chain if allowPChainFalback is enabled
+        if (!isRegistered(payloadBlockchainID)) {
+            require(
+                signingChainID == pChainID, "Initial registration must be signed by the P-Chain"
+            );
+        } else {
+            bool selfSigned = signingChainID == payloadBlockchainID;
+            bool pChainFallback = signingChainID == pChainID && allowPChainFallback;
+            require(selfSigned || pChainFallback, "Invalid signing chain");
+        }
```
geoff-vball, line 91: I think this would be simpler as:

```suggestion
        if (signingChainID == pChainID) {
            require(
                !isRegistered(payloadBlockchainID) || allowPChainFallback, "Blockchain already registered and pChainFallback disabled."
            );
        } else {
            require(signingChainID == payloadBlockchainID, "Invalid signing chain");
        }
```

icm-contracts/ethereum/tests/MerkleValidatorSetRegistry.t.sol
mdelle1: This signs with all validators and generally the length of proof flags is numLeaves + numHashes - 1 = 4 + 0 - 1 = 3

icm-contracts/ethereum/tests/MerkleValidatorSetRegistry.t.sol
mdelle1: I should have left this as l1TotalWeight but it is the weights of the validators 1 + 2 + 3 + 4

mdelle1: But also, here you can put arbitrary values, e.g. the root is also an arbitrary value, because in the register call it wont sanity check the payload, it will just check the signatures on top of it are valid and that the signers are in the currently stored root. The correctness of the payload itself will be on the AvalancheGo side

batconjurer (approved):

ylg-avalabs (approved):

icm-contracts/ethereum/tests/MerkleValidatorSetRegistry.t.sol
geoff-vball: Can you describe how the proofFlags are used? Or is it documented somewhere?

batconjurer: I can. So for a proof with a single leaf, each internal hash is a hash computed from the path from leaf to root along with a sibling hash. In a multi-merkle proof, a node may have both of its children leading to leafs being proven. It may also just have one child leading to a leaf being proven. The flags track the two cases so they can be handled separately.

geoff-vball:
@mdelle1 I've still got 1 lingering comment here

ylg-avalabs (approved):

geoff-vball (approved):

PR #1341, opened 2026-05-26
Remove CommonDestinationClient

### LINKED ISSUE #1343: Remove CommonDestinationClient
**Context and scope**
Remove `CommonDestinationClient`. It's a leaky abstraction that bakes too much functionality into the client, whereas we should be passing in the relevant parameters to the client to handle more thoughtlessly.

geoff-vball:
## Why this should be merged
Closes https://github.com/ava-labs/icm-services/issues/1343

Removes CommonDestinationClient to remove a layer of abstraction between clients. This is required to let the `messageHandler`s handle some of the abstractions between TeleporterV1 and TeleporterV2 instead of the client.

## How this works
Mostly by passing in more parameters to `SendTx()` instead of relying on them being available on `CommonDestinationClient`

## How this was tested
Existing tests

## How is this documented


utils/utils.go
```diff
@@ -57,6 +60,20 @@ func CheckStakeWeightExceedsThreshold(
 	return scaledTotalWeight.Cmp(scaledSigWeight) != 1
 }
 
+func SignedWarpMessageToAccessList(signedMessage *avalancheWarp.Message) types.AccessList {
```
geoff-vball, line 63: This used to be 
```func (c *destinationClient) AccessList(data txData) types.AccessList``` 
but does not need to be a receiver on the destination client at all

File: vms/evm/common.go [hunk 2]
```diff
@@ -47,18 +32,18 @@ type DestinationRPCClient interface {
...
 	EstimateBaseFee(ctx context.Context) (*big.Int, error)
 }
 
 type txData struct {
-	to            common.Address
-	gasLimit      uint64
-	gasFeeCap     *big.Int
-	gasTipCap     *big.Int
-	callData      []byte
-	signedMessage *avalancheWarp.Message
```
geoff-vball, line 60: We were passing the signed warp message through just to convert it into an access list, so we do the conversion earlier now.

vms/evm/common.go
```diff
@@ -239,12 +225,9 @@ func (s *concurrentSigner) waitForReceipt(
 // maximum priority fee per gas. The max fee per gas is set to the sum of the max base fee and the
 // max priority fee per gas.
 func getFeePerGas(
-	c CommonDestinationClient,
+	client DestinationRPCClient,
+	gasFeeConfig *GasFeeConfig,
 ) (*big.Int, *big.Int, error) {
-	rpcClient := c.RPCClient()
-	logger := c.Logger()
```
geoff-vball, line 245: This function no longer logs and relies on the caller to log an error

vms/evm/destination_client.go
```diff
@@ -105,11 +102,37 @@ func NewDestinationClient(
 		return nil, fmt.Errorf("failed to get chain ID from destination chain endpoint: %w", err)
 	}
 
-	var (
```
geoff-vball, line 108: Here, I simply switched the order of instantiation of the `destClient` and the `concurrentSigner`. They both require references to each other (currently, I can probably fix this), but it is currently cleaner to instantiate the client first

vms/evm/destination_client_test.go
```diff
@@ -136,25 +138,27 @@ func TestGetFeePerGas(t *testing.T) {
 			)
 
 			gasFeeCap, gasTipCap, err := destClient.getFeePerGas()
-			require.ErrorIs(t, test.expectedError, err)
+			require.ErrorIs(t, err, test.expectedError)
```
geoff-vball, line 141: This was the wrong order before. Now that we're returning a wrapped error from `getFeePerGas()` it was failing.

vms/evm/external_evm_destination_client.go
```diff
@@ -408,15 +382,7 @@ func (c *ExternalEVMDestinationClient) SimulateCall(
 		Data: callData,
 	}
 
-	result, err := c.ethClient.CallContract(ctx, callMsg, nil) // nil = latest block
-	if err != nil {
-		// Try to extract revert reason from error
-		c.logger.Debug("SimulateCall error details",
```
geoff-vball, line 414: I removed the logger from the client itself. The caller will eventually log this error up the stack.

vms/evm/common.go
```diff
@@ -288,34 +263,38 @@ func getFeePerGas(
 }
 
 func SendTx(
-	c CommonDestinationClient,
-	signedMessage *avalancheWarp.Message,
+	logger logging.Logger,
```
batconjurer, line 266: I guess this is ok. It's a hell of a lot of arguments

batconjurer (approved):

ylg-avalabs (approved):

mdelle1 (approved):

PR #1342, opened 2026-05-27
Merkle pruning

ylg-avalabs:
## Why this should be merged
Signed messages often include more signers than quorum requires. Each extra signer adds destination-chain calldata and gas. This prunes aggregates to the minimum quorum-meeting set for all aggregator output (merkle updates, ICM, etc.).

## How this works
`aggregateIfSufficientWeight` calls `pruneSignatureMapToQuorum`: pick the heaviest signers until quorum is met, then re-aggregate. No relayer changes — callers still use CreateSignedMessage.

## How this was tested
Unit tests

## How is this documented


batconjurer (dismissed):

relayer/validatorupdater/merkle_validator_updater.go
```diff
@@ -309,6 +313,7 @@ func (s *MerkleSetUpdater) sendUpdate(
 		if !ok {
 			return fmt.Errorf("primary network not found in validator sets at height %d", onChainPChainHeight)
 		}
+		signingWarpSet = pChainWarpSet
```
geoff-vball: We have `localValidatorsToWarpSet()` below. Can we also make a `warpSetToLocalValidators()` and use that on lines 317-327 instead of having the logic in this function?

signature-aggregator/aggregator/aggregator.go
```diff
@@ -916,6 +918,109 @@ func (s *SignatureAggregator) aggregateSignatures(
...
 
+// PruneBitSetSignatureToQuorum returns a BitSetSignature containing the smallest
+// subset of signers whose combined weight meets quorumPercentage, with the BLS
+// aggregate re-signed over the retained signers. validatorSet must be the
+// canonical set the bit indices refer to.
+// Returns bitSetSig unchanged if pruning is not possible (cache miss, missing
+// individual signature) or yields no reduction.
+func (s *SignatureAggregator) PruneBitSetSignatureToQuorum(
+	log logging.Logger,
+	unsignedMessage *avalancheWarp.UnsignedMessage,
```
geoff-vball: Lets just pass the message ID instead of the full message

signature-aggregator/aggregator/aggregator.go
```diff
@@ -916,6 +918,109 @@ func (s *SignatureAggregator) aggregateSignatures(
...
+			weight:   v.Weight,
+			pubKeyBz: PublicKeyBytes(v.PublicKeyBytes),
+		})
+	}
+
+	sort.SliceStable(signers, func(a, b int) bool {
+		if signers[a].weight != signers[b].weight {
+			return signers[a].weight > signers[b].weight
+		}
+		return signers[a].idx < signers[b].idx
```
geoff-vball: I don't think we should need two cases with `SliceStable()`

signature-aggregator/aggregator/aggregator.go
```diff
@@ -916,6 +918,109 @@ func (s *SignatureAggregator) aggregateSignatures(
...
+// canonical set the bit indices refer to.
+// Returns bitSetSig unchanged if pruning is not possible (cache miss, missing
+// individual signature) or yields no reduction.
+func (s *SignatureAggregator) PruneBitSetSignatureToQuorum(
+	log logging.Logger,
+	unsignedMessage *avalancheWarp.UnsignedMessage,
+	bitSetSig *avalancheWarp.BitSetSignature,
+	validatorSet validators.WarpSet,
+	quorumPercentage uint64,
+) (*avalancheWarp.BitSetSignature, error) {
```
geoff-vball: We should also do a `log = log.With(zap.Stringer("messageID", messageID))` to cut down on adding it to every log in this function

signature-aggregator/aggregator/aggregator.go
```diff
@@ -916,6 +918,109 @@ func (s *SignatureAggregator) aggregateSignatures(
 	return aggSig, vdrBitSet, nil
 }
 
+// PruneBitSetSignatureToQuorum returns a BitSetSignature containing the smallest
+// subset of signers whose combined weight meets quorumPercentage, with the BLS
+// aggregate re-signed over the retained signers. validatorSet must be the
+// canonical set the bit indices refer to.
+// Returns bitSetSig unchanged if pruning is not possible (cache miss, missing
+// individual signature) or yields no reduction.
+func (s *SignatureAggregator) PruneBitSetSignatureToQuorum(
```
geoff-vball: I don't think I like the fact that we're first getting the full aggregate signature and bitset, and then coming in here and doing all the work again in a rather complicated way. 

I think a better flow would either to use prune the signature map in all cases, and prune the signature set in `aggregateIfSufficientWeight()`, or we can create a different function `aggregateMinimalIfSufficientWeight()` that the merkle validator updater can call.

If we prune the signature set for all ICM messages, we'll save a bit of gas cost for the messages because we pay per signer there as well.

If you make these changes you can ignore the couple comments below this.

signature-aggregator/aggregator/aggregator_test.go
```diff
@@ -872,3 +872,245 @@ func TestPopulateSignatureMapFromCache(t *testing.T) {
...
+			blsPublicKey:      pubKey,
+			blsPublicKeyBytes: bls.PublicKeyToUncompressedBytes(pubKey),
+		}
+	}
+
+	// Canonical order is by uncompressed public-key bytes; preserve the input
+	// weight alongside each info by tagging the index before sort.
+	type tagged struct {
+		info   validatorInfo
+		weight uint64
```
geoff-vball: We should just add `weight` to `validatorInfo` instead of wrapping it

geoff-vball (changes requested):

signature-aggregator/aggregator/aggregator.go
```diff
@@ -803,7 +809,25 @@ func (s *SignatureAggregator) aggregateIfSufficientWeight(
...
+	// meets quorum. Fewer signers means smaller calldata on destination chains.
+	prunedSigMap := pruneSignatureMapToQuorum(
+		signatureMap,
+		canonicalValidators,
+		totalWeight,
+		quorumPercentage,
+	)
+	if pruned, original := len(prunedSigMap), len(signatureMap); pruned < original {
+		log.Debug(
+			"Pruned signature map to minimum quorum",
```
mdelle1: The check is just that pruned < original, it doesnt have to be at quorum though. It should probably just say "Pruned signature map"

signature-aggregator/aggregator/aggregator.go
```diff
@@ -825,6 +849,45 @@ func (s *SignatureAggregator) aggregateIfSufficientWeight(
...
+	}
+	sort.SliceStable(signers, func(a, b int) bool {
+		return signers[a].weight > signers[b].weight
+	})
+
+	pruned := make(map[int][bls.SignatureLen]byte, len(signers))
+	accumulated := big.NewInt(0)
+	for _, sgn := range signers {
+		pruned[sgn.idx] = signatureMap[sgn.idx]
+		accumulated.Add(accumulated, new(big.Int).SetUint64(sgn.weight))
```
mdelle1: I think you can re-use a buffer rather than creating a new big Int in every iteration, i.e.
define `weightBuf := new(big.Int)` before the loop on line 881, and then do
`accumulated.Add(accumulated, weightBuf.SetUint64(sgn.weight))`

signature-aggregator/aggregator/aggregator.go
```diff
@@ -803,7 +809,25 @@ func (s *SignatureAggregator) aggregateIfSufficientWeight(
...
+		quorumPercentage,
+	)
+	if pruned, original := len(prunedSigMap), len(signatureMap); pruned < original {
+		log.Debug(
+			"Pruned signature map",
+			zap.Stringer("messageID", unsignedMessage.ID()),
+			zap.Int("originalSigners", original),
+			zap.Int("prunedSigners", pruned),
+		)
+	}
```
geoff-vball: We can remove this before merging

signature-aggregator/aggregator/aggregator.go
```diff
@@ -825,6 +849,46 @@ func (s *SignatureAggregator) aggregateIfSufficientWeight(
...
+
+	pruned := make(map[int][bls.SignatureLen]byte, len(signers))
+	accumulated := new(big.Int)
+	weightBuf := new(big.Int)
+	for _, sgn := range signers {
+		pruned[sgn.idx] = signatureMap[sgn.idx]
+		accumulated.Add(accumulated, weightBuf.SetUint64(sgn.weight))
+		if utils.CheckStakeWeightExceedsThreshold(accumulated, totalWeight, quorumPercentage) {
+			break
+		}
```
geoff-vball: So this is going to do a lot of multiplication of big ints every time we call it. Might be trivial, but I think we should calculate the quorum weight once, and then just check if we've surpassed that yet

signature-aggregator/aggregator/aggregator_test.go
```diff
@@ -872,3 +875,231 @@ func TestPopulateSignatureMapFromCache(t *testing.T) {
 	// The expected weight is the weight of the first validator
 	require.Equal(t, connectedValidators.ValidatorSet.Validators[0].Weight, accWeight.Uint64())
 }
+
+// makeConnectedValidatorsWithWeights mirrors makeConnectedValidators but lets each
+// validator have its own weight. Validators are returned in canonical (pubkey)
+// order, paired with their signers in the same order.
+func makeConnectedValidatorsWithWeights(
```
geoff-vball, line 857: We shouldn't duplicate the code in `makeConnectedValidators`. We should probably get `makeConnectedValidators` call `makeConnectedValidatorsWithWeights` with a default weight

geoff-vball (approved):

PR #1352, opened 2026-05-29
Request validator signatures from the Signature Aggregator prioritized by weight

ylg-avalabs:
## How is this documented


signature-aggregator/aggregator/aggregator.go
```diff
@@ -320,7 +326,191 @@ func (s *SignatureAggregator) getCachedSignaturesForMessage(
...
+// Returns 0 if the threshold has already been met.
+func weightGap(accumulatedWeight *big.Int, totalWeight, quorumPercentage uint64) uint64 {
+	// required = ceil(totalWeight * quorumPercentage / 100)
+	required := new(big.Int).Mul(
+		new(big.Int).SetUint64(totalWeight),
+		new(big.Int).SetUint64(quorumPercentage),
+	)
+	denom := new(big.Int).SetUint64(quorumPercentageDenominator)
+	required.Add(required, new(big.Int).Sub(denom, big.NewInt(1)))
+	required.Quo(required, denom)
```
geoff-vball: I think we already have this functionality somewhere as a helper function for getting the required weight for a quorum

signature-aggregator/aggregator/aggregator.go
```diff
@@ -320,7 +326,191 @@ func (s *SignatureAggregator) getCachedSignaturesForMessage(
...
+		candidates = append(candidates, validatorIndex{index: i, weight: v.Weight})
+	}
+	sort.SliceStable(candidates, func(i, j int) bool {
+		if candidates[i].weight != candidates[j].weight {
+			return candidates[i].weight > candidates[j].weight
+		}
+		return candidates[i].index < candidates[j].index
+	})
+	return candidates
+}
```
geoff-vball: This will cause us to re-query validators we haven't received a response from yet, which I don't think we want. I think we should sort the validator set once by weight, and just iterate through it, stopping when it makes sense.

signature-aggregator/aggregator/aggregator.go
```diff
@@ -320,7 +326,191 @@ func (s *SignatureAggregator) getCachedSignaturesForMessage(
...
+	batchNodes := set.NewSet[ids.NodeID](0)
+	batchWeight := uint64(0)
+	for *cursor < len(candidates) && batchWeight < minBatchWeight {
+		c := candidates[*cursor]
+		*cursor++
+		v := vdrs.ValidatorSet.Validators[c.index]
+		added := false
+		for _, nodeID := range v.NodeIDs {
+			if vdrs.ConnectedNodes.Contains(nodeID) && !queriedNodes.Contains(nodeID) {
+				queriedNodes.Add(nodeID)
```
geoff-vball: Ah, I see that we track the nodes we've already queried here separately. I think this is needlessly overcomplicated. We should just sort the validators once by weight, have this as a list of `unqueriedNodes` and then split the slice into `batchNodes` and update `unqueriedNodes` with the remainder.

signature-aggregator/aggregator/aggregator.go
```diff
@@ -320,7 +326,191 @@ func (s *SignatureAggregator) getCachedSignaturesForMessage(
...
+	)
+	denom := new(big.Int).SetUint64(quorumPercentageDenominator)
+	required.Add(required, new(big.Int).Sub(denom, big.NewInt(1)))
+	required.Quo(required, denom)
+	if accumulatedWeight.Cmp(required) >= 0 {
+		return 0
+	}
+	gap := new(big.Int).Sub(required, accumulatedWeight)
+	if !gap.IsUint64() {
+		return ^uint64(0)
```
geoff-vball:
```suggestion
		return math.MaxUint64
```

signature-aggregator/aggregator/aggregator.go
```diff
@@ -320,7 +326,191 @@ func (s *SignatureAggregator) getCachedSignaturesForMessage(
...
+	})
+	return candidates
+}
+
+// pickNextBatch advances [cursor] through [candidates] and returns the set of node IDs
+// to query next. Validators are added to the batch in weight-descending order until the
+// cumulative weight of newly added validators is at least [minBatchWeight], or there are
+// no more candidates. [queriedNodes] is updated in place with any node added to the batch.
+// Returns the batch's node IDs and the cumulative weight of newly added validators.
+func pickNextBatch(
```
geoff-vball: We should probably also set a minimum batch size. If we have a network with 5 valdiators of equal weight, but the first one is offline, it doesn't make sense to send requests to the first 4 validators and then wait until the timeout to send the last request.

batconjurer (approved):

signature-aggregator/aggregator/aggregator.go
```diff
@@ -320,7 +333,177 @@ func (s *SignatureAggregator) getCachedSignaturesForMessage(
...
+				nodeIDs = append(nodeIDs, nodeID)
+			}
+		}
+		// Skip validators we have no connected node to query.
+		if len(nodeIDs) == 0 {
+			continue
+		}
+		queryable = append(queryable, queryableValidator{weight: v.Weight, nodeIDs: nodeIDs})
+	}
+	sort.SliceStable(queryable, func(i, j int) bool {
```
mdelle1: This `sort.SliceStable` is used elsewhere so it might be worth making this a utils function like in https://github.com/ava-labs/icm-services/pull/1354, it can be a follow up PR though 👍

signature-aggregator/aggregator/aggregator.go
```diff
@@ -332,193 +515,174 @@ func (s *SignatureAggregator) collectSignaturesWithRetries(
...
-			zap.Int("signatureMapSize", len(signatureMap)),
-			zap.Int("responsesExpected", responsesExpected),
-		)
+	// Sort the queryable validators by weight once. Each batch consumes a prefix of
+	// [remaining], so a validator is never queried more than once per aggregation.
+	remaining := queryableValidatorsByWeight(vdrs, signatureMap)
+	pendingResponses := 0
+	// Buffer the merged channel to accommodate every possible response without blocking
+	// the per-batch forwarder goroutines.
+	mergedChan := make(chan message.InboundMessage, len(vdrs.ValidatorSet.Validators))
```
geoff-vball: This is called `mergedChan` because it is shared across the batches?

signature-aggregator/aggregator/aggregator.go
```diff
@@ -332,193 +515,174 @@ func (s *SignatureAggregator) collectSignaturesWithRetries(
...
-		log.Debug(
-			"Aggregator collecting signatures from peers.",
-			zap.Int("validatorSetSize", len(vdrs.ValidatorSet.Validators)),
-			zap.Int("signatureMapSize", len(signatureMap)),
-			zap.Int("responsesExpected", responsesExpected),
-		)
+	// Sort the queryable validators by weight once. Each batch consumes a prefix of
+	// [remaining], so a validator is never queried more than once per aggregation.
+	remaining := queryableValidatorsByWeight(vdrs, signatureMap)
+	pendingResponses := 0
```
geoff-vball: I feel like we should use an atomic int for this. I think everything is okay right now, but it feels like a bug waiting to happen

signature-aggregator/aggregator/aggregator.go
```diff
@@ -332,193 +515,174 @@ func (s *SignatureAggregator) collectSignaturesWithRetries(
...
-	err := utils.WithRetriesTimeout(operation, notify, s.signatureRequestTimeout)
-	if err != nil {
-		logger.Warn(
-			"Failed to collect a threshold of signatures",
-			zap.Uint64("accumulatedWeight", accumulatedSignatureWeight.Uint64()),
-			zap.Uint64("totalValidatorWeight", vdrs.ValidatorSet.TotalWeight),
-			zap.Error(err),
-		)
-		return nil, errNotEnoughSignatures
+	logger.Warn(
```
geoff-vball: I don't think we want this warn log. If we don't collect enough signatures, we return and then log the exact same thing as an error

signature-aggregator/aggregator/aggregator.go
boraerincplusplus: Hey yall, just wanted to get some reps up and familiarize myself with code review. 

Noticed that `SortByWeightDescending`  in `utils/utils.go` uses `sort.SliceStable`, which uses reflection internally and may allocate objects on the heap. The `slices.SortStableFunc` from Go 1.21+ avoids both.

The docs explicitly mention this, and `avalanchego` is converging on this idiom such as in `avalanchego/utils/sorting.go`.

Performance is negligible at this call site, but might be worthwhile for consistency. Happy to send a follow-up PR if it's useful, and curious to hear what you think.

[https://pkg.go.dev/sort#SliceStable](https://pkg.go.dev/sort#SliceStable)
[https://pkg.go.dev/slices#SortStableFunc](https://pkg.go.dev/slices#SortStableFunc)

P.S here's a little article if anyone wants to nerd out about Go:
[To sort.Slice or to slices.SortFunc?](https://medium.com/@grzegorz.mika/to-sort-slice-or-to-slices-sortfunc-64838522b17e)

ylg-avalabs: Great catch, thanks for the detailed write-up (and the article). I've switched `SortByWeightDescending` over to `slices.SortStableFunc` + `cmp.Compare` in this PR. Appreciate the review!

signature-aggregator/aggregator/aggregator.go
```diff
@@ -332,193 +514,174 @@ func (s *SignatureAggregator) collectSignaturesWithRetries(
...
+					ctx,
+					log,
+					unsignedMessage,
+					reqBytes,
+					sourceSubnet,
+					batchNodes,
+					mergedChan,
+				)
+				if err != nil {
+					log.Warn("Failed to send batched signature request", zap.Error(err))
```
geoff-vball: I think we either need to `continue` here or return an error. I could be wrong though

geoff-vball (commented):
So, I'm not sure if I like how much added complexity this feature adds for the benefit we'll be getting. I think the scenario we want to avoid is querying a bunch of very small validators, but I think we could achieve this by sorting validators by weight, and then querying validators until we reach something like 95% of stake and the validator weight drops below 1% of stake.

If we do want it, I'm unsure of good value for `batchResponseTimeout` and `minBatchSize`.

signature-aggregator/aggregator/aggregator.go
mdelle1: `queryableValidatorsByWeight` collects all node IDs for a given validator, so couldn't 2 nodes corresponding to a single validator return valid signatures here, in which case we would be double counting the validator's weight?

signature-aggregator/aggregator/aggregator.go
```diff
@@ -320,7 +332,177 @@ func (s *SignatureAggregator) getCachedSignaturesForMessage(
...
+		requestID,
+		utils.DefaultAppRequestTimeout,
+		reqBytes,
+	)
+	if err != nil {
+		return 0, fmt.Errorf("failed to create app request message: %w", err)
+	}
+
+	for nodeID := range batchNodes {
+		s.network.RegisterAppRequest(ids.RequestID{
```
mdelle1: `sendBatch` registers all of the `batchNodes` with the network but returns `expectedResponses` which  only includes nodes `Send` reached. The caller correctly stops expecting those nodes but the network's `responseChan` is still registered against the full set. Perhaps calling `Send` first and registering/counting only `sentTo` so both sides agree is correct?

signature-aggregator/aggregator/aggregator.go
```diff
@@ -320,7 +327,157 @@ func (s *SignatureAggregator) getCachedSignaturesForMessage(
...
+// node), sorted by descending weight. The canonical validator set is already ordered by
+// public key, so a stable sort preserves that order for validators of equal weight.
+func queryableValidatorsByWeight(
+	vdrs *peers.CanonicalValidators,
+	signatureMap map[int][bls.SignatureLen]byte,
+) []queryableValidator {
+	queryable := make([]queryableValidator, 0, len(vdrs.ValidatorSet.Validators))
+	for i, v := range vdrs.ValidatorSet.Validators {
+		if _, has := signatureMap[i]; has {
+			continue
```
geoff-vball, line 366: I think we have a problem here where if we already have a signature cached, we don't add the validator to the returned array of `queryableValidator`, but then do all the weight calculations without accounting for the weight that has already signed the message.

Instead of dealing with cached signatures here, lets do it in `nodesToQuery()` so that we can add the weight to `cumulative`.

signature-aggregator/aggregator/aggregator.go
```diff
@@ -320,7 +327,157 @@ func (s *SignatureAggregator) getCachedSignaturesForMessage(
 	return signatureMap, accumulatedSignatureWeight
 }
 
-func (s *SignatureAggregator) collectSignaturesWithRetries(
+// queryableValidator is a single canonical validator that still needs to be queried,
+// paired with its weight and the connected node IDs to send the request to.
+type queryableValidator struct {
```
geoff-vball: Instead of creating this new type, let's just use the existing `Warp` type that also includes the public key.

geoff-vball (commented):
Sorry I forgot to submit this feedback earlier!

utils/utils.go
```diff
@@ -89,6 +91,15 @@ func SignedWarpMessageToAccessList(signedMessage *avalancheWarp.Message) types.A
...
 }
 
+// SortByWeightDescending sorts [items] in place by descending weight, using [weight]
+// to extract each item's weight. The sort is stable, so items of equal weight retain
+// their original relative order.
+func SortByWeightDescending[T any](items []T, weight func(T) uint64) {
+	slices.SortStableFunc(items, func(a, b T) int {
+		return cmp.Compare(weight(b), weight(a))
+	})
+}
```
geoff-vball: nit: This function should just be `SortDescending`. The input parameter could be anything

signature-aggregator/aggregator/aggregator.go
```diff
@@ -312,7 +319,169 @@ func (s *SignatureAggregator) getCachedSignaturesForMessage(
 	return signatureMap, accumulatedSignatureWeight
 }
 
-func (s *SignatureAggregator) collectSignaturesWithRetries(
+// validatorsByWeight returns the canonical validators sorted by descending weight. The
+// canonical validator set is already ordered by public key, so a stable sort preserves that
+// order for validators of equal weight.
+func validatorsByWeight(vdrs *peers.CanonicalValidators) []*validators.Warp {
```
geoff-vball: nit: This should just take a `[]*validators.Warp` as a parameter and rename func `copyValidatorsSortedByWeight`

signature-aggregator/aggregator/aggregator.go
```diff
@@ -312,7 +319,169 @@ func (s *SignatureAggregator) getCachedSignaturesForMessage(
...
+func finishResponses(responseChan <-chan message.InboundMessage) {
+	for resp := range responseChan {
+		resp.OnFinishedHandling()
+	}
+}
+
+// collectSignatures sends a single signature request to the highest-weight validators (see
+// [nodesToQuery]) and aggregates the responses. It returns once the accumulated weight meets
+// the required quorum plus buffer, falling back to just the required quorum once all responses
+// are in. The collection is bounded by [signatureRequestTimeout].
```
geoff-vball:
```suggestion
// collectSignatures sends a single signature request to the highest-weight validators (see
// [nodesToQuery]) and aggregates the responses. It returns once the accumulated weight meets
// the required quorum. The collection is bounded by [signatureRequestTimeout].
```

signature-aggregator/aggregator/aggregator.go
```diff
@@ -324,168 +493,144 @@ func (s *SignatureAggregator) collectSignaturesWithRetries(
...
+	)
 
-		failedSendNodes := make([]ids.NodeID, 0, responsesExpected)
-		for nodeID := range vdrSet {
-			if !sentTo.Contains(nodeID) {
-				responsesExpected--
-				failedSendNodes = append(failedSendNodes, nodeID)
-				s.metrics.FailuresSendingToNode.Inc()
-			}
+	if queryNodes.Len() > 0 {
```
geoff-vball: This `if` statement is a little weird. `requestSignatures` should just return `nil, nil` if `queryNodes.Len() == 0`

signature-aggregator/aggregator/aggregator.go
```diff
@@ -656,53 +801,41 @@ func (s *SignatureAggregator) isSubnetL1(ctx context.Context, subnetID ids.ID) (
...
 	}
 
 	validator, vdrIndex := connectedValidators.GetValidator(nodeID)
 	signature, valid := s.isValidSignatureResponse(log, unsignedMessage, response, validator.PublicKey)
-	// Cache any valid signature, but only include in the aggregation if the validator is not explicitly
-	// excluded, that way we can use the cached signature on future requests if the validator is
-	// no longer excluded
+	// Cache any valid signature, but only include in the aggregation if the validator is not
+	// explicitly excluded, so that we can reuse the cached signature on future requests if
+	// the validator is no longer excluded.
```
geoff-vball: We should flip the `if/else` here. We can do `if !valid... log and return` and then un-nest the rest of the clause

geoff-vball: I know it wasn't part of your PR, but might as well make the change with the other fixes

geoff-vball (commented):
Sorry for the delay with these reviews!

signature-aggregator/aggregator/aggregator.go
```diff
@@ -324,168 +492,142 @@ func (s *SignatureAggregator) collectSignaturesWithRetries(
...
+			zap.Uint64("signatureWeight", accumulatedSignatureWeight.Uint64()),
+			zap.Uint64("totalValidatorWeight", totalWeight),
 		)
+		return signedMsg, nil
 	}
 
-	err := utils.WithRetriesTimeout(operation, notify, s.signatureRequestTimeout)
+	// Process any responses that arrived without reaching quorum during the request, in case
+	// cached signatures plus the collected ones now suffice.
+	signedMsg, err = s.aggregateIfSufficientWeight(
```
geoff-vball: This block (and a bit below) don't matter anymore. Previously, with the quorum percentage buffer, we would try to get above the quorum + buffer, but then try to aggregate if between the quorom and quorum + buffer. Without the buffer, we don't need this call to `aggregateIfSufficientWeight`.

signature-aggregator/aggregator/aggregator.go
```diff
@@ -324,168 +492,119 @@ func (s *SignatureAggregator) collectSignaturesWithRetries(
...
 
-	err := utils.WithRetriesTimeout(operation, notify, s.signatureRequestTimeout)
+	// The caller logs this failure at error level, so don't log it again here.
+	return nil, errNotEnoughSignatures
+}
+
+// requestSignatures sends a single request to [queryNodes] and processes responses as they
+// arrive, returning a signed message once [quorumPercentage] of stake is accumulated, or
+// (nil, nil) if the threshold isn't reached before all responses are in or the context ends.
+func (s *SignatureAggregator) requestSignatures(
```
geoff-vball, line 537: Sorry, one last change I promise! If we return `nil, errNotEnoughSignatures` in the cases where we're returning `nil, nil` now, then the caller doesn't have to handle the `nil, nil` case and `collectSignatures()` can just return `s.requestSignatures()`

geoff-vball (approved):

PR #1358, opened 2026-06-03
Enforce `sendMessage` caller is the originating Teleporter messenger

mdelle1:


icm-contracts/ethereum/AvalancheValidatorSetRegistry.sol
```diff
@@ -72,6 +72,7 @@ contract AvalancheValidatorSetRegistry is IAvalancheValidatorSetRegistry, IAdapt
     function sendMessage(
         TeleporterMessageV2 calldata message
     ) external {
+        require(msg.sender == message.originTeleporterAddress, "unauthorized sender");
```
batconjurer, line 75: Can we move this check up into the TeleporterV2 contract? That would minimize the change footprint for one, but it would also be nice if all the Nick's method security assumptions could be located in the TeleporterV2 and decoupled from these verifiers. Then we might be able to reuse some of this verifier code in places where we don't have Nick's method.

batconjurer (changes requested):

icm-contracts/ethereum/AvalancheValidatorSetRegistry.sol
geoff-vball: We'll chat about this in standup :)

mdelle1: We need the send-side to enforce that the caller of `sendMessage` is the Teleporter contract. We do this by requiring `msg.sender == originTeleporterAddress`. This isn't enough on its own because the message author controls `originTeleporterAddress`, but the Teleporter contract receive-side [check pins](https://github.com/ava-labs/icm-services/blob/b527b7c4bad77c6f33f36492323b9538eef84f3a/icm-contracts/common/TeleporterMessengerV2.sol#L270) `originTeleporterAddress` to the actual Teleporter address, so these together enforce the `sendMessage` caller to be the Teleporter contract.

ylg-avalabs (approved):

geoff-vball (commented):
Could you add a unit test for this? Otherwise lgtm

geoff-vball (approved):

ylg-avalabs (approved):

batconjurer (approved):

PR #1361, opened 2026-06-03
`ZKAdapter` implements `IAdapter` and fixups to `verifyMessage`

mdelle1:
This PR makes the following changes to `ZKAdapter`:
- Implements the full `IAdapter`interface, previously only `IMessageVerifier`, adding a `sendMessage` function that originates a cross-chain message by emitting it as a `TeleporterV2MessageSent` log.`sendMessage` is gated with `require(msg.sender == message.originTeleporterAddress)` which ensures only the genuine Teleporter messenger can originate a message. See [this PR](https://github.com/ava-labs/icm-services/pull/1358) for more details
- `verifyMessage` was strengthened to bind the proven log to the message being verified. It now pins the log's emitter, event topic, and requires the proven log's payload to equal the actual message payload so an attestation for an unrelated log can't be accepted


icm-contracts/avalanche/verifiers/ethereum/ZKAdapter.sol
```diff
@@ -5,9 +5,12 @@
...
 pragma solidity ^0.8.30;
 
-import {ZKStateManager, ZKEventInfo} from "./ZKStateManager.sol";
-import {TeleporterICMMessage, IMessageVerifier} from "../../../common/ITeleporterMessengerV2.sol";
+import {ZKStateManager} from "./ZKStateManager.sol";
+import {
+    TeleporterICMMessage, TeleporterMessageV2
+} from "../../../common/ITeleporterMessengerV2.sol";
 import {Consensus, Execution, Receipt} from "./StateManagerLibrary.sol";
+import {IAdapter} from "../../../common/ITeleporterMessengerV2.sol";
```
geoff-vball: We're importing from "../../../common/ITeleporterMessengerV2.sol" twice

icm-contracts/avalanche/verifiers/ethereum/ZKAdapter.sol
```diff
@@ -21,9 +24,22 @@ struct Attestation {
...
+
+    /// @notice Address of the trusted source chain adapter that emits TeleporterV2MessageSent logs
+    address public immutable sourceEmitter;
+
+    event TeleporterV2MessageSent(bytes encodedMessage);
+
     constructor(
-        uint256 sourceChainId,
+        uint256 sourceChainId_,
+        uint32 sourceNetworkId_,
```
geoff-vball: I don't think we should be passing this into the constructor here. This contract should figure out the networkId that it's been deployed on, and only accept messages from the same networkId

icm-contracts/avalanche/verifiers/ethereum/ZKAdapter.sol
```diff
@@ -42,30 +58,40 @@ contract ZKAdapter is ZKStateManager, IMessageVerifier {
...
+     * trusted source-side adapter (`sourceEmitter`) whose payload equals the message being verified.
+     */
     function verifyMessage(
         TeleporterICMMessage calldata message
     ) external returns (bool) {
         Attestation memory att = abi.decode(message.attestation, (Attestation));
-        this.proveLogAndExecute(att.execProof, att.logProof);
-        return true;
-    }
+        require(message.sourceBlockchainID == bytes32(sourceChainId), "bad source chain");
```
geoff-vball, line 83: Do we need to specify the source blockchain in this contract? This contract should be able to verify messages from any source chain, and leave it up to the caller to decide if it wants to receive a message from that chain

icm-contracts/avalanche/verifiers/ethereum/ZKStateManager.sol
```diff
@@ -183,45 +181,29 @@ contract ZKStateManager is AccessControl {
...
      * @param execProof The execution proof linking a receipt root to a trusted beacon block root.
      * @param logProof The log proof establishing the receipt and log inclusion against a trusted receipts root.
      */
-    function proveLogAndExecute(
+    function verifyLogAndExtract(
         Execution.Proof calldata execProof,
         Receipt.Proof calldata logProof
-    ) external {
-        // Verify the execution state proof against a stored beacon block root.
+    ) external returns (bytes memory logData) {
```
geoff-vball: Named returns like this are usually considered bad code style. We should explicitly return logData after emitting the event.

icm-contracts/avalanche/verifiers/ethereum/ZKAdapter.sol
```diff
@@ -21,9 +24,22 @@ struct Attestation {
...
-contract ZKAdapter is ZKStateManager, IMessageVerifier {
+contract ZKAdapter is ZKStateManager, IAdapter {
+    /// @notice topic0 of the TeleporterV2MessageSent event which is the only log topic verifyMessage accepts
+    bytes32 private constant _MESSAGE_SENT_TOPIC = keccak256("TeleporterV2MessageSent(bytes)");
+
+    /// @notice The network ID (mainnet/testnet/local) messages must originate from
+    uint32 public immutable sourceNetworkId;
+
+    /// @notice Address of the trusted source chain adapter that emits TeleporterV2MessageSent logs
+    address public immutable sourceEmitter;
```
batconjurer: wouldn't this be another instance of this contract? If so, wouldn't Nick's method mean we don't need to store this address?

batconjurer (commented):
One question, but generally looks good

icm-contracts/avalanche/verifiers/ethereum/ZKAdapter.sol
mdelle1: So this contract is already chain-specific by construction since ZKStateManager only tracks beacon checkpoints for the chain it was initialized with -- it can't verify any source chain

ylg-avalabs (approved):

icm-contracts/avalanche/verifiers/ethereum/ZKAdapter.sol
```diff
@@ -42,30 +56,49 @@ contract ZKAdapter is ZKStateManager, IMessageVerifier {
             admin,
             superAdmin
         )
-    {}
+    {
+        sourceNetworkId = _networkIdForChain(sourceChainId_);
```
geoff-vball, line 60: We shouldn't be using the source's chainID here, we should use `block.chainid` to check what chain this contract is deployed on.

icm-contracts/avalanche/verifiers/ethereum/ZKAdapter.sol
```diff
@@ -21,9 +25,19 @@ struct Attestation {
 /**
  * @notice Adapts ZKStateManager for the Teleporter messaging protocol.
  */
-contract ZKAdapter is ZKStateManager, IMessageVerifier {
+contract ZKAdapter is ZKStateManager, IAdapter {
+    /// @notice topic0 of the TeleporterV2MessageSent event which is the only log topic verifyMessage accepts
+    bytes32 private constant _MESSAGE_SENT_TOPIC = keccak256("TeleporterV2MessageSent(bytes)");
+
+    /// @notice The network ID (mainnet/testnet/local) messages must originate from
+    uint32 public immutable sourceNetworkId;
```
geoff-vball, line 33: This should just be `networkId` which is the network this contract is deployed to. Then this contract should only accept messages that contain the same `sourceNetworkID`

icm-contracts/avalanche/verifiers/ethereum/ZKAdapter.sol
mdelle1: `sourceChainId_` is the chain we're proving messages from so either Ethereum mainnet or Sepolia, not the chain this contract is deployed on. `block.chainid` would be the id for the C-chain. So I think this is correct as is

icm-contracts/avalanche/verifiers/ethereum/ZKAdapter.sol
mdelle1: Similarly as above the network ID here should be `sourceNetworkId` not the id of the local network the contract is deployed on, since we are verifying incoming messages from a source chain

batconjurer (approved):

geoff-vball (approved):

ylg-avalabs (approved):

geoff-vball (approved):

PR #1373, opened 2026-06-12
`ZKVerifier` documentation

mdelle1:
Public documentation for the `ZKVerifier` contract. Issue: https://github.com/ava-labs/icm-services/issues/1303


geoff-vball (approved):

batconjurer (changes requested):
This doc references ethereum beacon state roots and slots without definitions.
Several structs have their fields referenced; it would good to inline the struct definitions. 
Similarly, I think some contract state is referenced (_allowedBeaconBlocks), but not explicitly mentioned that it is contract state.

batconjurer (approved):

PR #1374, opened 2026-06-17
Refactor message handler

geoff-vball:
## Why this should be merged
Part of a larger Application Relayer and Message Handler refactor.

## How this works
The main purpose of this PR is to move `ProcessMessage` to the `MessageHandler`. This means that the `MessageHandler` has to expose much less in it's interface.

It does have the side-effect of needing `relayerMetrics` in the message handler, but this can be cleaned up in a followup PR. 

`ProcessMessage` now takes two params, but this will be cleaned up in a later PR.

## How is this documented


messages/teleporter/message_handler.go
```diff
@@ -314,8 +314,69 @@ func (m *messageHandler) SendMessage(signedMessage *warp.Message) (common.Hash,
 	return txHash, nil
 }
 
-func (m *messageHandler) LoggerWithContext(logger logging.Logger) logging.Logger {
-	return logger.With(m.logFields...)
+// ProcessMessage relays the message to the destination chain by aggregating a signature for it
+// and sending it via SendMessage. It does not retry on failure or checkpoint the height.
+// Returns the transaction hash if the message is successfully relayed.
+func (m *messageHandler) ProcessMessage(
```
batconjurer, line 320: This method is precisely the same as in the off-chain-registry. Is there not a a good way to reuse this function?

geoff-vball: Yeah, I think we can extract the common functionality for warp messages (off-chain and on-chain), but I'm going to save that for a later step in this refactor

batconjurer (approved):

messages/off-chain-registry/message_handler.go
```diff
@@ -186,6 +190,65 @@ func (m *messageHandler) SendMessage(signedMessage *warp.Message) (common.Hash,
...
+		m.metrics.IncFailedRelayMessageCount("failed to check if message should be sent")
+		return common.Hash{}, fmt.Errorf("failed to check if message should be sent: %w", err)
+	}
+	if !shouldSend {
+		m.logger.Info("Message should not be sent")
+		return common.Hash{}, nil
+	}
+	unsignedMessage := m.unsignedMessage
+
+	startCreateSignedMessageTime := time.Now()
```
mdelle1, line 212: Should the `GetPChainHeightForDestination` be included in the signing time latency? or should we start the timer after line 218?

mdelle1:
LGTM just one small comment

messages/off-chain-registry/message_handler.go
geoff-vball: I think we want it to include getting the pchain height. That way the metric lines up with the ctx and timeout we create on line 214

mdelle1 (approved):

PR #1377, opened 2026-06-17
Implements `ZKValidatorSetRegistry` contract

mdelle1:
This PR adds `ZKValidatorSetRegistry`, a variant of `MerkleValidatorSetRegistry` that verifies ICM messages by checking a single SP1 zero-knowledge proof in place of the on-chain Merkle multi-inclusion proof and aggregate BLS signature. The validator set is still anchored by a Merkle root commitment, but the signing validators, multi-proof, and aggregate signature are no longer explicitly performed --- they are essentially absorbed by the ZK proof verification, and since they are private inputs to the SP1 prover, do not appear as extra calldata. 

The gas savings are that on-chain verification cost becomes a constant size ZK proof check that does not scale with the number of signers, in contrast with `MerkleValidatorSetRegistry`, where there is extra calldata associated gas needed to pass in the signer public keys along with hashing/aggregation work. The tradeoff however is extra off-chain cost for generating the SP1 proofs. 

The contract verifies the proof through the canonical `ISP1Verifier` gateway and checks the public values associated to the proof against its own state.

Unit tests are added using a mock SP1 proof verifier. The proof soundness and the guest SP1 program are out of scope and will follow in the e2e PR. 

Closes issue https://github.com/ava-labs/icm-services/issues/1369


icm-contracts/ethereum/ZKValidatorSetRegistry.sol
```diff
@@ -0,0 +1,248 @@
+// (c) 2026, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+// SPDX-License-Identifier: LicenseRef-Ecosystem
+pragma solidity ^0.8.30;
+
+import {IMerkleValidatorSetRegistry} from "./interfaces/IMerkleValidatorSetRegistry.sol";
+import {ICMMessage} from "../common/ICM.sol";
```
geoff-vball: It would be a little better to use `@common` instead of the relative path

icm-contracts/ethereum/ZKValidatorSetRegistry.sol
```diff
@@ -0,0 +1,248 @@
...
+import {ValidatorSetMerkleCommitment, ValidatorSets} from "./utils/ValidatorSets.sol";
+import {
+    TeleporterMessageV2Parsing,
+    TeleporterICMMessage,
+    TeleporterMessageV2
+} from "../common/TeleporterMessageV2.sol";
+import {IAdapter} from "../common/ITeleporterMessengerV2.sol";
+import {IWarpMessenger} from "../avalanche/subnet-evm/IWarpMessenger.sol";
```
geoff-vball:
```suggestion
import {IWarpMessenger} from "@subnet-evm/IWarpMessenger.sol";
```

icm-contracts/ethereum/ZKValidatorSetRegistry.sol
```diff
@@ -0,0 +1,248 @@
...
+        // Check the ZK proof's public values against the on-chain commitment
+        require(pv.sourceBlockchainID == blockchainID, "public value chain ID mismatch");
+        require(pv.root == commitment.root, "public value root mismatch");
+        require(pv.messageHash == sha256(signedData), "public value message hash mismatch");
+        require(
+            ValidatorSets.verifyWeight(pv.signedWeight, commitment.totalWeight),
+            "stake-weighted quorum threshold not met"
+        );
+
+        ISP1Verifier(sp1Verifier).verifyProof(attestationProgramVKey, publicValues, proofBytes);
```
batconjurer, line 246: This requires changes the proof currently in https://github.com/ava-labs/solana-contracts. We need to calculate the sum in in the zk proof and check it against the input signed weight.

batconjurer (commented):
Generally LGTM. We need to decide if this is actually a flow we are going to support.

icm-contracts/ethereum/ZKValidatorSetRegistry.sol
mdelle1: Hey I'm planning on writing the guest program in a follow up PR

icm-contracts/ethereum/ZKValidatorSetRegistry.sol
```diff
@@ -0,0 +1,248 @@
...
+        bool allowPChainFallback_,
+        ISP1Verifier sp1Verifier_,
+        bytes32 attestationProgramVKey_
+    ) {
+        avalancheNetworkID = avalancheNetworkID_;
+        pChainID = pChainID_;
+        allowPChainFallback = allowPChainFallback_;
+        sp1Verifier = sp1Verifier_;
+        attestationProgramVKey = attestationProgramVKey_;
+        _valSetCommitments[pChainID_] = ValidatorSetMerkleCommitment({
```
geoff-vball, line 80: Can we just pass in a `ValidatorSetMerkleCommitment` as the constructor arg?

icm-contracts/ethereum/ZKValidatorSetRegistry.sol
```diff
@@ -0,0 +1,248 @@
...
+     */
+    function verifyICMMessage(
+        ICMMessage calldata message,
+        bytes32 avalancheBlockchainID
+    ) public view {
+        require(pChainInitialized(), "P-chain not initialized");
+        require(isRegistered(avalancheBlockchainID), "No validator set registered to given ID");
+        require(message.sourceNetworkID == avalancheNetworkID, "Network ID mismatch");
+
+        bytes memory signedData = ValidatorSets.buildUnsignedWarpMessage(
```
geoff-vball, line 188: This can be in another PR, but I think we should split `buildUnsignedWarpMessage` into two separate functions - one for p-chain messages and one for other messages since the encoding is completely different anyway

icm-contracts/ethereum/ZKValidatorSetRegistry.sol
```diff
@@ -0,0 +1,248 @@
...
+     * @dev `attestation` is `abi.encode(bytes publicValues, bytes proofBytes)`. The public values
+     * are produced by the guest program, which performs the Merkle multi-inclusion proof and the
+     * aggregate BLS12-381 signature check in-circuit; this function checks those public
+     * values against the on-chain state. The final `verifyProof` call reverts on an invalid ZK proof.
+     */
+    function _verifyZKAttestation(
+        bytes calldata attestation,
+        bytes memory signedData,
+        bytes32 blockchainID
+    ) internal view returns (bool) {
```
geoff-vball: We never return false from this function. I assume `ISP1Verifier.verifyProof` will revert if it doesn't verify, so I would probably just have this function return nothing and revert on failure.

icm-contracts/ethereum/ZKValidatorSetRegistry.sol
mdelle1: Created issue: https://github.com/ava-labs/icm-services/issues/1379

icm-contracts/ethereum/ZKValidatorSetRegistry.sol
mdelle1: I suppose we could though I flattened the arguments like this just to have parity with `MerkleValidatorSetRegistry`

geoff-vball (approved):

icm-contracts/ethereum/ZKValidatorSetRegistry.sol
batconjurer: I had some frree time, so I fixed that up here https://github.com/ava-labs/solana-contracts/pull/3

batconjurer (approved):

PR #1387, opened 2026-06-24
feat(oracle): add oracle signature aggregation pipeline

boraerincplusplus:
## How is this documented


icm-contracts/avalanche/teleporterV2/OracleAdapter.sol
```diff
@@ -0,0 +1,282 @@
...
+struct OracleMessage {
+    // Identifies the external chain or data source (e.g. "solana", "bitcoin").
+    string sourceType;
+    // Program or contract address on the source chain.
+    string sourceAddress;
+    // Destination contract on this L1 that receives the decoded payload.
+    address destContract;
+    // Block or slot number on the source chain at which the event occurred.
+    uint64 sourceBlockHeight;
+    // Monotonically increasing per (sourceType, sourceAddress). Enforces replay protection.
```
geoff-vball: In teleporter we just use one nonce for all sources. I'm not sure if we need to split this up by type and address

icm-contracts/avalanche/teleporterV2/OracleAdapter.sol
```diff
@@ -0,0 +1,282 @@
...
+            abi.encode(
+                oracleMsg.sourceType,
+                oracleMsg.sourceAddress,
+                oracleMsg.destContract,
+                oracleMsg.sourceBlockHeight,
+                oracleMsg.nonce,
+                oracleMsg.payload
+            )
+        );
+        if (warpPayloadHash != msgHash) revert PayloadMismatch();
```
geoff-vball: If the warp payload is the exact same as the oracle message, we don't need to pass in the oracle message as a parameter to this function call

icm-contracts/avalanche/teleporterV2/OracleAdapter.sol
```diff
@@ -0,0 +1,282 @@
...
+
+        // 6. Emit before external call (CEI pattern), then deliver to destination.
+        //    The destContract is part of the signed warp payload (verified in step 3),
+        //    so the relayer cannot redirect to an arbitrary contract.
+        emit OracleMessageReceived(
+            messageID, oracleMsg.sourceType, oracleMsg.sourceAddress, oracleMsg.destContract
+        );
+
+        IOracleMessageReceiver(oracleMsg.destContract).receiveOracleMessage(
+            warp.sourceChainID,
```
geoff-vball: If the `sourceChainID` is always the chain we're on, and we're checking it in this function, we don't need to pass it down.

icm-contracts/avalanche/teleporterV2/OracleAdapter.sol
```diff
@@ -0,0 +1,282 @@
...
+    // -------------------------------------------------------------------------
+
+    /**
+     * @notice Returns true if the (sourceType, sourceAddress) pair is on the allowlist.
+     */
+    function isAllowed(
+        string calldata sourceType,
+        string calldata sourceAddress
+    ) external view returns (bool) {
+        return _allowedSources[keccak256(abi.encode(sourceType, sourceAddress))];
```
geoff-vball: This seems like a convoluted way to keep this map. It would be easier as a two-level mapping

`mapping(string => mapping(address => bool))`

signature-aggregator/aggregator/oracle_aggregator.go
```diff
@@ -0,0 +1,57 @@
...
+// OracleSignatureAggregator collects BLS signatures for oracle attestation
+// messages. It uses the same aggregation logic as SignatureAggregator but
+// routes requests to handler ID 4 (oracle attestation) instead of handler
+// ID 2 (native warp).
+//
+// The counterpart on each validator is oracle/validator.OracleVerifier,
+// registered at the same handler ID. Callers pass the Solana (or other
+// source-chain) lookup hint as the justification argument to
+// CreateSignedMessage; validators forward it to their sidecar.
+type OracleSignatureAggregator struct {
```
geoff-vball: I'm not sure I love this pattern. Are we able to create a signature aggregator object that is able to send to both handlers?

icm-contracts/tests/suites/oracle/oracle_suite_test.go
```diff
@@ -0,0 +1,252 @@
...
+	realSidecar = nil
+	if localNetworkInstance != nil {
+		localNetworkInstance.TearDownNetwork()
+		localNetworkInstance = nil
+	}
+}
+
+// waitForTCP polls addr until a TCP connection succeeds or deadline is exceeded.
+func waitForTCP(addr string, timeout time.Duration) {
+	deadline := time.Now().Add(timeout)
```
geoff-vball: For timeouts, we usually use a pattern using a context timeout, a ticker, and a select statement. [Here's an example](https://github.com/ava-labs/icm-services/blob/merkle-sign/icm-contracts/tests/flows/services/diff_updater.go#L245). This way the function will also respect any timeouts that the context already has included in it.

icm-contracts/tests/suites/oracle/oracle_suite_test.go
```diff
@@ -0,0 +1,252 @@
...
+		buildCmd.Dir = avalancheGoRoot
+		buildOut, buildErr := buildCmd.CombinedOutput()
+		log.Info(string(buildOut))
+		Expect(buildErr).Should(BeNil())
+
+		configPath := filepath.Join(repoRoot, "build/solanarpc-config.json")
+		configJSON := fmt.Sprintf(`{"rpc_url": %q}`, solanaRPCURL)
+		Expect(os.WriteFile(configPath, []byte(configJSON), 0o600)).Should(BeNil())
+
+		realEndpoint := fmt.Sprintf("127.0.0.1:%d", realSidecarPort)
```
geoff-vball: I don't think we need the `real` in the names here

icm-contracts/tests/flows/oracle/oracle_attestation.go
```diff
@@ -0,0 +1,435 @@
...
+	// Step 2: fetch the full transaction.
+	txRaw := post(map[string]any{
+		"jsonrpc": "2.0", "id": 1,
+		"method": "getTransaction",
+		"params": []any{txSig, map[string]any{
+			"encoding":                       "json",
+			"maxSupportedTransactionVersion": 0,
+		}},
+	})
+	var txResp struct {
```
geoff-vball: We should create some sort of Solana utility file for these kinds of structs

PR #1395, opened 2026-07-01
ICM message verification path e2e test for `ZKValidatorSetRegistry`

### LINKED ISSUE #1389: Add Go-side verification path e2e test for `ZKValidatorSetRegistry`

mdelle1:
The roundtrip verification test will be done as a separate nightly e2e test. This PR only tests the verification path using a Groth16 ZK proof. This proof is generated offline over a synthetic validator set from Rust scripts located in `/scripts/tools/zk-proof-gen` over a fixed message, which is submitted to a deployed `ZKValidatorSetRegistry` contract and verified on-chain. 

A full roundtrip requires generating a fresh proof per message over live aggregated signatures, which is too slow and costly for per-PR CI. The nightly test will cover this flow. 

Closes https://github.com/ava-labs/icm-services/issues/1389


icm-contracts/ethereum/external/sp1-contracts/v6.1.0/SP1VerifierGroth16.sol
```diff
@@ -0,0 +1,88 @@
...
+
+// Source: https://github.com/succinctlabs/sp1-contracts/blob/d3629729c3216eb51bd4859d027a8eb729399fa4/contracts/src/v6.1.0/SP1VerifierGroth16.sol
+// Modified: Updated Solidity pragma from ^0.8.20 to ^0.8.30 and renamed contract from SP1Verifier to SP1VerifierGroth16
+pragma solidity ^0.8.30;
+/* solhint-disable */
+
+import {ISP1Verifier, ISP1VerifierWithHash} from "../ISP1Verifier.sol";
+import {Verifier} from "./Groth16Verifier.sol";
+
+/// @title SP1 Verifier
```
batconjurer, line 11: Do we really want to keep copy pasting code from other libraries or should we add them to the lib subdir as submodules?

foundry.lock
```diff
@@ -0,0 +1,14 @@
+{
```
geoff-vball: Does this file accomplish anything for us? Our libs are currently locked to specific versions by virtue of being submodules.

icm-contracts/ethereum/external/sp1-contracts/v6.1.0/SP1VerifierGroth16.sol
mdelle1: The contract needed a rename to work with `abi_bindings.sh`, but it depends if we want to pull a whole repo as a submodule for one file? What are people's thoughts here

geoff-vball (approved):

ylg-avalabs (approved):

PR #1403, opened 2026-07-07
project(sidecar): Add OracleAdapter and MockOracleReceiver contracts with Go bindings

boraerincplusplus:
## NOTE:
**This is a draft, not to be merged just yet.** I'm opening this up to mark it for review and make the diff viewable on GitHub UI.

## Why this should be merged
Introduces Solidity contracts and generated Go bindings for the on-chain half of the oracle attestation pipeline. Ships one adapter, one mock receiver, one Go helper for constructing the delivery payload

## How this works
- `OracleAdapter.sol` implements `IAdapter` (verify + send). Plugs into `TeleporterMessengerV2` as the message verifier for oracle deliveries.
- `OracleAdapter.t.sol` contains Foundry unit tests covering happy path, source allowlisting, replay rejection, wrong-source-chain rejection, and access control 
- `MockOracleReceiver.sol`: test `ITeleporterReceiver` implementation.
- `OracleAdapter.go` + `MockOracleReceiver.go` are auto-generated ABI bindings and account for the bulk of the LOC diff

## How this was tested
- Foundry unit tests (`OracleAdapter.t.sol`) contains 9 test cases covering the verify path, admin methods, and replay protection.
- Full E2E pipeline in the icm-services E2E suites is in the subsequent PRs, including live Solana devnet verification


icm-contracts/avalanche/mocks/MockOracleReceiver.sol
```diff
@@ -0,0 +1,68 @@
+// (c) 2026, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+// SPDX-License-Identifier: LicenseRef-Ecosystem
+
+pragma solidity 0.8.30;
```
batconjurer:
```suggestion
pragma solidity ^0.8.30;
```

icm-contracts/avalanche/mocks/MockOracleReceiver.sol
```diff
@@ -0,0 +1,68 @@
...
+
+import {ITeleporterReceiver} from "@teleporter/ITeleporterReceiver.sol";
+
+/**
+ * @dev Test receiver for oracle E2E tests. Records the last delivered oracle message
```
batconjurer, line 11: nit: de-indent the body of this comment just so that it is in line with the rest of the repo

icm-contracts/avalanche/teleporterV2/tests/OracleAdapter.t.sol
```diff
@@ -0,0 +1,249 @@
+// (c) 2026, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+// SPDX-License-Identifier: LicenseRef-Ecosystem
+
+pragma solidity 0.8.30;
```
batconjurer:
```suggestion
pragma solidity ^0.8.30;
```

icm-contracts/avalanche/teleporterV2/tests/OracleAdapter.t.sol
```diff
@@ -0,0 +1,249 @@
+// (c) 2026, Ava Labs, Inc. All rights reserved.
```
batconjurer, line 1: Please move this file into the tests subdir

icm-contracts/avalanche/teleporterV2/OracleAdapter.sol
```diff
@@ -0,0 +1,268 @@
+// (c) 2026, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+// SPDX-License-Identifier: LicenseRef-Ecosystem
+
+pragma solidity 0.8.30;
```
batconjurer, line 6:
```suggestion
pragma solidity ^0.8.30;
```

icm-contracts/avalanche/teleporterV2/OracleAdapter.sol
```diff
@@ -0,0 +1,268 @@
...
+    // Identifies the external chain or data source (e.g. "solana", "bitcoin").
+    string sourceType;
+    // Program or contract address on the source chain.
+    string sourceAddress;
+    // Destination contract on this L1 that receives the decoded payload.
+    address destContract;
+    // Block or slot number on the source chain at which the event occurred.
+    uint64 sourceBlockHeight;
+    // Monotonically increasing globally across all oracle sources. Enforces replay protection.
+    uint64 nonce;
```
batconjurer: Is there a reason not to use a uint256 here? The underlying type is going to be padded out to 32 bytes anyway.

batconjurer: Same question for sourceBlockHeight

icm-contracts/avalanche/teleporterV2/OracleAdapter.sol
```diff
@@ -0,0 +1,268 @@
...
+    function sendMessage(
+        TeleporterMessageV2 calldata message
+    ) external override {
+        WARP_MESSENGER.sendWarpMessage(abi.encode(message));
+    }
+
+    /**
+     * @notice Verify a validator-attested oracle message.
+     *
+     * @dev The calling transaction MUST include the signed warp oracle message in its
```
batconjurer, line 140: I actually like this formatting better. Shame it isn't what we are using throughout the rest of the codebase. Best to be consistent though.

icm-contracts/avalanche/teleporterV2/OracleAdapter.sol
```diff
@@ -0,0 +1,268 @@
...
+     * @inheritdoc IMessageVerifier
+     */
+    function verifyMessage(
+        TeleporterICMMessage calldata message
+    ) external override returns (bool) {
+        uint32 warpIndex = abi.decode(message.attestation, (uint32));
+
+        // 1. Read the precompile-verified warp message. The BLS aggregate was already
+        //    checked against this L1's validator set during block execution.
+        (WarpMessage memory warp, bool valid) = WARP_MESSENGER.getVerifiedWarpMessage(warpIndex);
```
batconjurer, line 155: Do you have a write-up of the whole flow somewhere? In my mental model, there was no need for TeleporterV2 to "send" a message. Is the idea that relayers need to watch for an event from TeleporterV2 in order to know when to collect signatures? No that can't be it, because then someone would need to call the send function of TeleporterV2 and that would requires listening to an external party (like Solana). In such case, why not just make the relayer listen?

icm-contracts/avalanche/teleporterV2/OracleAdapter.sol
boraerincplusplus: makes total sense. thx

icm-contracts/avalanche/teleporterV2/tests/OracleAdapter.t.sol
boraerincplusplus: 👍

icm-contracts/avalanche/teleporterV2/OracleAdapter.sol
boraerincplusplus: my logic was that if I extend IAdapter, I can plug the OracleAdapter into TeleporterV2, giving a uniform interface, piggying back off of the existing replay protection, gas limit handling, and callback function `receiveTeleporterMessage`.  Not implementing `sendMessage` would either mean that I don't extend IAdapter, or a Liskov violation. I noticed that ZKAdapter just emits a log (which I assume the Avalanche to Ethereum relayer picks up), and WarpAdapter calls the `sendWarpMessage,` so I just copied that for now. I haven't though much about the outbound side of this, but depending on what we come up with, it's subject to change. I thought of it like a placeholder for now.

batconjurer: I see what you're going for, but with this approach, you will need the send message to be called. Otherwise the message isn't loaded into the slots and this line will fail. If we can just emit an event, that would be simpler.

boraerincplusplus: now emitting an event instead of Warp 👍

icm-contracts/avalanche/teleporterV2/tests/OracleAdapter.t.sol
```diff
@@ -0,0 +1,242 @@
...
+// See the file LICENSE for licensing terms.
+
+// SPDX-License-Identifier: LicenseRef-Ecosystem
+
+pragma solidity 0.8.30;
```
mdelle1:
```suggestion
pragma solidity ^0.8.30;
```

icm-contracts/avalanche/mocks/MockOracleReceiver.sol
```diff
@@ -0,0 +1,66 @@
...
+        address, // originSenderAddress is address(0) for oracle messages
+        bytes calldata message
+    ) external override {
+        if (msg.sender != teleporterMessenger) {
+            revert OnlyTeleporterMessenger(msg.sender, teleporterMessenger);
+        }
+
+        (
+            string memory sourceType,
+            string memory sourceAddress,
```
mdelle1, line 61: I believe the `sourceBlockHeight` and `nonce` are encoded as `uint256` in the OracleAdapter but here they are decoded as `uint64`s

icm-contracts/avalanche/mocks/MockOracleReceiver.sol
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * @inheritdoc ITeleporterReceiver
+     */
+    function receiveTeleporterMessage(
+        bytes32 sourceBlockchainID,
+        address, // originSenderAddress is address(0) for oracle messages
+        bytes calldata message
+    ) external override {
+        if (msg.sender != teleporterMessenger) {
+            revert OnlyTeleporterMessenger(msg.sender, teleporterMessenger);
```
mdelle1: I would change `OnlyTeleporterMessenger` to `UnauthorizedSender`

icm-contracts/avalanche/mocks/MockOracleReceiver.sol
```diff
@@ -0,0 +1,66 @@
...
+        ) = abi.decode(message, (string, string, uint64, uint64, bytes));
+
+        lastSourceChainID = sourceBlockchainID;
+        lastSourceType = sourceType;
+        lastSourceAddress = sourceAddress;
+        lastSourceBlockHeight = sourceBlockHeight;
+        lastNonce = nonce;
+        lastPayload = oraclePayload;
+        ++receiveCount;
+    }
```
mdelle1, line 82: I think it would be a good idea to emit a `OracleMessageReceived` event here like
```
event OracleMessageReceived(
    bytes32 indexed sourceBlockchainID,
    string sourceType,
    string sourceAddress,
    uint64 sourceBlockHeight,
    uint64 nonce,
    bytes payload
);
```

icm-contracts/avalanche/teleporterV2/OracleAdapter.sol
```diff
@@ -0,0 +1,272 @@
...
+        emit OracleMessageSent(message);
+    }
+
+    /**
+     * @notice Verify a validator-attested oracle message.
+     *
+     * @dev The calling transaction MUST include the signed warp oracle message in its
+     * access list. The warp precompile verifies the BLS aggregate during block
+     * execution before this function runs.
+     * message.attestation must be abi.encode(uint32 warpIndex, OracleMessage oracleMsg).
```
mdelle1: Should be:
message.attestation must be abi.encode(uint32 warpIndex)

boraerincplusplus: 👍

icm-contracts/avalanche/mocks/MockOracleReceiver.sol
boraerincplusplus: gotchu

icm-contracts/avalanche/mocks/MockOracleReceiver.sol
boraerincplusplus: 👍

icm-contracts/avalanche/teleporterV2/OracleAdapter.sol
```diff
@@ -0,0 +1,272 @@
...
+ *
+ * @custom:security-contact https://github.com/ava-labs/icm-contracts/blob/main/SECURITY.md
+ */
+contract OracleAdapter is IAdapter {
+    IWarpMessenger public constant WARP_MESSENGER =
+        IWarpMessenger(0x0200000000000000000000000000000000000005);
+
+    address public owner;
+
+    // keccak256(abi.encode(sourceType, sourceAddress)) => allowed
```
geoff-vball: I think this comment is wrong now

icm-contracts/avalanche/teleporterV2/OracleAdapter.sol
```diff
@@ -0,0 +1,272 @@
...
+
+    /**
+     * @notice Emitted when an oracle message passes verification and is marked for delivery.
+     * @param nonce         Global oracle nonce used as the replay-protection key.
+     * @param sourceType    External chain type (e.g. "solana").
+     * @param sourceAddress Source program/contract address.
+     * @param destContract  Destination contract that will receive the payload.
+     */
+    event OracleMessageVerified(
+        uint256 indexed nonce, string sourceType, string sourceAddress, address indexed destContract
```
geoff-vball: So indexed event fields are usually used by third parties that want to query every event that has a specific value of that field. You can index either up to 3 or 4 (I think you have can 4 indexes for an event, but I think the first is taken up by the event signature or something). I can imagine someone wanting to query *all message sent from Solana* or *all messages from sourceAddress x*, and I think both of those are probably more useful than indexing the nonce.

icm-contracts/avalanche/teleporterV2/OracleAdapter.sol
geoff-vball: Just to chime in here... Sometimes TeleporterV2 will have to "send" a message - depending on what underlying mechanism it's using. For instance, if TeleporterV2 is using the WarpAdapter, it will have to "send" the message, and I imagine it would be the same if we used some third part service. 

There is however a "class" of cross-chain mechanisms that don't require this step, and only require an event be emitted - such as what we have going on with the merkle set registry etc.

icm-contracts/avalanche/teleporterV2/OracleAdapter.sol
```diff
@@ -0,0 +1,277 @@
...
+     * access list. The warp precompile verifies the BLS aggregate during block
+     * execution before this function runs.
+     * message.attestation must be abi.encode(uint32 warpIndex).
+     * message.sourceBlockchainID must equal this chain's blockchain ID.
+     *
+     * @inheritdoc IMessageVerifier
+     */
+    function verifyMessage(
+        TeleporterICMMessage calldata message
+    ) external override returns (bool) {
```
ylg-avalabs, line 150: In the verifyMessage, it seems not read message.message. We did the check in WarpAdapter: https://github.com/ava-labs/icm-services/blob/main/icm-contracts/avalanche/teleporterV2/WarpAdapter.sol#L76

icm-contracts/avalanche/teleporterV2/OracleAdapter.sol
```diff
@@ -0,0 +1,277 @@
...
+
+        // 4. Source allowlist check. Validators also enforce this per-node, but the on-chain
+        //    check ensures a rogue validator cannot deliver to an unconfigured source.
+        if (!_allowedSources[oracleMsg.sourceType][oracleMsg.sourceAddress]) {
+            revert SourceNotAllowed(oracleMsg.sourceType, oracleMsg.sourceAddress);
+        }
+
+        // 5. Replay protection on the global oracle nonce. Independent of Teleporter's own
+        //    replay protection — guards against direct verifyMessage calls outside TeleporterV2.
+        if (_processedMessages[oracleMsg.nonce]) revert AlreadyProcessed(oracleMsg.nonce);
```
ylg-avalabs, line 175: `verifyMessage` is public, and anyone can call it. Because anyone can include the validator-signed warp message in their own transaction's access list and call `verifyMessage` directly, an attacker can "burn" any pending oracle nonce without delivering it. The subsequent legitimate delivery through TeleporterMessengerV2.receiveCrossChainMessage → verifyMessage then reverts with AlreadyProcessed, permanently blocking that oracle message.

`verifyMessage` should be restricted to calls from `TeleporterMessengerV2` (e.g. msg.sender == teleporterMessenger)

mdelle1 (approved):

geoff-vball:
Closing this for now until someone picks it back up

PR #1407, opened 2026-07-09
project(sidecar): Add oracle E2E test harness with mock and solanarpc sidecar specs

boraerincplusplus:
## NOTE: (PR3)
**This is a draft.** This is a draft, not to be merged just yet. I'm opening this up to mark it for review. 

## Why this should be merged
PRs #1405 and #1406 introduced the contracts and signature aggregator endpoint for oracle attestation, but had no integration test covering the full pipeline. This PR closes that gap with an E2E test suite that exercises the complete path from Solana event -> BLS signature collection -> warp message delivery -> on-chain callback.

## How this works:
- tests/cmd/oracle-sidecar/: A minimal mock gRPC binary that implements the `OracleSidecar` service and accepts every `Verify` request unconditionally, without needing a real sidecar or Solana RPC. Built into `build/oracle-sidecar` alongside the other ICM service binaries.
- `icm-contracts/tests/suites/oracle/`: A Ginkgo suite with two specs:
  1) Mock sidecar spec, which always runs. Spins up the mock sidecar on port 9900 and runs `OracleAttestation` with dummy data.
  2) Solanarpc sidecar spec, which only runs if `SOLANA_RPC_URL` is set. Builds the real sidecar/ binary from the `boraplusplus/oracle-sidecar` branch of AvalancheGo, writes a config pointing at the given RPC endpoint, and passes it to a second L1.
- `icm-contracts/tests/flows/oracle/oracle_attestation.go`: The shared test flow, parameterized by L1 and an optional Solana RPC URL. Covers the full pipeline in eight steps: 
       1) deploy OracleAdapter, 2) deploy TeleporterMessengerV2, 3) deploy MockOracleReceiver, 4) start the signature aggregator, 5) construct and ABI-encode an `OracleMessage`, 6) call POST /oracle/aggregate-signatures, 7) submit the signed warp message via `receiveCrossChainMessage`, and finally, 8) assert MockOracleReceiver recorded the correct payload, source address, and chain ID. Includes two sad-path cases delivery from a non-allowlisted source reverts with SourceNotAllowed, and replay of an already-delivered nonce reverts with AlreadyProcessed.
- `icm-contracts/tests/flows/oracle/solana.go`: Discovers a recent Memo Program transaction from a live Solana RPC endpoint, walking up to 20 candidates (an arbitrary upper limit) until it finds one with an actual Memo instruction (handling both top-level and CPI inner instructions, and v0 transactions with address lookup tables). The raw Ed25519 signature bytes become the justification field passed to validators.
- `BeforeSuite`starts the required sidecar process(es), TCP polls until they are ready, writes the validator chain config files pointing to their sidecar endpoints, and brings up a `LocalAvalancheNetwork` with one or two L1s depending on whether SOLANA_RPC_URL is set. AfterSuite kills both sidecar processes and tears down the network. One L1 tests `MockSidecar` one tests the real one.

## How this was tested
The mock sidecar path runs in CI as part of the services E2E suite. The solanarpc path was validated manually by setting SOLANA_RPC_URL=https://api.devnet.solana.com and AVALANCHEGO_PATH pointing to a binary from the oracle sidecar branch, confirming a real Memo Program transaction is fetched, signatures are collected at handler ID 4, and the signed message is delivered and accepted on-chain.


icm-contracts/tests/flows/oracle/oracle_attestation.go
```diff
@@ -0,0 +1,351 @@
...
+		msgPayload = []byte("e2e-test-payload")
+		justification = []byte("dummy-solana-tx-signature")
+	}
+
+	ginkgo.By("Step 4: Allowlist source on OracleAdapter")
+	allowTx, err := adapterContract.SetAllowedSource(deployOpts, "solana", sourceAddress, true)
+	Expect(err).Should(BeNil())
+	utils.WaitForTransactionSuccess(ctx, l1Info.EthClient, allowTx.Hash())
+
+	ginkgo.By("Step 5: Request BLS aggregate signature from validators")
```
geoff-vball: I don't think we've really used `ginkgo.By` in any of our tests. Does it basically just checkpoint the test?

icm-contracts/tests/flows/oracle/oracle_attestation.go
```diff
@@ -0,0 +1,351 @@
...
+			GasTipCap:  gasTipCap,
+			Value:      common.Big0,
+			Data:       data,
+			AccessList: icmutils.SignedWarpMessageToAccessList(signedMsg),
+		})
+		tx = utils.SignTransaction(tx, fundedKey, l1Info.EVMChainID)
+		utils.SendTransactionAndWaitForFailure(ctx, l1Info.EthClient, tx)
+	}
+
+	ginkgo.By("Sad path 1: delivery from non-allowlisted source is rejected (SourceNotAllowed)")
```
geoff-vball: Sad path 1 and then next going to step 6 seems awkward.

icm-contracts/tests/flows/oracle/solana.go
```diff
@@ -0,0 +1,161 @@
...
+			}
+		}
+	}
+	Expect(false).To(BeTrue(), "no Memo instruction found in the %d most recent Memo-tagged transactions", candidateLimit)
+	return solanaTxData{}
+}
+
+// tryExtractMemoInstruction fetches the transaction with the given signature and
+// scans its top-level and inner instructions for one that invokes the Memo
+// program. Returns (data, slot, true) on the first match, or (nil, 0, false)
```
geoff-vball: Is it possible to return `nil, 0, true` from this function?

icm-contracts/tests/utils/services_utils.go
```diff
@@ -46,10 +46,13 @@ const (
...
-	cmd := exec.Command("./scripts/build.sh")
-	out, err := cmd.CombinedOutput()
-	log.Info(string(out))
+	repoRoot, err := GetRepoRoot()
 	Expect(err).Should(BeNil())
+	cmd := exec.Command(filepath.Join(repoRoot, "scripts/build.sh"))
+	cmd.Dir = repoRoot
+	out, cmdErr := cmd.CombinedOutput()
+	log.Info(string(out))
+	Expect(cmdErr).Should(BeNil())
```
geoff-vball: I assume this was an issue in the repo previously? Could you split this out into it's own feature PR?

scripts/build_oracle_sidecar.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+# Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+# See the file LICENSE for licensing terms.
+
+set -o errexit
+set -o nounset
+set -o pipefail
```
geoff-vball: I'm a fan of condensing this into `set -euo pipefail` which I think is what we use in most of our scripts.

icm-contracts/avalanche/teleporterV2/OracleAdapter.sol
batconjurer: I want to understand the intended owner of these contracts.

abi-bindings/go/teleporterV2/OracleAdapter/packing.go
batconjurer: Is this file auto-generated? I suspect it is not given the lack of do not edit header. We should not put manually created files in this sub-dir. Instead, add the OracleAdapter contract to the default contract list in the abi-bindings script.

icm-contracts/avalanche/teleporterV2/OracleAdapter.sol
batconjurer: can you break this decoding into a helper method? It would also mean that abi packings would automatically be generated by the abi-bindings script so that you wouldn't need to do it manually in the e2e tests.

icm-contracts/tests/flows/oracle/oracle_attestation.go
```diff
@@ -0,0 +1,351 @@
...
+	. "github.com/onsi/gomega"
+	"go.uber.org/zap"
+)
+
+// oracleMsgABI encodes the oracle message payload that OracleVerifier expects.
+// Layout mirrors OracleMessage in network/p2p/oracle/message.go on the
+// boraplusplus/sidecar-verifier branch.
+var oracleMsgABI abi.Arguments
+
+func init() {
```
batconjurer: see my above comment about having a decodeOracleMessage helper. Then this boiler plate will be auto-generated by abi-bindings script.

icm-contracts/tests/flows/oracle/oracle_attestation.go
```diff
@@ -0,0 +1,351 @@
...
+// endpoint and uses its slot/program/payload as the oracle payload, exercising
+// the real solanarpc sidecar end-to-end.
+func OracleAttestation(
+	ctx context.Context,
+	log logging.Logger,
+	avalancheNetwork *network.LocalAvalancheNetwork,
+	l1Info testinfo.L1TestInfo,
+	solanaRPCURL string,
+) {
+	ginkgo.By("Step 1: Deploy OracleAdapter")
```
batconjurer, line 144: can you break up this function with some helpers? It will make it much easier to read/maintain

icm-contracts/avalanche/teleporterV2/OracleAdapter.sol
batconjurer, line 32: is there a reason to prefer this over a blockchain id?

icm-contracts/tests/flows/oracle/solana.go
```diff
@@ -0,0 +1,167 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package oracle
```
batconjurer: This file does not contain a test suite, so it does not belong in flows. This should go into test/utils or somesuch

icm-contracts/tests/utils/teleporter.go
```diff
@@ -492,7 +493,10 @@ func DeployWarpAdapterContract(
 	testInfo testinfo.NetworkTestInfo,
 	fundedKey *ecdsa.PrivateKey,
 ) common.Address {
-	byteCode, err := deploymentUtils.ExtractByteCodeFromFile("./out/WarpAdapter.sol/WarpAdapter.json")
+	repoRoot, err := GetRepoRoot()
```
batconjurer: why was this change necessary?

scripts/e2e_test.sh
```diff
@@ -106,13 +106,20 @@ source "$REPO_PATH"/scripts/versions.sh
...
-rm -rf $BASEDIR/avalanchego
-BASEDIR=$BASEDIR AVALANCHEGO_BUILD_PATH=$BASEDIR/avalanchego "${REPO_PATH}/scripts/install_avalanchego_release.sh"
-
-export AVALANCHEGO_BUILD_PATH=$BASEDIR/avalanchego
-export AVALANCHEGO_PATH=$AVALANCHEGO_BUILD_PATH/avalanchego
-export AVAGO_PLUGIN_DIR=$AVALANCHEGO_BUILD_PATH/plugins
+# Install the avalanchego binary and subnet-evm plugin, unless the caller
+# provided a custom build via AVALANCHEGO_PATH (e.g. the oracle suite requires
+# a build with sidecar-verifier support).
+if [ -z "${AVALANCHEGO_PATH:-}" ]; then
```
batconjurer, line 118: I'm not sure how I feel about this. Can we just make the script always use an avalanche go version with sidecar support?

batconjurer (changes requested):

icm-contracts/avalanche/teleporterV2/OracleAdapter.sol
boraerincplusplus: Yeah so they're ownable to allow the L1 operator to set the allowlist. Unlike WarpAdapter which is stateless and is the same across the network, this is intended to be stateful, per administrator. Switched it to OpenZeppelin.

boraerincplusplus: 👍

icm-contracts/tests/flows/oracle/oracle_attestation.go
boraerincplusplus: 👍

abi-bindings/go/teleporterV2/OracleAdapter/packing.go
boraerincplusplus: Yeah, at first I wanted to be less invasive, so I didn't add them to DEFAULT_AVALANCHE_CONTRACT_LIST, but  I changed that now, and so the functions are auto-generated as well

icm-contracts/tests/flows/oracle/oracle_attestation.go
boraerincplusplus: 👍

icm-contracts/avalanche/teleporterV2/OracleAdapter.sol
boraerincplusplus: External chains are not Avalanche L1s or subnets. We could assign Solana an arbitrary number in 32 byte space, and it would actually be incredibly low probability that the creation of another L1 produces a tx hash to collide with it. I just chose to use `sourceType` to logically separate external chains from Avalanche L1s. What do you think

icm-contracts/tests/flows/oracle/solana.go
boraerincplusplus: 👍

icm-contracts/tests/utils/teleporter.go
boraerincplusplus: This was from when I ran it through go test/gingko run, but after changing the entry point to ./scripts/e2e-test.sh, it is no longer necessary. good catch thx

scripts/e2e_test.sh
boraerincplusplus: Yeah I agree with you bu we can't pin it yet. The sidecar support lives on my avalanchego branch, and for the sake of simplicity we decided not to push to have it merged yet. I think that will come around after I'm gone as the teams merge into a monorepo and this goes towards release. If we pin the script to a build of my branch today, every other suite silently fails unless we keep rebasing it to absorb all `avalanchego` changes, and suites that depend on a different unmerged avalanchego change (such as acp118 valset change from last week)  would fail. This version allows both to coexist. I added a TODO for after it is merged.

PR #1424, opened 2026-07-27
Only skip bloom filter check on the primary network

geoff-vball:
## Why this should be merged
Partially reverts the changes from https://github.com/ava-labs/icm-services/pull/1421. We'll only skip the bloom filter check for the primary network for now. We'll need to revisit how to properly do this for all networks when SAE becomes enabled on L1s.

## How is this documented

### DIFF
--- a/relayer/listener.go
+++ b/relayer/listener.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/constants"
 	"github.com/ava-labs/avalanchego/utils/logging"
 	"github.com/ava-labs/icm-services/relayer/config"
 	"github.com/ava-labs/icm-services/types"
@@ -137,6 +138,7 @@ func newListener(
 	sub := evm.NewSubscriber(
 		logger,
 		blockchainID,
+		sourceBlockchain.GetSubnetID() == constants.PrimaryNetworkID,
 		ethWSClient,
 		ethRPCClient,
 		errChan,
--- a/types/types.go
+++ b/types/types.go
@@ -49,33 +49,48 @@ func NewICMBlockInfo(
 	header *types.Header,
 	ethClient ethereum.LogFilterer,
 	topics [][]common.Hash,
+	isPrimaryNetwork bool,
 ) (*ICMBlockInfo, error) {
-	var logs []types.Log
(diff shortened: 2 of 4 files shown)


mdelle1 (approved):

batconjurer (approved):

relayer/listener.go
```diff
@@ -137,6 +138,7 @@ func newListener(
 	sub := evm.NewSubscriber(
 		logger,
 		blockchainID,
+		sourceBlockchain.GetSubnetID() == constants.PrimaryNetworkID,
```
boraerincplusplus, line 141: I was confused by this at first because `PrimaryNetworkID` is `ids.Empty` but I guess that's idiomatic? It's the one network whose ID is zero?

vms/evm/subscriber_test.go
```diff
@@ -67,6 +67,7 @@ func makeSubscriberWithMockEthClient(t *testing.T, errChan chan error) (*Subscri
 	subscriber := NewSubscriber(
 		logging.NoLog{},
 		blockchainID,
+		false,
```
boraerincplusplus, line 70: NewSubscriber is used in two places, this is one of them 👍

vms/evm/subscriber.go
```diff
@@ -198,7 +201,7 @@ func (s *Subscriber) subscribe(retryTimeout time.Duration) error {
 // and writes them to the blocks channel consumed by the listener
 func (s *Subscriber) blocksInfoFromHeaders() {
 	for header := range s.headers {
-		block, err := relayerTypes.NewICMBlockInfo(s.logger, header, s.rpcClient, s.topics)
+		block, err := relayerTypes.NewICMBlockInfo(s.logger, header, s.rpcClient, s.topics, s.isPrimaryNetwork)
```
boraerincplusplus, line 204: 👍

types/types.go
```diff
@@ -85,6 +100,23 @@ func NewICMBlockInfo(
 	}, nil
 }
 
+// bloomContainsAnyEventTopic reports whether the block's bloom filter indicates the presence of at
```
boraerincplusplus, line 103: This was taken out by my PR, so adding it back in makes sense.

boraerincplusplus (approved):
LGTM

relayer/listener.go
geoff-vball: Yes I think so. I thought it might have been the hash of `ids.Empty`, but looks like it never gets overwritten

PR #1428, opened 2026-07-29
Warp-agnostic Message Handler

geoff-vball:
## Why this should be merged
Today the relayer assumes every message protocol sends its messages through the Warp precompile: the listener hardcodes the Warp event filter per protocol type, and message handlers receive a pre-parsed *warp.UnsignedMessage. This prevents relaying messages from protocols that don't emit Warp logs — e.g. an IAdapter implementation that emits its own contract events, or the TeleporterV2 Merkle validator-set registry delivery path.

This PR makes the relayer core Warp-agnostic. The relayer now only deals in raw source-chain messages, and each message protocol's handler factory owns both which logs to subscribe to and how to decode its own payload.

## How this works
New types.SourceMessage replaces types.WarpMessageInfo. It carries the protocol-agnostic fields the relayer needs for routing (SourceBlockchainID, ProtocolAddress, SourceTxID) plus an opaque Payload — the log data as encoded by the protocol, passed through unparsed.

MessageHandlerFactory gains an EventFilter() method returning the source-chain log filter (in ethereum.FilterQuery.Topics format) for that protocol's messages. The listener now gets its subscription filter from the registered factory instead of the hardcoded EventFilterForProtocol switch, which is removed. Factories for protocols whose messages aren't read from logs (off-chain registry) return nil.

NewMessageHandler and GetMessageRoutingInfo take a *types.SourceMessage instead of a *warp.UnsignedMessage. Warp-based protocols (Teleporter, TeleporterV2, off-chain registry) unpack the unsigned Warp message from the payload themselves via types.UnpackWarpMessage; the shared types.WarpEventFilter helper builds their filter.

MessageCoordinator.ProcessBlock takes the protocol address of the subscription that produced the block's logs and constructs a SourceMessage directly from each log — no Warp parsing in the coordinator. ProcessWarpMessage is renamed to ProcessMessage; the relay API and FetchWarpMessage (still Warp-specific by nature, since it looks up by Warp message ID) construct SourceMessages accordingly.

Architecture diagram in docs/architecture.md updated to match.

## How is this documented


batconjurer (commented):
seems generally good to me, but there is some merit to the copilot findings

relayer/message_coordinator.go
```diff
@@ -231,56 +232,55 @@ func (mc *MessageCoordinator) ProcessMessageID(
...
-	for _, warpLogInfo := range icmBlockInfo.Logs {
-		warpMessage, err := types.NewWarpMessageInfo(warpLogInfo)
-		if err != nil {
-			mc.logger.Error("Failed to parse log", zap.Error(err))
-			continue
-		}
-		appRelayer, handler, err := mc.getAppRelayerMessageHandler(warpMessage)
+	for _, log := range icmBlockInfo.Logs {
+		message := types.NewSourceMessage(blockchainID, protocolAddress, log)
+		appRelayer, handler, err := mc.getAppRelayerMessageHandler(message)
```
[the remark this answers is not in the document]
boraerincplusplus: I was gonna be like "yo the AI is onto something" but I guess if we are running our own infrastructure and we have the config set up to listen to teleporter events, we don't really care about this, right? Like if someone else wants to run a relayer pointing to some other events, none of our business, but there's no way that someone can mess with our config

relayer/api/relay_message.go
```diff
@@ -70,16 +70,19 @@ func relayMessageAPIHandler(logger logging.Logger, messageCoordinator *relayer.M
 		}
 		address := common.HexToAddress(req.SourceAddress)
```
boraerincplusplus, line 72: 👍

boraerincplusplus (approved):

messages/off-chain-registry/message_handler.go
```diff
@@ -68,15 +69,26 @@ func NewMessageHandlerFactory(
...
 	logger logging.Logger,
-	unsignedMessage *warp.UnsignedMessage,
+	message *types.SourceMessage,
 	destinationClient vms.DestinationClient,
 	signatureAggregator *aggregator.SignatureAggregator,
 	metrics messages.Metrics,
 	signingSubnetID ids.ID,
 	quorumNumerator uint64,
 ) (messages.MessageHandler, error) {
+	unsignedMessage, err := types.UnpackWarpMessage(message.Payload)
```
mdelle1: Should we move this helper to the `warp` handlers package, e.g. calling `warp.UnpackMessage` so that the `types` package has no knowledge of how to unpack a warp message

messages/off-chain-registry/message_handler.go
```diff
@@ -94,8 +106,12 @@ func (f *factory) NewMessageHandler(
 }
 
 func (f *factory) GetMessageRoutingInfo(
-	unsignedMessage *warp.UnsignedMessage,
+	message *types.SourceMessage,
 ) (messages.MessageRoutingInfo, error) {
+	unsignedMessage, err := types.UnpackWarpMessage(message.Payload)
```
mdelle1: Same comment as above

messages/off-chain-registry/message_handler_test.go
```diff
@@ -36,6 +37,16 @@ var (
 	destinationBlockchainID       ids.ID
 )
 
+// sourceMessage returns the source chain message that [unsignedMessage] was provided to the
+// relayer as, in the protocol agnostic form the message handler factory receives.
+func sourceMessage(unsignedMessage *warp.UnsignedMessage) *types.SourceMessage {
```
mdelle1:
```suggestion
// sourceMessage returns the source chain message that [unsignedMessage] was provided to the
// relayer as, in the protocol agnostic form the message handler factory receives.
func toSourceMessage(unsignedMessage *warp.UnsignedMessage) *types.SourceMessage {
```

mdelle1: Minor comment

relayer/message_coordinator.go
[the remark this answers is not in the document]
mdelle1: The issue is that a topic-only filter will pick up events from wrong contracts regardless of how correct the config is

ylg-avalabs (approved):

messages/off-chain-registry/message_handler.go
geoff-vball: We don't actually have a `warp` package, but I've moved this function to `utils` for now and eliminated the `types` package.

mdelle1 (approved):

PR #1435, opened 2026-08-04
Boraplusplus/filter by hash

boraerincplusplus:
## How is this documented

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -15,6 +15,7 @@ import (
 	"github.com/ava-labs/icm-services/utils"
 	ethereum "github.com/ava-labs/libevm"
 	"github.com/ava-labs/libevm/common"
+	"github.com/ava-labs/libevm/common/hexutil"
 	"github.com/ava-labs/libevm/core/types"
 	"go.uber.org/zap"
 )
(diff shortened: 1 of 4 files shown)


types/types.go
```diff
@@ -33,6 +34,16 @@ type ICMBlockInfo struct {
...
 
+// BlockHead is the subset of a newHeads notification the relayer uses. The
+// node-reported hash is kept verbatim: recomputing it client-side is not
+// reliable for chains whose headers carry fields this client does not know
+// about (e.g. SAE chains).
+type BlockHead struct {
+	Hash   common.Hash  `json:"hash"`
+	Number *hexutil.Big `json:"number"`
+	Bloom  types.Bloom  `json:"logsBloom"`
+}
```
geoff-vball: If the `newHeads` subscription returns for fields than this, we should just take all the fields and use what we need. No need for a new type just to throw away information that might be useful later if we're not going to gain any efficiency by doing so

boraerincplusplus: for sure, I'm happy to increase the scope of this struct. it's not about efficiency, but just keeping at least the `Hash` field.

types/types.go
```diff
@@ -33,6 +34,57 @@ type ICMBlockInfo struct {
...
+// HeaderSerializable in avalanchego's plugin/evm/customtypes). All fields are
+// optional so notifications from chains that omit some of them still decode;
+// the upstream serializable types cannot be reused here because each chain
+// family's generated decoder requires fields the other family omits, and
+// their "hash" is a marshal-only computed field that is dropped on unmarshal.
+//
+// The node-reported hash is kept verbatim: recomputing it client-side is not
+// reliable for chains whose headers carry fields this client cannot encode
+// (e.g. SAE chains).
+type BlockHead struct {
```
geoff-vball, line 50: Oh, I was hoping we could just use a type that already exists in Avalanchego. But I guess if we want to be able to parse from both libevm and subnet evm we need custom code here? If that's the case I think it would actually be better to just use the 3 fields you had before, and link to the places where we can find additional fields later.

boraerincplusplus: gotcha 👍

geoff-vball (approved):

iansuvak (approved):

geoff-vball (approved):

iansuvak (approved):

PR #1437, opened 2026-08-05
Catch up

boraerincplusplus:
## Why this should be merged
This closes the second half of the message loss issue on chains that serve geth style RPC. The live path fix (fetching logs by node reported block hash) covers blocks arriving over the websocket, but catch-up still queries logs by block number ranges. The upper end of a catch-up range is anchored to the newest block announced by the websocket node, and the HTTP request serving that range can land on a node that has not seen the newest blocks yet.

## How this works
Catch-up now splits the range in two. Everything except the last 10 blocks is processed exactly as before, with chunked range queries. Blocks old enough to be behind every node's view are safe to query this way, and going block by block over a long outage would be far too slow.

The relayer first fetches the block head by number, which returns a NotFound error when the serving node does not have the block yet, and retries until a node that has it answers. The head carries the node reported block hash, and the logs are then fetched by that hash, where an unknown hash is again an explicit, retried error. The hash has to come from the node because recomputing it client side does not work for chains whose headers carry fields this client cannot encode.

The 10 block boundary is a deliberate margin: measured disagreement between nodes behind our endpoints is sub-second, and 10 blocks is roughly 10 to 20 seconds of tolerated lag on the C-Chain. Spans of 10 blocks or fewer, which is what a typical reconnect produces, are processed entirely in strict mode. Can toggle this up or down as needed.

The only wiring change is wrapping the existing HTTP client so it can also fetch heads by number.

## How this was tested
Unit tests and load testing. I added a catch-up version to the load test scenario.

## How is this documented

### DIFF
--- a/vms/evm/subscriber.go
+++ b/vms/evm/subscriber.go
@@ -15,6 +15,7 @@ import (
 	"github.com/ava-labs/icm-services/utils"
 	ethereum "github.com/ava-labs/libevm"
 	"github.com/ava-labs/libevm/common"
+	"github.com/ava-labs/libevm/common/hexutil"
 	"github.com/ava-labs/libevm/core/types"
 	"github.com/ava-labs/libevm/ethclient"
 	"github.com/ava-labs/libevm/rpc"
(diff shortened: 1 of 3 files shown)


vms/evm/subscriber.go
```diff
@@ -118,18 +154,74 @@ func (s *Subscriber) ProcessFromHeight(startingHeight uint64, endingHeight uint6
...
+	}
 
-		err := s.processBlockRange(fromBlock, toBlock)
-		if err != nil {
-			s.errChan <- fmt.Errorf("failed to process block range: %w", err)
+	// Range queries are only trustworthy for blocks old enough that every node
+	// behind a load-balanced endpoint has them; the newest blocks are checked
+	// strictly, one by one. See strictTailBlocks.
+	strictStart := startingHeight
+	if endingHeight-startingHeight+1 > strictTailBlocks {
```
batconjurer, line 166: I'm not 100% sure about go's binop format convention, but I notice no spaces around them here, but a space on either side in the line below

geoff-vball (approved):

vms/evm/subscriber.go
boraerincplusplus: it's a gofmt quirk, when one operator has higher precedence, (+ and - over >) it will snip out the spaces from around the tighter one to make the ordering visible as well.

PR #1451, opened 2026-08-20
`zkrelayer` : Message Handler implementation 

mdelle1:
This PR adds the zk message handler package, implementing the `MessageHandlerFactory` and `MessageHandler` interfaces for TeleporterV2 messages originating on Ethereum and verified on the destination via the `ZKAdapter`, using Boundless proofs of Ethereum consensus. 

The handler recognizes `TeleporterV2MessageSent` events emitted by the adapter, decodes the packed `TeleporterMessageV2` payload, derives each message's beacon slot and receipt log position, and routes messages by the destination chain encoded in the message itself. 

Follow ups: the delivery chain (including proof construction and messenger submission), the consensus-transition updater, protocol type registration, etc.

Includes unit tests for config parsing, event decoding, routing, and log location.

Makes progress on Issue: https://github.com/ava-labs/icm-services/issues/1426


relayer/zkrelayer/message.go
```diff
@@ -0,0 +1,45 @@
+package zkrelayer
+
+import (
+	"github.com/ava-labs/libevm/common"
+)
+
+// MessageState represents a message's position in the relayer pipeline.
+// There are 4 states: detected, submitted, delivered, and failed.
+type MessageState string
```
batconjurer: I'm not the expert on idiomatic go, but wouldn't an enum be more appropriate?

batconjurer: Oh, maybe it doesn't play well with the json tooling

relayer/zkrelayer/store.go
```diff
@@ -0,0 +1,74 @@
...
+
+import (
+	"encoding/json"
+	"os"
+
+	"github.com/ava-labs/libevm/common"
+)
+
+// Store is an interface for persisting the relayer state.
+type Store interface {
```
batconjurer: @iansuvak was unhappy about the idea of adding state to the relayer, although that was in re the checkpointing of processed blocks post-helicon. Looks like we might end up with added state anyway

relayer/zkrelayer/message.go
mdelle1: I believe Go has no built in enum type and these typed constants are the equivalent. I could use iota to enumerate with ints, but I preferred to use explicit strings since otherwise it would be a bit more unreadable when state is loaded in e.g. there would be entries like "`state`": 2 I'd have to translate

relayer/zkrelayer/store.go
iansuvak: Existing relayer already has state although it's minimal and we should keep it minimal

geoff-vball (commented):
What do we expect to be the trigger for messages to be relayed with the zk relayer? Are they being sent through an `TeleporterMessengerV2` contract on Ethereum?

batconjurer:
> What do we expect to be the trigger for messages to be relayed with the zk relayer? Are they being sent through an `TeleporterMessengerV2` contract on Ethereum?

Currently, the design is that they will listen for an event emitted by one of the IAdapters called by `TeleporterMessngerV2`. Then they will create the zk proof off line somehow and then relay it.

geoff-vball (commented):
I don't think we need any of this code. We definitely don't want to duplicate the DB for this. 

The current relayer implementation checkpoints progress through a hash of `protocolAddress, sourceBlockchainID, destinationBlockchainID, originSenderAddress, destinationAddress`. The zkrelayer will just have a different protocol address than anything else being relayed and will be checkpointed separately.

All you should really need to implement to get the ZK relayer implementation up and running is a new implementation of `message_handler.go`

mdelle1:
> I don't think we need any of this code. We definitely don't want to duplicate the DB for this.
> 
> The current relayer implementation checkpoints progress through a hash of `protocolAddress, sourceBlockchainID, destinationBlockchainID, originSenderAddress, destinationAddress`. The zkrelayer will just have a different protocol address than anything else being relayed and will be checkpointed separately.
> 
> All you should really need to implement to get the ZK relayer implementation up and running is a new implementation of `message_handler.go`

Refactored this PR to implement `MessageHandler`

batconjurer (approved):

geoff-vball (approved):

PR #1453, opened 2026-08-21
Subscribe filter log

ylg-avalabs:
## How is this documented


geoff-vball (commented):
So, I think as discussed in standup (sorry if we weren't that clear), that I think we can actually keep most of the old behaviour almost identical, except we will need to:

- Push all empty blocks to the heap where we track completed blocks
ex: if block `3` was the last block with logs in it, and we receive logs in block `6`, then we should push `4` and `5` to the heap (this could be done immediately). The relayer (or listener, or subscriber, not sure were) will have to keep track of the last block it received that that logs in it.
- Have a periodic check (~1 hour) that checks the newest block height, and tries to checkpoint the relayer some number of blocks back (to provide a buffer on delays getting logs from head), which will obviously be a no-op if that block is older than the most recent logs we've received.

geoff-vball:
I think this PR is still doing too much. It might be simpler to start back from scratch. All we need to do is 

1. Use the WS to get new filter heads
2. When we get a new block header with logs, insert all the block numbers between the new block and the last seen block to the checkpoint heap
3. Have a periodic goroutine that will check the current block height, and checkpoint if we haven't seen any new blocks in a while.

PR #1465, opened 2026-08-26
Fix receipt marking gas cost

geoff-vball:
## Why this should be merged
Closes https://claude.ai/security?project=scanproj_019LCMtLJaK3JHExaMcFeX7Z&finding=scan_018ZtbcWNnFGkytCFgch5FPX-3492944

## How this works
Introduces a non-retryable error - deterministic message send failures that will never resolve. We let the message fail and checkpoint the block anyway to prevent potential griefing attacks

## How is this documented


icm-contracts/avalanche/teleporter/tests/MarkReceiptGasBenchmarkTests.t.sol
```diff
@@ -0,0 +1,167 @@
...
+ * The sends that create this state happen in setUp() rather than in the test body. Foundry runs
+ * setUp() as its own call, so the slots they touch are cold again when the measured receive runs
+ * - matching a real delivery, which is always a separate transaction. Measuring the sends and the
+ * receive in one test body instead leaves the SentMessageInfo slots warm and understates the cost
+ * by the cold-access surcharge on every receipt.
+ */
+contract MarkReceiptGasBenchmarkTest is TeleporterMessengerTest {
+    // Sample sizes. Marginal cost comes from differences between them, so the fixed cost of the
+    // receive path cancels out. Three sizes rather than two so the result also shows whether the
+    // cost is linear in the receipt count - if it were superlinear, no per-receipt constant in
```
batconjurer: nit: I would adjust the comment slightly to make clear that linear is the expected behavior.

icm-contracts/avalanche/teleporter/tests/MarkReceiptGasBenchmarkTests.t.sol
```diff
@@ -0,0 +1,167 @@
...
+
+    /**
+     * @dev Sends `count` fee-bearing messages and records a receipt for each into `receipts`.
+     * Every receipt names a distinct, never-before-used relayer reward address so that marking it
+     * pays a cold zero-to-nonzero SSTORE on _relayerRewardAmounts.
+     */
+    function _stageReceipts(
+        TeleporterMessageReceipt[] storage receipts,
+        uint256 count,
+        uint256 salt
```
batconjurer: nit: To harden against accidental salt reuse, a global salt counter could be used and incremented within this function. Then the salt would not be exposed as a parameter.

messages/teleporter/message_handler_test.go
```diff
@@ -367,3 +411,133 @@ func TestSendMessageAlreadyDelivered(t *testing.T) {
 	_, err = handler.(*messageHandler).SendMessage(signedMessage)
 	require.NoError(t, err)
 }
+
+// A delivery that reverts having consumed its entire gas limit ran out of gas. That is
+// deterministic, so the relayer must surface it as non-retryable rather than re-broadcasting and
+// paying for the same reverted transaction again.
+func TestSendMessageOutOfGasIsNonRetryable(t *testing.T) {
```
batconjurer, line 488: can we get a boundary test on this? So if we reduce the number of receipts slightly is does get relayed?

batconjurer (dismissed):

iansuvak (approved):

batconjurer (approved):

PR #1470, opened 2026-08-27
Fix panic on incorrect signature length

geoff-vball:
## Why this should be merged
Fixes https://claude.ai/security?project=scanproj_019LCMtLJaK3JHExaMcFeX7Z&finding=scan_018ZtbcWNnFGkytCFgch5FPX-3492413

## How is this documented


batconjurer (approved):

mdelle1 (approved):

signature-aggregator/aggregator/aggregator.go
```diff
@@ -1051,5 +1044,21 @@ func (s *SignatureAggregator) unmarshalResponse(responseBytes []byte) (blsSignat
...
 	}
+	// An absent signature field is equivalent to an empty response: the node has not
+	// seen the message yet.
+	if len(sigResponse.Signature) == 0 {
+		return blsSignatureBuf{}, nil
+	}
+	// The signature length is chosen by the responding peer, so it must be validated before
+	// the conversion below: converting a slice shorter than the array panics at runtime, and
+	// converting a longer one would silently truncate.
+	if len(sigResponse.Signature) != bls.SignatureLen {
```
mdelle1, line 1055: Why is this check removed from isValidSignatureResponse and added here instead?

mdelle1:
One minor comment but LGTM

signature-aggregator/aggregator/aggregator.go
geoff-vball: Because this function returns a `blsSignatureBuf`, which copies `sigResponse.Signature` into that struct. The comment above this line explains the behaviour, where that conversion truncates longer responses and will panic on a shorter one.

PR #1477, opened 2026-09-01
Remove unused osxcross checkout from release workflows

ylg-avalabs:
## Why this should be merged
fix: https://app.notion.com/p/3c710c29114e80bba97ddf6809b1864b?v=05851eb6c16a41eaab0132afeb8ac2b2&p=3c710c29114e8005b234c4aa6da8e7a0&pm=s

## How is this documented


geoff-vball (commented):
If you can tag this branch with something that will trigger the release flows like `icm-relayer-test` to make sure this flow works, then we can merge this fix in.

PR #1481, opened 2026-09-03
Fix TeleporterMessengerV2 retryMessageExecution hash mismatch

ylg-avalabs:
## Why this should be merged
fix: https://app.notion.com/p/3c710c29114e80bba97ddf6809b1864b?v=05851eb6c16a41eaab0132afeb8ac2b2&p=3c710c29114e80a88979f325859d7e50&pm=s

## How is this documented


icm-contracts/common/TeleporterMessengerV2.sol
```diff
@@ -822,16 +812,38 @@ contract TeleporterMessengerV2 is ITeleporterMessengerV2, ReentrancyGuards, Init
...
         bytes32 messageID,
         bytes32 sourceBlockchainID,
-        TeleporterMessage memory message
+        TeleporterMessageV2 calldata message
     ) private {
         receivedFailedMessageHashes[messageID] = keccak256(abi.encode(message));
 
         // Emit a failed execution event for anyone monitoring unsuccessful messages to retry.
-        emit MessageExecutionFailed(messageID, sourceBlockchainID, message);
+        emit MessageExecutionFailed(messageID, sourceBlockchainID, _toLegacyMessage(message));
```
geoff-vball, line 827: I think we should be able to store the `TeleporterMessageV2` and use the full struct to hash for failed messages instead of the the legacy message. I think it might require passing the `TeleporterMessageV2` to `_handleInitialMessageExecution` and converting to a legacy message there.

PR #1487, opened 2026-09-09
Filter logs

geoff-vball:
## Why this should be merged
The relayer subscribed to newHeads and processed every block on the source chain: a notification per block, a bloom check, and often an eth_getLogs fetch, regardless of whether the block carried any messages. On a quiet chain almost all of that work found nothing. This PR subscribes to only the logs the listener's message protocol cares about, so the relayer does no per-block work for blocks that produced no relevant logs.

## How this works 
Subscription. The subscriber now calls SubscribeFilterLogs with the protocol's EventFilter instead of subscribing to heads. A block's logs arrive one notification at a time with no end marker, so on the first notification for a block the subscriber fetches the block's full matching log set by hash and ignores the block's remaining notifications. Log fetches by hash always go over the WS connection, since the subscribed node is guaranteed to have receipts for any block it has notified about or reported as its head.

Block ranges. ICMBlockInfo now covers an inclusive range [FromBlock, ToBlock] rather than a single height. The subscriber tracks the highest block it has dispatched; when the subscription reports a block above it, the blocks in between contain no matching logs and are folded into that block's range. Catch-up range queries emit one range per block with logs, folding in the empty blocks before it, plus one trailing empty range. The checkpoint manager commits ranges in order via StageCommittedHeights(from, to), using a min-heap of ranges keyed by start height, and tolerates overlapping or already-committed ranges.

Catch-up bound. Catch-up used to wait for the first live header to bound itself. With filtered logs that notification may not arrive for hours, so Subscribe now reads the subscribed node's head over the WS connection immediately after opening the subscription and dispatches catch-up up to that head. The subscription is opened before the bound is read, so no block can fall between the two. Blocks accepted in that small window are processed by both paths; relaying is idempotent and the checkpoint manager handles the overlap.

Idle checkpointing. On a chain with no relevant logs the checkpointed height would otherwise stay flat. An hourly ticker in the listener calls ProcessIdleBlocks, which catches up to the current head if the subscription has reported nothing since the previous tick. It runs real catch-up rather than assuming the range is empty, so a notification still in flight cannot be checkpointed over.

Removed. The bloom filter shortcut and its primary-network bypass, the isPrimaryNetwork plumbing, the newHeads header decoder, IsCatchup, and the UInt64Heap utility. The remaining BlockHeader type only needs hash and number.

## How is this documented


vms/evm/subscriber.go
```diff
@@ -320,24 +472,97 @@ func (s *Subscriber) subscribe(retryTimeout time.Duration) error {
...
+	}
+	notify := func(err error, duration time.Duration) {
+		s.logger.Info(
+			"get block number failed, retrying...",
+			zap.Duration("retryIn", duration),
+			zap.Error(err),
+		)
+	}
+
+	if err := utils.WithRetriesTimeout(operation, notify, utils.DefaultRPCTimeout); err != nil {
```
iansuvak: not a huge deal but we are using the same timeout  per request/operation here and for the total retries budget.

vms/evm/subscriber.go
```diff
@@ -291,6 +393,53 @@ func (s *Subscriber) Subscribe(retryTimeout time.Duration) error {
 	if err != nil {
 		return fmt.Errorf("failed to subscribe to node: %w", err)
 	}
+
+	// The head must be read from the subscribed node (over the WS connection)
+	// after the subscription is open. A node that is behind the subscribed
+	// node could report a height that the subscription will never deliver
+	// logs for, leaving a gap between catch-up and the subscription.
+	return s.catchUpToHead()
+}
```
iansuvak, line 412: Is there a race here? we subscribe prior to calling `s.catchUpToHead()` Can't subscribtion update the `highestDispatched` before the range is determined for catchup?

geoff-vball: Good catch!

relayer/application_relayer_test.go
```diff
@@ -109,7 +109,7 @@ func TestProcessHeightSkipsNonRetryableMessage(t *testing.T) {
 
 	relayer := newTestApplicationRelayer(t, ctrl, metrics)
 	errChan := make(chan error, 1)
-	relayer.ProcessHeight(1, []messages.MessageHandler{handler}, errChan)
+	relayer.ProcessBlocks(1, 1, []messages.MessageHandler{handler}, errChan)
```
batconjurer, line 112: it is worth testing the presence of a non-retryable message in a range?

relayer/application_relayer_test.go
```diff
@@ -120,7 +120,7 @@ func TestProcessHeightSkipsNonRetryableMessage(t *testing.T) {
 
 // Failures that are not marked non-retryable keep the existing behaviour: they surface to the
 // listener rather than being silently swallowed.
-func TestProcessHeightSurfacesRetryableFailure(t *testing.T) {
+func TestProcessBlocksSurfacesRetryableFailure(t *testing.T) {
```
batconjurer, line 123: similar question here

batconjurer (commented):
A couple of small Q's about test coverage. Otherwise looks good

iansuvak (approved):

ylg-avalabs (approved):

